PR #4, opened 2018-09-20
(Temporary) caching cdn handler until we sort out bundling

maartenbreddels:
(replaces #3)
# Issue to solve
For loading js modules, we have to choose a system, either the notebook extension method or the jupyter lab bundling. This new loader (frontend) will be a temporary solution that will 'just work' until we find the proper way to do this. It will try to fetch from `<server>/voila/require/module@semver`, which at the server will
 * try to find the cached version and return that
 * if not found will fetch it from a CDN (default is unpkg) and write it to cache

But only for modules in a whitelist (security concern)

# Issues:
It will not respect semver, so will never fetch a patched version once it cached a version. 

# Where cache?
It first tries `{prefix}/share/jupyter/voila_cache/{module}/{semver}`, and on failure will use the user directory.

### DIFF
--- a/js/loader.js
+++ b/js/loader.js
@@ -12,7 +12,8 @@ var requirePromise = function (pkg) {
 };
 
 export function requireLoader(moduleName, moduleVersion) {
-    return requirePromise(["" + moduleName]).catch(function (err) {
+    var path = `/voila/require/${moduleName}@${moduleVersion}`
+    return requirePromise([path]).catch(function (err) {
         var failedId = err.requireModules && err.requireModules[0];
         if (failedId) {
             console.log(`Falling back to unpkg.com for ${moduleName}@${moduleVersion}`);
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
We may also want to support the use case where all the widgets are in a single bundle like with jupyterlab.

The loading of external bundles could be activated as an opt-in.

There are benefits to both use cases (all in one bundle, and bundles loaded separately). @jasongrout may have an opinion on that.

jasongrout:
I think it makes a lot of sense to support the local widgets only use case.

SylvainCorlay:
We started having some in-depth discussions with @maartenbreddels about having a jupyterlab remix as a front-end for this, reusing the jupyterlab extension mechanism etc.

Would love to brainstorm with you about this.

jasongrout:
Sounds great. Let's talk today or tomorrow.

Juanlu001:
Would this help in making the load times a bit faster?

maartenbreddels:
Yes it would, what you can also do is put e.g. bqplot.js in the voila/static directory for now (If you open the browser dev console you should see it first tries to fetch that file before trying the cdn

trungleduc:
Closed due to inactivity

PR #20, opened 2018-11-20
WIP: Cleanup of handlers/urls

maartenbreddels:
For voila, I think it does not make sense (at least by default, or optional) to have authentication. At least to have a different behaviour for the notebook and the voila server extension.

What this allows, is to have a notebook server (normal authentication) but serve notebooks via voila without. This is very much a proof of concept and deserves more consideration. It exposes some of the issues that with notebook/jupyter_server has when you want to do this. For instance, to make `ZMQChannelsHandler` not check auth, we need to return a fake user. 

I think in general, voila shakes the box a bit, because instead of exposing kernels/notebooks as a security issue, now all of the sudden a kernel can be safe to expose (since it does not allow execution). Also to keep in mind is that multiuser all of the sudden makes sense.

Long term vision/idea: You could have a notebook run under your account, it ran a simulation or report for ~2 hours. Now users can connect to voila, they just want to see this report, which can be without authentication, or with authentication that has nothing to do with the notebook authentication.

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -23,15 +23,13 @@ import tornado.ioloop
 import tornado.web
 
 from traitlets.config.application import Application
-from traitlets import Unicode, Integer, Bool, Type, default
+from traitlets import Unicode, Integer, Bool, default
 
 from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
 from jupyter_server.services.kernels.handlers import KernelHandler, ZMQChannelsHandler
 from jupyter_server.base.handlers import path_regex
 from jupyter_server.services.contents.largefilemanager import LargeFileManager
 from jupyter_server.utils import url_path_join
-from jupyter_server.auth.login import LoginHandler
-
 
 from .paths import ROOT, STATIC_ROOT, TEMPLATE_ROOT
 from .handler import VoilaHandler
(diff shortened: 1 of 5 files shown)


timkpaine:
Why not have it support an overrideable login handler like the server does, then the default login handler can just do no auth.

voila/app.py
```diff
@@ -150,8 +150,8 @@ def start(self):
         handlers = []
 
         handlers.extend([
-            (url_path_join(base_url, r'/api/kernels/%s' % _kernel_id_regex), KernelHandler),
-            (url_path_join(base_url, r'/api/kernels/%s/channels' % _kernel_id_regex), ZMQChannelsHandler),
+            (url_path_join(base_url, r'/voila/api/kernels/%s' % _kernel_id_regex), KernelHandler),
+            (url_path_join(base_url, r'/voila/api/kernels/%s/channels' % _kernel_id_regex), ZMQChannelsHandler),
```
SylvainCorlay: 👍 it should indeed not be the same handler as for the notebook.

maartenbreddels:
@timkpaine yes that was a good idea actually

I started doing that, but decided to first get to a clean PR that sets up the handlers correctly and cleans up the code. I'll open a PR later for the auth.

@SylvainCorlay if you decide to merge it, just squash it, I just want to keep the history of this branch so I didn't squash it myself.

voila/utils.py
```diff
@@ -0,0 +1,13 @@
...
+# Distributed under the terms of the BSD 3-Clause License.                  #
+#                                                                           #
+# The full license is in the file LICENSE, distributed with this software.  #
+#############################################################################
+
+from jupyter_server.utils import url_path_join
+
+def add_base_url_to_handlers(base_url, handlers):
+    """Adds the base_url in front of the urls in the Tornado handlers"""
+    return [tuple([url_path_join(base_url, url)] + list(tail)) for url, *tail in handlers]
```
maartenbreddels: btw, this is py3 only, I guess that is ok

maartenbreddels:
Replaced bu #41

PR #27, opened 2018-11-23
WIP - Use templates from jupyter_server

SylvainCorlay:
We should not need to duplicate the base templates from `jupyter_server`.


maartenbreddels:
I'm not sure they should go into jupyter_server, should they?

SylvainCorlay:
Closing. Now that we have the templating mechanism, this is not relevant anymore.

Issue #36, opened 2018-12-05
Voila mode button/option


maartenbreddels:
Like appmode, https://github.com/oschuett/appmode it would be good to have a button or option in the menu bar for the classical notebook to open a notebook in voila (also make voila more 'visible'). For Jupyter lab, we could add a command to the command palette, and possibly add it to the menu.

jtpio:
That would be great indeed.

So the idea would be to open a new tab pointing to `/voila/render/notebook`?

DougRzz:
Could Voila serve a Rise slideshow? This could allow Voila to make use of the Rise slide type dropdown menus and enable the user to specify the cells to serve.

timkpaine:
@maartenbreddels it would be cool to add a "Share with voila" option. Basically a button you click that generates a unique (and maybe token-protected) link that you could send to someone else and have them (and only them) see. And also have an option to share a publicly viewable version.

maartenbreddels:
> So the idea would be to open a new tab pointing to `/voila/render/example`?

yes, that's now implemented.



> Could Voila serve a Rise slideshow

I think it could, but that would use a different template. Since we use nbconvert, and nbconvert already has a template for this, it is good to keep in mind. Maybe a really small template is required.



> @maartenbreddels it would be cool to add a "Share with voila" option.

Absolutely, that's is what I have in mind for the long term. I think the token idea would be something that could be implemented in jupyter_server, and then voila could make use of this.

SylvainCorlay:
I think that having a voila template to generate slides and work with rise is a great idea. We should probably track this with a separate issue and have a discussion with Damian on this.

maartenbreddels:
Yes, let us close this issue, feel free to open new issues @DougRzz and @timkpaine

PR #37, opened 2018-12-07
Add voila button for classical notebook, ala appmode

maartenbreddels:
Adds a button to the toolbar of the classical notebook, which makes voila more visible. 
screen shot 2018-12-07 at 11 48 15
solved #36


SylvainCorlay:
Great!

PR #40, opened 2018-12-07
WIP: extension to add voila to the command palette

maartenbreddels:
Gives an option to open the current notebook with the command palette:
screen shot 2018-12-07 at 13 46 28

 * Do we want 1 voila jlab extension? If yes, I could move it from `packages/jupyterlab-voila` to just `jupyterlab-voila`
 * Do we want to add it to the menu? If so, what would be a good place? View?
 * Do we want it to come with a preset shortcut key (e.g. cmd-something-v)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ __pycache__
 \#*#
 .#*
 .coverage
-src
 
 *.swp
 *.map
--- a/packages/jupyterlab-voila/package.json
+++ b/packages/jupyterlab-voila/package.json
@@ -1,0 +1,57 @@
+{
+  "name": "jupyterlab-voila",
+  "version": "0.1.0",
+  "description": "A JupyterLab extension for Voila",
+  "keywords": [
+    "jupyter",
+    "jupyterlab",
+    "jupyterlab-extension"
+  ],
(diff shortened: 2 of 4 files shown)


maartenbreddels:
To install: `jupyter labextension install ./packages/jupyterlab-voila`

For quick develop workflow:
```
jupyter labextension install ./packages/jupyterlab-voila --no-build
jupyter lab build  # never sure if this step is need before --watch
jupyter lab --watch
```
@blink1073 do you know if `jupyter lab build` is needed above?

jtpio:
>     * Do we want to add it to the menu? If so, what would be a good place? View?

View would be a good place. There is already "Presentation Mode" there could also be "Voila Mode" / "Render with Voila" / "View in Voila"

maartenbreddels:
@SylvainCorlay I would be ok with just adding to the command palette, and keep it to this. Then we have the scaffolding ready.

blink1073:
No, you don't need build in there.  It builds as part of the watch command.

jtpio:
Replaced by https://github.com/QuantStack/voila/pull/217 and https://github.com/QuantStack/voila/pull/250 with more to come.

Thanks @maartenbreddels for the initial work!

PR #44, opened 2018-12-09
Return version with `voila --version`

jtpio:
### Before

```bash
$ voila --version
0.0
```

### After

```bash
$ voila --version
0.0.6
```


SylvainCorlay:
Thanks!

PR #46, opened 2019-01-22
feat: use jupyterlab-manager output widget

### LINKED ISSUE #19: Support for @interact
ipywidget's `@interact` does not appear to work with voila

- widgets are all properly created
- the output widget works perfectly in a standalone fashion
- but the output for the `interactive` object does not seem to be generated by the `nbexecute` nbconvert preprocessor.

### LINKED ISSUE #23: TODO: use jupyterlab-manager's output widget
Most of the logic is duplicated from the version of the output widget from the jupyterlab manager. We should use the upstream implementation instead.

msuperina:
Most of the widget manager logic is duplicated from the
@jupyter-widgets/jupyterlab-manager package. This PR aims at using the
package directly.
The main benefit of this approach is support for output of "stream" messages.
It becomes possible to render and interact with interactive widgets.
An example of interactive widget has been added in
notebooks/interactive.ipynb

Fix #19
Fix #23


js/manager.js
```diff
@@ -1,149 +1,144 @@
...
+/***************************************************************************
+* Copyright (c) 2018, Voila contributors                                   *
+*                                                                          *
+* Distributed under the terms of the BSD 3-Clause License.                 *
+*                                                                          *
+* The full license is in the file LICENSE, distributed with this software. *
+****************************************************************************/
+
+import { RenderMimeRegistry, standardRendererFactories } from '@jupyterlab/rendermime';
+import { WidgetManager as JupyterLabManager } from '@jupyter-widgets/jupyterlab-manager/lib/manager';
```
msuperina: @jupyter-widgets/jupyterlab-manager does not have an API to consume the manager class directly. 
What approach would work best here ? 
I could always submit a PR there and see if they want to expose the manager ?

js/webpack.config.js
```diff
@@ -9,7 +9,7 @@ var loaders = [
           'babel-preset-stage-0',
         ].map(require.resolve),
     },
-    exclude: /node_modules/
+    exclude: /node_modules(?!(\/|\\)@jupyter-widgets(\/|\\)jupyterlab-manager)/
```
msuperina: If I import the jupyterlab-manager WidgetManager  class from node-modules it comes as es5 class and cannot be extended using the es6 extends syntax. If instead I add it to our babel transpilation process it is possible to extend it as babel sees it as es6 class.

js/manager.js
```diff
@@ -1,149 +1,144 @@
...
+import { HTMLManager, requireLoader } from '@jupyter-widgets/html-manager';
+import { WidgetRenderer } from '@jupyter-widgets/jupyterlab-manager/lib/renderer';
+import * as output from '@jupyter-widgets/jupyterlab-manager/lib/output';
+import * as base from '@jupyter-widgets/base';
+import * as controls from '@jupyter-widgets/controls';
+
+if (typeof window !== "undefined" && typeof window.define !== "undefined") {
+    window.define("@jupyter-widgets/base", base);
+    window.define("@jupyter-widgets/controls", controls);
+    window.define("@jupyter-widgets/output", output);
```
msuperina: I would like to find a way to get rid of those lines if possible.
If I remove `window.define("@jupyter-widgets/base", base);` I cannot load the bqplot example which requires the resource from unpkg and needs an external `@jupyter-widgets/base`

js/manager.js
```diff
@@ -1,149 +1,144 @@
...
+            moduleName === '@jupyter-widgets/output'
+        ) {
+            return super.loadClass(className, moduleName, moduleVersion);
+        }
+        else {
+            return htmlManager.loadClass(className, moduleName, moduleVersion);
+        }
+    }
+
+    _registerWidgets() {
```
msuperina: This method exists because the jupyterlab-manager API does not expose the plugin extension directly. I don't know how keen they are to change their API. Given that I see recurrent names of people in all jupyter related repositories I hope this comment can be picked up.

msuperina:
This is still a WIP and needs testing and cleanup. I thought opening this PR would help people playing with this branch.

SylvainCorlay:
@msuperina thanks for your contribution, this is great, and quite a feat!

I will put some cycles into this tomorrow!

msuperina:
@SylvainCorlay thanks for looking into this. I will wait for your feedback before any refactoring.

js/manager.js
```diff
@@ -1,149 +1,144 @@
...
+    async loadClass(className, moduleName, moduleVersion) {
+        if (
+            moduleName === '@jupyter-widgets/base' ||
+            moduleName === '@jupyter-widgets/controls' ||
+            moduleName === '@jupyter-widgets/output'
+        ) {
+            return super.loadClass(className, moduleName, moduleVersion);
+        }
+        else {
+            return htmlManager.loadClass(className, moduleName, moduleVersion);
```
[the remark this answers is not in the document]
maartenbreddels: I'd say, use the original loader idea. Since you already check those 3 modes, I think you could do:
```
...
 else 
      return this.loader(moduleName, moduleVersion)
```

js/manager.js
maartenbreddels: I'm not sure I follow you, this looks fine to me. I guess you see a reason not to have this code, could you explain?

js/manager.js
```diff
@@ -1,149 +1,144 @@
...
+    const rendermime = new RenderMimeRegistry({
+        initialFactories: standardRendererFactories
+    });
+    rendermime.addFactory({
+        safe: false,
+        mimeTypes: [WIDGET_MIMETYPE],
+        createRenderer: options => new WidgetRenderer(options, manager)
+    }, 1);
+    return rendermime;
+}
```
maartenbreddels: newline missing

voila/static/main.js
```diff
@@ -23,4 +23,4 @@ require(['libwidgets'], function(lib) {
             }
         );
     }
-});
+});
```
maartenbreddels: missing newline

js/manager.js
msuperina: Sorry I realize my comment was not clear. This piece of code exists in the jupyterlab-manager package and a slight modification of their API could allow us to fully consume their module without replicating some of the code.

js/manager.js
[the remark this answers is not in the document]
msuperina: I am not sure I understand: for the base, controls and output modules I want to use the jupyterlab-manager loadClass, but for the rest of the modules I want to keep using the html-manager loadClass, which internally handles promise rejections and allows me to not replicate such code.

vidartf:
I did not look through this in detail, but in general, I think it would be appropriate to have a look at refactoring the lab manager in order to make it more friendly for library use. If you want to open an issue on the widgets repo outlining some of the current pain-points, that would be an excellent starting point for such a discussion.

js/manager.js
SylvainCorlay: cc @jasongrout I think that this could be a meaningful change for the next release of the JupyterLab manager.

js/manager.js
SylvainCorlay: Yes, unfortunately this is still required for the bundle-based means of loading custom widgets.

Eventually, we will have voila templates that don't load AMD bundles and will not require this.

Maybe a good way to move towards a cleaner solution would be to move this out of this file and include it into a the nbconvert template.

js/webpack.config.js
SylvainCorlay: wow, gotcha

SylvainCorlay:
Hey @msuperina quick note on this: your git config email is `marzio@debian-mac.home` which does not match your GitHub email. You may want to change this so that github credits you for this work.

js/webpack.config.js
maartenbreddels: This puzzled me a bit, but it is the other way around. The output was an ES6 class (https://github.com/jupyter-widgets/ipywidgets/blob/2b91eac459bf4929c3b92794bfb0dc7deef2a353/packages/jupyterlab-manager/src/tsconfig.json#L12) which babel then tried to inherit from, and tried to construct without using new.
I've not configured babel 7 using more modern target browsers such that it output es6 classes for voila as well. The bundle size also seems to have halved.

js/manager.js
maartenbreddels: I think we can do that in a different PR ok?

js/manager.js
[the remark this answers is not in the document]
maartenbreddels: I duplicated the code, I think in this case it's cleaner for the moment, so it's safer in case we want to refactor this.

js/webpack.config.js
msuperina: Apologies, I should have suggested a babel upgrade as you did in your commit. I preferred not to as I am not a maintainer of this repo and such a change could have challenge the acceptance of this PR.

msuperina:
@SylvainCorlay @maartenbreddels I was very busy with work and could not follow up in the past few days. Is there any work you need me to do on this PR at this stage ?

maartenbreddels:
Thanks for taking this on, I think we're finished, unless Sylvain thinks otherwise. Would be good if you could test it. Also, feel free to change the log so that the email matches the github email address, now you won't get credits.

js/manager.js
SylvainCorlay: ok

SylvainCorlay:
@msuperina I am ready to get this in. Do you want to change your committer info so that your github accounts gets the proper credit?

msuperina:
@SylvainCorlay no please go ahead if this is what is stopping you 😃

SylvainCorlay:
If you want me to do it, I am fine as well. I am used to doing this sort of changes. Just give me the email address you want to use.

SylvainCorlay:
I just rebased the PR attributing Marzio's commits to his GitHub account (msuperina at users.noreply.github.com). Will merge when all is green.

Thanks @maartenbreddels for doing the work on babel7.

PR #47, opened 2019-01-22
Add support for `python -m voila`

jtpio:
So voila can be launched with either `voila` or `python -m voila`.


maartenbreddels:
good idea, if you could fix the missing newline, that would be great.

SylvainCorlay:
Discarding the failing tests which are unrelated.

Issue #48, opened 2019-01-23
Template/plugin system


maartenbreddels:
As discussed with @SylvainCorlay the template system could be pretty light weight and flexible if we adopt the following approach:
 * `$PREFIX/etc/jupyter/voila/conf.(yaml|json)` holds a general default configuration, which for now would only have the following options:
   ```
   default_nbconvert_template: default
   ```
 * `$PREFIX/share/jupyter/voila/nbconvert_templates/{nbconvert_template_name}/` directory which would be added to the template search path, to enable overloading of voila.tpl or base.tpl (or any of the nbconvert templates used)
 * `$PREFIX/share/jupyter/voila/static/{nbconvert_template_name}/` for any static content. Ideally this would enable similar to the templates directory overloading of files, such that for instance a new template could be based on an existing template, but just have a different css or logo image.

Different from what I discussed with @SylvainCorlay , I propose for the last two directories, the following layout:
 * `$PREFIX/share/jupyter/voila/{nbconvert_template_name}/nbconvert_templates` 
 * `$PREFIX/share/jupyter/voila/{nbconvert_template_name}/static`
 * (optionally `$PREFIX/share/jupyter/voila/{nbconvert_template_name}/template` so that we can also overload 404.html etc)

This reflects better the directory structure of the voila Python package, and makes it easier to fiddle with templates (make a symlink, manually remove or rename the directory).

This system makes it easy to conda/pip/apt-get install a template, since it just needs to put it in a directory, no post scripts are needed.

Open ideas:
* Ideally, a template could also 'inherit' from a different template. So the search path for templates and static paths just gets appended with the list of extra template paths.
 * The default template could go outside of the Python package, sharing the exact same system as the other templates.

SylvainCorlay:
I like that solution very much.

Indeed, this is more elegant to put the static directory into the template directory.

SylvainCorlay:
Closing as fixed since #49 was merged.

PR #50, opened 2019-01-24
Use classical notebook extension mechanism for enabling e.g. widgets

maartenbreddels:
If classical notebook extensions are installed/enabled, voila now picks those up and serves them (with the exception of jupyter-js-widgets).
This does not avoid other templates from using a different system, and with #48 we could show how to do this, e.g.
 * using a CDN
 * using this new approach of using the classical notebook system
 * a jupyter lab single bundle approach.


SylvainCorlay:
I just rebased the PR.

SylvainCorlay:
Checked locally and this works fine.

PR #53, opened 2019-01-24
Fixup template path

### LINKED ISSUE #51: Error when trying to install v0.0.7 as jupyter server extension
Looks like in the new release there's a bug when trying to use voila as server extension:

[W 21:40:22.117 ServerApp] Error loading server extension voila.server_extension
...
        from .paths import ROOT, TEMPLATE_ROOT, STATIC_ROOT
    ImportError: cannot import name 'TEMPLATE_ROOT'

From what I can see, paths.py doesn't have that name defined:
https://github.com/QuantStack/voila/blob/master/voila/paths.py

SylvainCorlay:
Fixes #51

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -34,7 +34,7 @@ from jupyter_server.services.config import ConfigManager
 from jupyter_server.base.handlers import FileFindHandler
 from jupyter_core.paths import jupyter_config_path, jupyter_path
 
-from .paths import ROOT, STATIC_ROOT
+from .paths import ROOT, STATIC_ROOT, collect_template_paths
 from .handler import VoilaHandler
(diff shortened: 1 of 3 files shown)


voila/app.py
```diff
@@ -133,52 +133,16 @@ def parse_command_line(self, argv=None):
         self.template_paths = []
         self.static_paths = [self.static_root]
         if self.template:
-            self._collect_template_paths(self.template)
+            collect_template_paths(
+                self.nbconvert_template_paths,
```
maartenbreddels: `nbconvert_template_paths` is already the list of paths, no need to search again. However, I think we **should** move this path searching to the handler, so we can have per notebook or parametrized templates.

Issue #43, opened 2018-12-08
Serve a RISE slideshow


DougRzz:
Enable Voila to serve a [RISE](https://github.com/damianavila/RISE)  slideshow 

This would be a great way of hosting and sharing Jupyter notebook content in a slidey way.  I suggest Voila should directly make use of the RISE slide-type dropdown menus and other parts of the RISE Jupyterlab extension UI.  

@damianavila , what do you think about this proposal?     ref: https://github.com/QuantStack/voila/issues/36#issuecomment-445200786

SylvainCorlay:
Quick note, voila supports special nbconvert templates

See e.g. the example with a screencast template:

![screencap-voila-gridstack](https://user-images.githubusercontent.com/1765949/48071165-eae79b00-e1d9-11e8-86e3-df9901179e8c.gif)

The idea is to encode layouts / extra information in the notebook metadata to 

 - encode layout in the case of the gridstack template (e.g. with the same spec as the deprecated jupyter dashboard)
 - another section for layout in a phosphor dock area layout.
 - another section for rendering as a presentation with rise...

maartenbreddels:
An example of this is in #54, however, I am not sure voila should include this PR. It does show how you could (if you tweak the template further) build a RISE slideshow using the current template system, and even release it on pypi.

jtpio:
Note: there is now a `voila-reveal` template to serve slideshows with Voila: https://github.com/voila-dashboards/voila-reveal

cc @DougRzz

SylvainCorlay:
Closing as addressed by voila-reveal.

PR #54, opened 2019-01-25
Example template: RISE/reveal

maartenbreddels:
I don't think we should merge this, it could actually be a separate package (not now, but when the template system is matured).
This is to see how far we can get using the current template system to implement #43 .
Although this did need a code change, I think we could change the template to work without it. We now need the SlideExporter because it adds a few resources (see https://github.com/jupyter/nbconvert/blob/fb540b8918a45d2b2b85a82b5552f1fac6bba356/nbconvert/exporters/slides.py#L173 ) 
If we hardcode these in the template (which I based on https://github.com/jupyter/nbconvert/blob/fb540b8918a45d2b2b85a82b5552f1fac6bba356/nbconvert/templates/html/slides_reveal.tpl ) then it should work.
Also, this includes all resources in the static directory, the nbconvert server redirects these to a CDN, not sure if we want to implement that as well.


maartenbreddels:
Example usage:
```
$ python -m voila notebooks/rise_example.ipynb --template reveal -strip_sources=False
```

maartenbreddels:
Replaced by https://github.com/voila-dashboards/voila-reveal

PR #60, opened 2019-01-29
WIP: allow progressive and ajax/SPA templates

maartenbreddels:
This PR allows the template system to load progressively, or use ajax to dynamically load the notebook output.
The idea (which I feel is not 100% solid yet, since it feels not that easy to explain / not super intuitive) is to have 2 phase, phase 0 and 1.
 * phase 0 renders voila.html (using the html template system)
 * phase 1 executes the notebook and render the .tpl (using nbconvert)

By default, voila executes both phases (and if voila.html is not present, phase 0 will be effectively skipped).

Progressive loading works the following:
 * phase 0 render the first half of the page, basically `<html>...<body>....`
 * output is flushed to the client, giving instant feedback to the user
 * phase 1 continues, the notebook gets executed, and the second half of the page gets rendered, basically `<script that removes spinner> .... </body></html>`

A SPA (single page app) that would like to use ajax, could use the following strategy.
 * indicate in conf.json that only phase 0 should be executed (by default)
 * phase 0 renders the full page and includes some JS to request the same page again, with argument `phase=1`
 * phase 1 gets executed, returned to the page, where custom code can insert the html and execute all the script tags.

An example of how it looks using the 'progressive' template:
```
$ python -m voila notebooks/basics.ipynb --template progressive
```
template-progressive
(using ajax kind looks the same)

There is quite some duplication of template snippets, although I am not sure we can nicely fix this without creating confusion. Also, I don't like the phase=0/1 naming; we can think about a different name, or have different tornado handlers for each phase.

I do like that it is becoming quite flexible. I can see, with this PR, that one could create a single page dashboard that could load different notebooks without a page reload for instance.


SylvainCorlay:
As discussed in person, closing in favor of #133. (Will reconsider later).

PR #69, opened 2019-02-20
Fixes for templates, add support for tornado settings

timkpaine:
- Fix for hardcoded `default` template
~~- Adding in CSP endpoint~~ moved to #89 
~~- Exposing tornado settings, coupled with CSP endpoint allows you to put voila behind iframe~~

~~As an example, embed voila in an iframe and set~~

```python
v = Voila()
v.tornado_settings = {'headers': {'Content-Security-Policy': "frame-ancestors 'self' localhost:*"}}
```

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -20,7 +20,7 @@ import tornado.ioloop
 import tornado.web
 
 from traitlets.config.application import Application
-from traitlets import Unicode, Integer, Bool, List, default
+from traitlets import Unicode, Integer, Bool, Dict, List, default
 
 from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
 from jupyter_server.services.kernels.handlers import KernelHandler, ZMQChannelsHandler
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Hi Tim,

looking good, the template_name workaround should not have been necessary, it actually exposes a bug. We can consider it a workaround and keep it in for now.
Could you rename `extra_tornado_settings` to `tornado_settings` so it is consistent with the classical notebook?

timkpaine:
Will do!

timkpaine:
moved the CSP stuff to a standalone PR in https://github.com/QuantStack/voila/pull/89

timkpaine:
@maartenbreddels rebased so it should be cleaner to see

maartenbreddels:
Looks good tim! Thanks.
I had some issue testing this from the command line due to nested quotes,  the magic line (for bash/zsh) is:
```
voila notebooks/basics.ipynb --autoreload=True --Voila.tornado_settings=$'{"headers":{"Content-Security-Policy":"frame-ancestors \'*\' localhost:*; report-uri /api/security/csp-report" }}'
```

For further reference, this is needed if you want to tell a browser it is ok that voila is rendered in an iframe from a different domain (its parent), otherwise they will not allow it.

Issue #70, opened 2019-02-26
Loading screen while widgets load


timkpaine:
There is a brief period when we just show the empty white box while the widgets load/python executes. We should think of a good loading screen and find a way to hook it in. Initial thought is wait for `kernel-ready`, but its conceivable users could program an infinite loop in their last cell.

maartenbreddels:
This is demonstrated in #60, but I will split off the progressive part. It works by first sending the voila.html template (a regular jinja template), flush it to the browser (so it can do progressive rendering). After this, the kernel is started, the notebook executed and the nbconvert template rendered and send to the browser.
Sounds good?

SylvainCorlay:
Closing as subperseded by discussions on progressive rendering.

PR #71, opened 2019-02-28
allow progressive loading/templates

maartenbreddels:
Before rendering/server the voila.tpl (nbconvert) template, we (optionally) send an html page (voila.html template) to allow progressive loading. A template can show a spinner (example below), or already show a portion of the UI.

This new template is basically a split of voila.tpl into voila.html (the header and a loading screen), and voila.tpl (the rest of the body, and the end tags).

voila-progresssive

This solves #70 and is a subset of #60.


SylvainCorlay:
As discussed in person, closing in favor of #133. (Will reconsider later).

PR #75, opened 2019-03-07
Blacklisting nbextensions and workaround for QGrid

maartenbreddels:
Since we load extensions that are expecting a notebook environment (and cannot distinguish between widgets extensions or other extensions) some will try to load modules and give ugly errors.
image

Apart from the `jupyter-js-widgets/extension` there are some other extensions that make no sense at all, I've blacklisted them.

Some extensions may actually work, like QGrid as long as a module is defined. (see also https://github.com/QuantStack/voila/issues/72 )

### DIFF
--- a/share/jupyter/voila/template/default/nbconvert_templates/base.tpl
+++ b/share/jupyter/voila/template/default/nbconvert_templates/base.tpl
@@ -39,8 +39,15 @@
     integrity="sha256-Ae2Vz/4ePdIu6ZyI/5ZGsYnb+m0JlOmKPjt6XZ9JJkA="
     crossorigin="anonymous">
 </script>
+
 <script>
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
As discussed in person, this may not be a good way forward. Although we should probably work with the Qgrid folks and suggest they make two notebook extensions, one for the widget only, and one using the class notebook APIs. We would only be using the former in voila naturally.

jtpio:
Can this be closed?

maartenbreddels:
Yes, if we still want to a have a solution this should be something configurable (and we could ship a default blacklist).

jtpio:
qgrid should be handled properly (voila 0.1.9): https://mybinder.org/v2/gist/jtpio/6016a5a2e0a184aaa11543e3a9da292f/master?urlpath=%2Fvoila%2Frender%2Fqgrid.ipynb

Issue #77, opened 2019-03-11
generated response includes 12k lines of css


jf---:
I noticed that when hosting a notebook with voila, that in the initial response, 12k lines of `bootstrap css` is included. While most `js` code is cached, this initial response is not and takes little under 3.00 seconds to complete. About 95% of the response data is the `css` code. It seems low hanging fruit to just link the dependency ( as is done for font-awesome ) or is served as a separate file, such that the browser can perform caching.

this applies to the template:

`voila/share/jupyter/voila/template/default/nbconvert_templates/voila.tpl`

here the [`css code`](https://github.com/QuantStack/voila/blob/340cd65776ac75cf864c0f219de3f696be5c28c7/share/jupyter/voila/template/default/nbconvert_templates/voila.tpl#L6) is inlined in the template 

whereas

`voila/share/jupyter/voila/template/gridstack/nbconvert_templates/voila.tpl`

[inlines](https://github.com/QuantStack/voila/blob/340cd65776ac75cf864c0f219de3f696be5c28c7/share/jupyter/voila/template/gridstack/nbconvert_templates/voila.tpl#L4) the bootstrap css

This reduces the response considerable...

maartenbreddels:
I am not sure #78 is the right solution, but I indeed think we should not inline all the css this way, I'll give it some thought.

jf---:
> I am not sure #78 is the right solution

Agreed, personally I much prefer your #60 approach, closing the PR, leaving the issue open.

SylvainCorlay:
Now that we are loading CSS sepatetely (in master), I think that we can close this specific issue.

PR #78, opened 2019-03-11
fixes issue #77

jf---:
inlined bootstrap is now linked rather than inlined in initial response


jf---:
See #77

jf---:
closing, #60 resolves the issue and is a more elegant approach

PR #79, opened 2019-03-12
Expose __version__

jtpio:
Expose `__version__`, which can be useful in the context of a notebook:

image


maartenbreddels:
Thanks Jeremy!

PR #80, opened 2019-03-13
Cull idle kernels

jtpio:
Expose `cull_idle_timeout` and `cull_interval` to be able to cull idle kernels when using voila as a standalone app. `cull_busy` and `cull_connected` shouldn't be needed (for now) since there is one kernel started at every page load.

This is especially useful for long running voila instances to be able to free resources.

Usage: `python -m voila --Voila.cull_idle_timeout=10 --Voila.cull_interval=10`

Potential fix for https://github.com/QuantStack/voila/issues/55

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -101,6 +101,16 @@ class Voila(Application):
         )
     )
 
+    cull_idle_timeout = Integer(0, config=True,
+        help="""Timeout (in seconds) after which a kernel is considered idle and ready to be culled.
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I didn't know this existed, by I think it's a useful thing to configure for voila.

I'm working on a PR to use the Configurable machinery of traitlets to help out instead.

jtpio:
https://github.com/QuantStack/voila/pull/84 looks really good and is more generic.

Let's close this PR then.

PR #89, opened 2019-03-15
Support embedding voila in an iframe by adding the CSPReportHandler

timkpaine:


timkpaine:
Without the route you shouldn't be able to put a voila app inside an iframe at all

timkpaine:
Also let me squash these down into one commit

maartenbreddels:
I did the following:
 * create index.html with the following:
```html
<html>
    <body>
        <iframe src='http://localhost:8866' style="width: 100%; height: 100%"></iframe>
    </body>
</html>
```
* run a http server
```
python -m http.server 8008
```
* start voila using `voila notebooks/basics.ipynb --autoreload=True --Voila.tornado_settings=$'{"headers":{"Content-Security-Policy":"frame-ancestors localhost:*;" }}'`

* `open 'http://localhost:8008/embed.html'`
This shows me the voila page, and no access to `/api/security/csp-report` is made. This only happens when the CSP header is missing.

Maybe I am missing something and this is really required to be in voila, but I fail to understand why it is there if not really needed.

timkpaine:
Hm if I run without this patch I see no voila output and this:
Screen Shot 2019-04-04 at 10 39 00 AM

timkpaine:
let me try to make a standalone example like you did to see if it might be something else

timkpaine:
@maartenbreddels looks like the CSPReportHandler might just be a failover if the response headers are disallowing access? Not sure tbh but it looks like it may be fine with just the `tornado_settings` flags. Going to close this, it was only 2 lines anyway so its easy to add if we need it at a later date.

Issue #85, opened 2019-03-15
widgets interact not presenting a slider


afonit:
```
# Name                    Version                   Build  Channel
voila                     0.0.9                      py_0    conda-forge
```

I did see this:
https://github.com/QuantStack/voila/issues/19

And I am able to get widgets working using the pattern shown in that report.

However when I use widgets in other patterns like the below they don't work:

```python
import ipywidgets as widgets
from ipywidgets import interact

def what(item):
    return item*5
interact(what, item=widgets.IntSlider(value=2, min=1, max=10))
```

Here is what it looks like with voila:
![image](https://user-images.githubusercontent.com/1798497/54455123-61b91580-4731-11e9-9952-ce6afdd932c6.png)

Here is the chrome console output:
![image](https://user-images.githubusercontent.com/1798497/54455185-8a410f80-4731-11e9-903e-6b2bfdb39965.png)

-----------------
Here is what the jupyter widget looks like in the regular notebook:
![image](https://user-images.githubusercontent.com/1798497/54455044-37675800-4731-11e9-9c59-80ab3b26a922.png)

maartenbreddels:
I've been digging into this, and the source of the problem is that nbconvert gets a [clear_output](https://github.com/jupyter/nbconvert/blob/0ef92837822da09821b96955074002726b08a99f/nbconvert/preprocessors/execute.py#L478) after the output msg. Since voila uses nbconvert, we have the same issue.
To see the issue, put the code in a notebook, and execute:
```
python -m nbconvert ~/src/QuantStack/voila/issue/85.ipynb --to ipynb --log-level=DEBUG --execute --stdout
```
I see the output as an empty list. It was also noted by @SylvainCorlay in item 3 at #19

maartenbreddels:
https://github.com/jupyter/nbconvert/pull/969 partly solves the issue, I still have difficulty getting the output to show. TBC.

maartenbreddels:
This should be solved by https://github.com/QuantStack/voila/pull/91
using master voila and master nbconvert it should work, for testing it needs the branch that is used in .travis.yml

psychemedia:
Is this fix going to be released any time soon? I think the current pip version is still showing this issue?

afeiszli:
Just tried out voila and had the same issue with interact. Installed using pip.

leogout:
Hello, I have the same problem, to fix this I had to install Jupyterlab + Ipywidgets in my project directory:
```
[...]

PR #91, opened 2019-03-21
Implement Output widget at nbconvert level for interact

maartenbreddels:
As mentioned in #19 and #85 there are some issues with interact. Interact(ion) uses the Output widget a lot, which relies on the frontend to handle display and clear_output differently. 

A part of the fix is in https://github.com/jupyter/nbconvert/pull/969 which deals with `wait=True`, but this only solves part of the issue.

When nbconvert is used to execute the notebook, there is no OutputWidget that can hook into the display mechanism to redirect display and clear_output calls. This custom executor creates a 'fake' Output widget that mirrors this behaviour.

This is to show that it is possible to support interact with voila, I am not sure this is the best solution. Other solution are:
 1. Implement display/clear_output redirects kernel side, in Python this is simply monkey patching display and clear_output and possibly stdout/stderr, but I am not sure this will work with with xeus-python, and similar solutions would be needed for all the other kernels.
 2. Do no use nbconvert for the execution, only start executing once a frontend is connected.

I think the core problem is that interact(ion) relies on the Output widget being connected to the frontend. We could solve some problems by actually modifying `Output.ouputs` at the kernel side, and never invoke display/clear_output, but directly invoke methods on the Output widget.

If we want to go forward with this approach (which would be the most compatible one), this requires a few more tweaks and a few tests (wait=True for the output widget it no supported)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ before_install:
   - conda info -a
   - conda create -q -n test-environment -c conda-forge python=$PYTHON_VERSION jupyter_server pytest==3.10.1 pytest-cov nodejs flake8
   - source activate test-environment
+  # we need this branch to perform testing
+  - pip install --upgrade --no-deps --force-reinstall git+https://github.com/maartenbreddels/nbconvert.git@widget_state#egg=nbconvert
 install:
   - pip install ".[test]"
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ before_install:
   - conda info -a
   - conda create -q -n test-environment -c conda-forge python=$PYTHON_VERSION jupyter_server pytest==3.10.1 pytest-cov nodejs flake8 ipywidgets
   - source activate test-environment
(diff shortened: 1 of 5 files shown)


maartenbreddels:
While testing found a bug in ipywidgets:
https://github.com/jupyter-widgets/ipywidgets/pull/2351

display_id in the Output widget is not supported yet, and we could do a refactor to do code deduplication with nbconvert's ExecutePreprocessor.

To perform testing we need to use https://github.com/jupyter/nbconvert/pull/900, coverage for execute.py is 100%.

enryH:
Could you provide an example? Using
```python
from ipywidgets import interact

def foo(x):
    print(x)

interact(foo, x=(0, 10))
```
yields as output only
```
5
```
without the slider.

SylvainCorlay:
@enryH this appears to work. Are you talking about voila or nbconvert?

enryH:
voila. 

> @enryH this appears to work. Are you talking about voila or nbconvert?

voila

Issue #94, opened 2019-03-26
No chart output for plotlywidget example


timkpaine:
Not sure where to start with a fix for this one, any pointers? I'm trying to run this example:
https://plot.ly/python/figurewidget-app/

maartenbreddels:
Assuming this is solved by #97, if not, please reopen.

PR #97, opened 2019-04-04
Fix displayed not triggered

maartenbreddels:
The backbone 'displayed' event was never triggered. This caused some widgets, for instance ipysheet, not to be shown. See https://github.com/QuantStack/voila/issues/87#issuecomment-479570714


timkpaine:
@maartenbreddels just checked out locally I think this fixes perspective, plotlywidget #94, and qgrid???
Screen Shot 2019-04-04 at 9 21 05 AM

maartenbreddels:
Great, I assume you also needed the workaround in https://github.com/QuantStack/voila/pull/75/files to get qgrid working?

timkpaine:
@maartenbreddels not unless its also on your `fix_displayed_not_triggered` branch. I just checked out your branch and ran and these 3 work perfectly.

maartenbreddels:
Ok, so this branch + that qgrid fix right?

Issue #16, opened 2018-11-20
Rendering local images in markdown


SylvainCorlay:
At the moment, images in relative directories w.r.t. notebooks are not rendered.

Serving these directories would break the security model of voila, which is not supposed to expose the actual notebook content.

This raises the question of how deal with images. Could nbconvert explicitly render inline images?

freddupont:
hello nice work on this project very impressive and potentially very useful!
In my developpement patern I copy past lots of picture in the markdown cell from PDF viewver to jupyter
When I tryed a notebook with multiple image recorded inside the notebook, the display by voila server was that each image was the same (always the first one).
I don't know if this is a bug or a feature request

freddupont:
You will find an example notebook with 2 pictures  one blue, one yellow, its render correctly on jupyter, not on voila

https://gist.github.com/freddupont/ab5df4c5a26b2efe5152f8b72a16fe28

maartenbreddels:
I didn't know of this functionality actually, and I can confirm that Jupyter lab and the classical notebook render it correctly, but nbconvert does not. Since we rely on nbconvert, we inherit this issue.

SylvainCorlay:
Same, I did not know of this feature.

maartenbreddels:
It seems the issue is fixable for nbconvert, hopefully, we can get this in the next release. I also noticed Jupyter lab is missing this feature (I could not paste an image), but maybe master has it.

maartenbreddels:
Digging into this a bit, I think we can have a customized markdown2html filter, which instead of putting in the image src, embed the image content. 
@SylvainCorlayI think what you suggested initially (render inline image) is probably the best approach.
I think we can do this with minimal code, and possible could push this into nbconvert.

SylvainCorlay:
> Digging into this a bit, I think we can have a customized markdown2html filter, which instead of putting in the image src, embed the image content.

This is really great news!

SylvainCorlay:
@freddupont  this is fixed in master (and requires nbconvert master).

Your example notebook actually gave @maartenbreddels some inspiration to fix the issue!

Thanks to everyone involved.

PR #101, opened 2019-04-05
new: render image inline in html

maartenbreddels:
Based on https://github.com/jupyter/nbconvert/pull/980 
this allows us to render images inline, which would solve #16 
Attachment (with mime bundles) and local files work (which are retrieved using the content manager)
image

Since we depend on the content manager to be used, we may not be able to get this into nbconvert, unless we put the file retrieval in a separate method.

NOTE: should not be merged unless https://github.com/jupyter/nbconvert/pull/980 gets merged.


SylvainCorlay:
Yeah!

maartenbreddels:
🎉 now it feels like weekend :)

PR #106, opened 2019-04-10
Fix undefined manager

jtpio:
`WidgetManager` takes a manager as a second argument which was not defined in this case, resulting in a: `ReferenceError: manager is not defined`

Tested with:

```python
from ipywidgets import IntSlider, Label, Output

out = Output()
slider = IntSlider(min=0, max=100)
label = Label(value='Test')

@out.capture()
def on_change(change):
    display(label)

slider.observe(on_change, names='value')

with out:
    display(slider)    
out
```

Although this change seems to fix the `manager is not defined` error, the code posted above still doesn't work as expected (need more investigation).

### DIFF
--- a/js/src/manager.js
+++ b/js/src/manager.js
@@ -28,8 +28,15 @@ export class WidgetManager extends JupyterLabManager {
 
     constructor(kernel) {
         const context = createContext(kernel);
-        const rendermime = createRenderMimeRegistry();
+        const rendermime = new RenderMimeRegistry({
+            initialFactories: standardRendererFactories
+        });
         super(context, rendermime);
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Sorry this fell between the cracks. Will investigate this week.

SylvainCorlay:
(rebased)

SylvainCorlay:
The issue with displaying widgets in an output widget appears to be unrelated to this (but related to the issue people experienced with `ipympl` at the workshop).

Will be merging this and iterate on the output widget.

jtpio:
Sounds good, seems to be a different issue indeed.

PR #109, opened 2019-04-12
Add data-base-url to the default template body

jtpio:
Many classic notebook extensions use the `data-base-url`attribute from the body to load static assets.

[For example with `ipyleaflet`](https://github.com/jupyter-widgets/ipyleaflet/blob/afa4d6ea290661fc4e78ef5683293e5e8053be6f/js/src/notebook.js#L2):

```javascript
__webpack_public_path__ = document.querySelector('body').getAttribute('data-base-url') + 'nbextensions/jupyter-leaflet/';
```

This change adds the `data-base-url` attribute to the body of the default template, so the attribute can correctly be resolved.

### Example with `ipyleaflet`

Using the example from https://ipyleaflet.readthedocs.io/en/latest/api_reference/marker.html

#### Before

image

The marker is not loaded. Dev tools log:

```
GET http://localhost:8866/voila/render/notebooks/nullnbextensions/jupyter-leaflet/2273e3d8ad9264b7daa5bdbf8e6b47f8.png 404 (Not Found)
```

#### After

image

### DIFF
--- a/share/jupyter/voila/template/default/nbconvert_templates/voila.tpl
+++ b/share/jupyter/voila/template/default/nbconvert_templates/voila.tpl
@@ -53,7 +53,7 @@ div#notebook-container{
 {%- endblock html_head_css -%}
 
 {% block body %}
-<body>
+<body data-base-url="{{resources.base_url}}voila/">
   <div tabindex="-1" id="notebook" class="border-box-sizing">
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I agree, @mariobuikhuizen just had an issue with that as well, for ipyvuetify we need to load some resources as well.

PR #115, opened 2019-04-14
Reverse proxy

timkpaine:
Here's the use case:

I want to host multiple voila apps and dynamically spawn them. Rather than using my current strategy of launching against a unique port bound to 0.0.0.0 and using an iframe, i want to reverse proxy through my main app. The app will serve notebooks on `/voila/<notebook name>` and proxy through to voila instances running against local host.

Using this code, the server will set `server_url` to `/`, so that the server side hosts routes in the usual way, but `base_url` will be set to `/voila/<notebook name>` so client requests get proxied to the right voila instance.

Previously these were coupled, meaning if you set base_url to that path, the client requests would go there, but the server would also serve at that path (meaning the actual path it expected would be `/voila/notebook/voila/notebook`)

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -101,7 +101,9 @@ class Voila(Application):
         'strip_sources': 'VoilaConfiguration.strip_sources',
         'autoreload': 'Voila.autoreload',
         'template': 'VoilaConfiguration.template',
-        'theme': 'VoilaConfiguration.theme'
+        'theme': 'VoilaConfiguration.theme',
+        'base_url': 'Voila.base_url',
+        'server_url': 'Voila.server_url',
     }
     connection_dir_root = Unicode(
         config=True,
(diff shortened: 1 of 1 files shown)


timkpaine:
@maartenbreddels @SylvainCorlay any ideas on why i'd have no widgets but no errors? 
I'm proxying the http and websocket requests:
[proxy handlers for http and ws](https://github.com/timkpaine/dashboard/blob/master/dashboard/proxy_handler.py)
[routes](https://github.com/timkpaine/dashboard/blob/master/dashboard/app.py#L131)



image

timkpaine:
nvm i think i'm dumb, i send the proxy websocket into an infinite loop. it should be `self.write` not `self.ws.write`

timkpaine:
yessssss
tmp

GrahamDumpleton:
What needs to be done to get this completed and merged?

Is there any other way in the interim of being able to set `base_url` when using `voila` from command line, and not as a server extension?

maartenbreddels:
Thanks for your patience, we had to rebase and review. Thanks for your contributions!

SylvainCorlay:
Thanks!

timkpaine:
I've updated this to work on prod:
https://github.com/timkpaine/dashboard/commit/8c5d39c1658d257a72e78170b18ff86487bbf9bf

timkpaine:
Screen Shot 2019-05-28 at 11 37 08 AM

@SylvainCorlay @maartenbreddels @vidartf if you need any examples for the conf

SylvainCorlay:
@timkpaine this is a really nice demo!

I think we now have a really nice foundation for the dashboarding workshop next week.

PR #119, opened 2019-04-17
Better connection string log

timkpaine:
Depending on whether `tornado_settings` is setup for SSL will print
Screen Shot 2019-04-17 at 1 39 37 PM
or
Screen Shot 2019-04-17 at 1 39 42 PM

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -311,7 +311,11 @@ class Voila(Application):
 
     def listen(self):
         self.app.listen(self.port)
-        self.log.info('Voila listening on port %s.' % self.port)
+        if self.tornado_settings.get('ssl_options'):
+            host = 'https://0.0.0.0:%s' % self.port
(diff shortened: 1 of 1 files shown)


choldgraf:
This would make my life better by a small but significant amount :-)

SylvainCorlay:
@timkpaine this appears to need a rebase now that we open a web browser. Would you like to rebase? I am happy to do it otherwise.

maartenbreddels:
I think the way forward is to open a PR with https support and uncomment https://github.com/QuantStack/voila/blob/a66eb1fcf619da0b3a3cf39d5cd2c3261f51027a/voila/app.py#L226 so we get this for free.
Do you agree @timkpaine ?

timkpaine:
Yeah this is encompassed by the other PR

PR #122, opened 2019-04-17
Fix for phosphor and allow extension pass through for custom CDNs, nonstandard paths, etc

timkpaine:
image


timkpaine:
@maartenbreddels @SylvainCorlay mind giving this a look, should make voila play nicer with extensions that are closely tethered to phosphor.

timkpaine:
@SylvainCorlay @maartenbreddels can you guys at least cherry-pick the phosphor define, currently voila is completely broken for any widgets that use phosphor

PR #131, opened 2019-05-11
Introduce new JupyterLab template

SylvainCorlay:
Including the new jupyterlab template from https://github.com/jupyter/nbconvert/pull/1005 into voila.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,24 @@
+from __future__ import print_function
+
 from setuptools import setup, find_packages, Command
 from setuptools.command.sdist import sdist
 from setuptools.command.build_py import build_py
+from setuptools.command.develop import develop
 from setuptools.command.egg_info import egg_info
+from setuptools.command.bdist_egg import bdist_egg
 
 from subprocess import check_call
 
 import os
 import sys
 
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib import urlopen
+
 from distutils import log
+
 log.set_verbosity(log.DEBUG)
 
 here = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -119,6 +129,108 @@ def run(self):
...
+        theme_light_dest = os.path.join('voila', 'resources', 'theme-light.css')
+        theme_dark_dest = os.path.join('voila', 'resources', 'theme-dark.css')
+        print('bad')
+        try:
+            print('_download')
+            css = self._download(css_url)
+            print('after')
+            theme_light = self._download(theme_light_url)
+            theme_dark = self._download(theme_dark_url)
+            print('after after')
```
maartenbreddels: remove debug prints?

SylvainCorlay:
> Look good! However, I'd really like to see the css go to files in the template/default/static directory.

The CSS files are now in the static directory!

maartenbreddels:
Yes, html looks much cleaner. I don't understand why we need the css preprocessor. Can't we drop that and put that css in a pygments.css and serve that in the static folder as well?

SylvainCorlay:
In the case of nbconvert, the reason for inlining is that the CSS is computed on the fly with pygments, and you can change the style programmatically.

I put the generated CSS in a string itself because using CSS variables in a pygments themes only works with a PR in pygments that made it to master but not yet in a release.

The generated CSS is short anyways.

maartenbreddels:
Ok, so once https://github.com/jupyter/nbconvert/pull/1005 is released, we can get rid of css preprocessor right? In that case I'm happy if we merge this, and I can rebase #133

SylvainCorlay:
> Ok, so once jupyter/nbconvert#1005 is released, we can get rid of css preprocessor right? In that case I'm happy if we merge this, and I can rebase #133

That is not what I meant, I think that we will still need a css preprocessor, unless we make the upstream one flexible enough to accommodate our use case.

Great if we can merge this and move forward!

Issue #111, opened 2019-04-13
Include `localhost:<port>` somewhere in the startup


choldgraf:
This is a minor annoyance, but it would be great if running `voila` also did:

1. Try to auto-open a browser at the port that's exposing voila
2. Print `localhost:<port-number>` in the output message just to make it easier to copypaste or ctrl+click and launch

SylvainCorlay:
Fixed!

PR #136, opened 2019-05-14
Usability: Open browser, display url and shortcut to serving directory

maartenbreddels:
To make voila easier to use, we now open the browser (this is done analogous to the notebook as close as possible). To avoild opening the browser, pass `--no-browser`.

Also, analogous to the notebook, it now shows the url it is running at:
```
...
[Voila] Voila is running at:
http://localhost:8866/
```

This can be customized, using `--Voila.custom_display_url='foobar'` and `--Voila.ip=''` for instance, again similar to the classical notebook.

A third improvement is hosting directories. Previously this was done using `--Voila.root_dir='/path/to'` and no giving extra arguments. With this PR, if the passed argument is a directory, it wil set the root dir.

Future: I've left in some commented out code for the tokens and SSL, I don't think we should attack that now.

### DIFF
--- a/share/jupyter/voila/template/default/templates/browser-open.html
+++ b/share/jupyter/voila/template/default/templates/browser-open.html
@@ -1,0 +1,18 @@
+{# This template is not served, but written as a file to open in the browser,
+   passing the token without putting it in a command-line argument. #}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="refresh" content="1;url={{ open_url }}" />
+    <title>Opening voila</title>
+</head>
+<body>
+
+<p>
+    This page should redirect you to voila. If it doesn't,
+    <a href="{{ open_url }}">click here to go to voila</a>.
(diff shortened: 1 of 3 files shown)


timkpaine:
This one has ssl, I think it's good to have otherwise why have a link at all if it won't work sometimes:https://github.com/QuantStack/voila/pull/119

SylvainCorlay:
:+1:

maartenbreddels:
Cool, happy to put in the SSL. How do you configure voila to use https btw? Because we don't support the certfile yet, I didn't think support for it make sense.

timkpaine:
We don't yet either but how much more do you have to do manually vs passing the settings through to tornado?

maartenbreddels:
I don't want to put too much in this PR. If we add in SSL we can simply uncomment https://github.com/QuantStack/voila/pull/136/files#diff-024301d95616622c19f49b8cd78d33c3R224 and have https as protocol.

maartenbreddels:
This closes #111

PR #137, opened 2019-05-14
Give an error when a template is not found

### LINKED ISSUE #129: Throw an error when a custom template is given but not found
I found myself confused recently when I was trying to demo creating a custom template in #126 . I'd run Voila with a custom template like `--template=mytemplate`, and everything would run fine, but the final output was not the HTML I expected.

Right now, this is because Voila simply appends the custom template path to the list of templates it expects. For me, this led to confusion because I was expecting Voila to error if it didn't find my template path. Instead, it just reverted to the default template.

What do folks think about raising an error in this case, rather than reverting to default? I feel like if somebody gives a custom template, they'd rather know that the path wasn't correctly found, rather than have it silently work, but not in the way expected.

maartenbreddels:
Closes #129 
Example:
```bash
$ voila notebooks/basics.ipynb  --template=bla   
  File "/Users/maartenbreddels/miniconda3/envs/voila/bin/voila", line 11, in <module>
    load_entry_point('voila', 'console_scripts', 'voila')()
  File "/Users/maartenbreddels/miniconda3/envs/voila/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "/Users/maartenbreddels/src/QuantStack/voila/voila/app.py", line 294, in initialize
    self.setup_template_dirs()
  File "/Users/maartenbreddels/src/QuantStack/voila/voila/app.py", line 303, in setup_template_dirs
    self.template)
  File "/Users/maartenbreddels/src/QuantStack/voila/voila/paths.py", line 78, in collect_template_paths
    raise ValueError('No template sub-directory with name %r found in the following paths:\n\t%s' % (template_name, paths))
ValueError: No template sub-directory with name 'bla' found in the following paths:
	/Users/maartenbreddels/src/QuantStack/voila/share/jupyter/voila/template
	/Users/maartenbreddels/Library/Jupyter/voila/template
	/Users/maartenbreddels/miniconda3/envs/voila/share/jupyter/voila/template
	/usr/local/share/jupyter/voila/template
	/usr/share/jupyter/voila/template
```

### DIFF
--- a/voila/paths.py
+++ b/voila/paths.py
@@ -36,14 +36,17 @@ def collect_template_paths(
 
     # We look at the usual jupyter locations, and for development purposes also
     # relative to the package directory (with highest precedence)
-    template_directories = \
-        [os.path.abspath(os.path.join(ROOT, '..', 'share', 'jupyter', 'voila', 'template', template_name))] +\
-        jupyter_path('voila', 'template', template_name)
+    search_directories = \
+        [os.path.abspath(os.path.join(ROOT, '..', 'share', 'jupyter', 'voila', 'template'))] +\
+        jupyter_path('voila', 'template')
 
-    for dirname in template_directories:
-        if os.path.exists(dirname):
+    found_at_least_one = False
+    for search_directory in search_directories:
+        template_directory = os.path.join(search_directory, template_name)
+        if os.path.exists(template_directory):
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
This looks good to me.

PR #144, opened 2019-05-20
 fix(build): add missing presets for babel

mariobuikhuizen:
In commit e6d4504 babel-loader was replaced, thereby removing the presets. Without presets babel will not transform and polyfill the source code.

Commit e6d4504 also creates a directory "src" which is ignored in .gitignore and it introduces a build folder "lib" which should be ignored in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ __pycache__
 \#*#
 .#*
 .coverage
-src
 
 *.swp
 *.map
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Great, I forgot to commit `babel.config.js`, and babel didn't complain about that. It now actually transpiles :)

I must say, given that typescript does not help with polyfills nor does it want to support it https://github.com/Microsoft/TypeScript/issues/20095 I'm giving babel a +1 over ts.

Issue #145, opened 2019-05-22
Split js bundles


maartenbreddels:
Working on #133 and [ipyvuetify](https://github.com/mariobuikhuizen/ipyvuetify) we're realizing (me and @mariobuikhuizen) that we need to split the massive 15mb bundle into, and start thinking a bit on a JS API:
 * A bundle that provides the standard way of starting the kernel, based on PageConfig. This can be called when the kernel id is known in the front end.
 * A bundle that provides the widget manager, without the `@jupyter-wdigets/controls` package (which bloats the package) since the controls package may not always be needed (eg pure bqplot or ipyvolume output, or ipyvuetify). 
 * A bundle that provides the `@jupyter-wdigets/controls` package.

Template creators may want to remake one or more of those bundles.

SylvainCorlay:
> ~15MB~ 5MB

That is not *too* bad :)

maartenbreddels:
For slow connections, it is (mobile for instance).

But the main point is that we can already do all the work (connecting to the kernel, getting all widget data in) before that big bundle loads.

Now we first get the 5mb, then it will connect to the kernel, and start getting in the widget data.

PR #147, opened 2019-05-22
Refactor voila js

maartenbreddels:
This cleans up some artifacts of the beginning of voila:
 * bundle is now called voila.js
   * placed in share/jupyter/voila/template/default/static
   * exports WidgetManager and connectKernel
 * main.js replaces WidgetApplication, so it's easier to make change WidgetManager, or the way/moment the kernel is connected to, instead of having to write a new bundle.
 * sidefx of the new main.js is that https://github.com/QuantStack/voila/issues/55 seems to be solved.
 * other resources are also moved to template's static folder


maartenbreddels:
Notes:
 * this attacks #145 partly, in that we now have a bit of an API making it easier to have radical changes in templates
 * We should do a careful check in a new environment if the distribution packages everything up before we make a new release.

SylvainCorlay:
Thanks. This looks great and is much cleaner now. I like how it makes (almost) all javascript assets data files instead of package data.

PR #148, opened 2019-05-22
feature(configuration): server extension shares configuration with voila app

maartenbreddels:
Before, only `Voila` was configurable, but options should also be configurable for the server extension. These common options/traits should now go into `VoilaConfiguration` (spelled out completely to avoid confusion with the traitlet `Config` class).
Examples:
```
$ jupyter server --VoilaConfiguration.template=default --VoilaConfiguration.theme=dark
$ jupyter notebook --no-browser --VoilaConfiguration.template=gridstack
```
For voila, the `--template` is an alias, so these two are equivalent
```
$ voila . --no-browser --VoilaConfiguration.template=gridstack
$ voila . --no-browser --template=gridstack
```

Docs should reflect this (maybe @choldgraf wants to be involved here), we could do this in a separate PR.
# TODO
 - [x] `strip_sources` should move to this class
 - [x] more tests

### DIFF
--- a/tests/app/config_paths_test.py
+++ b/tests/app/config_paths_test.py
@@ -14,7 +14,7 @@ def voila_config_file_paths_arg():
 
 
 def test_config_app(voila_app):
-    assert voila_app.template == 'gridstack'
+    assert voila_app.voila_configuration.template == 'gridstack'
 
 
 def test_config_kernel_manager(voila_app):
--- a/tests/app/execute_test.py
+++ b/tests/app/execute_test.py
@@ -17,6 +17,7 @@ def test_hello_world(http_client, base_url):
     assert response.code == 200
     html_text = response.body.decode('utf-8')
     assert 'Hi Voila' in html_text
+    assert 'print' not in html_text, 'by default the source code should be stripped'
     assert 'gridstack.css' not in html_text, "gridstack should not be the default"
 
 
--- a/tests/app/no_strip_sources_test.py
+++ b/tests/app/no_strip_sources_test.py
@@ -1,0 +1,20 @@
+import pytest
+import json
+import os
+
+
+BASE_DIR = os.path.dirname(__file__)
+
+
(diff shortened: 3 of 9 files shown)


choldgraf:
Just to clarify, this wouldn't change the user-facing documentation currently there, right? It's not an API change, but it exposes more of the configuration to users?

Agree that it should be documented...maybe specifically about the extra configuration options? Or maybe this is something that we could tackle easily w/ autodoc

maartenbreddels:
No, it does not invalidate the docs, since `--template=gridstack` is still valid.
Undocumented was that `--template=gridstack` was an alias for `--Voila.template=gridstack`, and is changed to `--VoilaConfiguration.template=gridstack`, which now can also be passed to jupyter [server|notebook|lab].

choldgraf:
Ah that's awesome, so you can configure Voila as a server extension when you initialize the jupyter server - that makes sense now. We should definitely document that (I'm fine if we open a separate issue about it so it doesn't block this PR)

SylvainCorlay:
I have been experimenting with this, and this looks fine.

I want to look at what @zsailer has been doing in JupyterServer wrt server extensions before merging.

SylvainCorlay:
@maartenbreddels one small regret is that we might have wanted to have the syntax

```
jupyter server --Voila.theme=dark
```

instead of

```
jupyter server --VoilaConfiguration.theme=dark
```

choldgraf:
@SylvainCorlay is it too late for that? I agree that's a more user-friendly/intuitive API

SylvainCorlay:
@choldgraf I think it should be possible but I am not sure of the cleanest way to do this. (`Voila` is a bit of an important name for a configuration util class, so it should probably be achieved with some traitlets trick).

maartenbreddels:
I agree, I had the idea of renaming voila to VoilaApp to enable this

(from mobile phone)

PR #153, opened 2019-05-24
chore(deps,js): move to jupyterlab-manager 0.40

maartenbreddels:
Change in https://github.com/jupyter-widgets/ipywidgets/pull/2265 caused voila to break, as noted by @SylvainCorlay.
In order to test if https://github.com/jupyter-widgets/ipywidgets/pull/2418 works in voila, we first need to move to manager 0.40.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -8,7 +8,7 @@
     "@jupyter-widgets/base": "^1.2.2",
     "@jupyter-widgets/controls": "^1.4.2",
     "@jupyter-widgets/html-manager": "^0.15.2",
-    "@jupyter-widgets/jupyterlab-manager": "^0.39.1",
+    "@jupyter-widgets/jupyterlab-manager": "^0.40.0",
     "@jupyter-widgets/output": "^1.1.2",
     "@jupyterlab/coreutils": "^2.2.1",
     "@jupyterlab/outputarea": "^0.19.1",
--- a/js/src/manager.js
+++ b/js/src/manager.js
@@ -27,8 +27,9 @@ export class WidgetManager extends JupyterLabManager {
 
     constructor(kernel) {
         const context = createContext(kernel);
+        const settings = createSettings();
         const rendermime = createRenderMimeRegistry();
-        super(context, rendermime);
+        super(context, rendermime, settings);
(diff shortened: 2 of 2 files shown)


js/package.json
```diff
@@ -8,7 +8,7 @@
     "@jupyter-widgets/base": "^1.2.2",
     "@jupyter-widgets/controls": "^1.4.2",
     "@jupyter-widgets/html-manager": "^0.15.2",
-    "@jupyter-widgets/jupyterlab-manager": "^0.39.1",
+    "@jupyter-widgets/jupyterlab-manager": "^0.40.0",
```
SylvainCorlay: you probably need to update html-manager to 0.16 then?

maartenbreddels: Good question. I thought html-manager was not a dependency anymore, but it is, only for the requireLoader
https://github.com/QuantStack/voila/blob/ab241427955d6dd917e74962d6365937f9c3fffe/js/src/manager.js#L10
So it's fine I think. I'd rather see that requireLoader go into main.js, and it being an argument of the widget manager, so also this part is pluggable.

maartenbreddels:
Travis failure: The bundle is now so large, it refuses to build with source mapping. Locally I had to disable source mapping to build it.

SylvainCorlay:
> Travis failure: The bundle is now so large, it refuses to build with source mapping. Locally I had to disable source mapping to build it.

:unamused:

js/package.json
SylvainCorlay: Yeah, sorry for a minute I thought that html-manager depended on jupyterlab-manager.

SylvainCorlay:
bumping jupyterlab-manager increases the size of the bundle so much?

maartenbreddels:
I'm exploring what is happening now using https://www.npmjs.com/package/webpack-bundle-analyzer

maartenbreddels:
It seems that bumping the manager adds +10MB in from jupyterlab, why that happens, I am not sure yet. The bundle is now ~25mb instead of ~5mb.

SylvainCorlay:
Fixed in #277

PR #154, opened 2019-05-24
WIP: Chore: use css variables

maartenbreddels:
This is a follow up of #153, once that is merged and https://github.com/jupyter-widgets/ipywidgets/pull/2418 + released, we can bump the @jupyter-widgets/jupyterlab-manager version number and have voila with theming support:
image


maartenbreddels:
Note: if bqplot gets loaded now, the theme changes white, I'm assuming that is expected, since bqplot ships the css variables?

SylvainCorlay:
Yes, I did exactly the same work as this PR yesterday, which triggered

https://github.com/bloomberg/bqplot/pull/855

and the work on removing postcss from ipywidgets.

SylvainCorlay:
Closing since merged #156 with only the removal of postcss.

maartenbreddels:
Ah, I got confused, I thought we need master of ipywidgets, so meaning we also had to bump the version number, but that was not needed. Explains how you got it to work without! :)

SylvainCorlay:
We need master ipywidgets for the bqplot PR, for it to continue working in the classic notebook.

PR #155, opened 2019-05-25
Use jupyterlab_pygments to generate syntax coloring CSS

SylvainCorlay:
Since `pygments` 2.4.1 is out, I could release `jupyterlab_pygments` and make use of it.


maartenbreddels:
Good to see so many lines being removed!

PR #159, opened 2019-05-26
Use plural for template(s) data dir

SylvainCorlay:
I think that the `tiebreaker` for this discussion is the case of Jupyter kernels, which is plural.


SylvainCorlay:
cc @aschlaep for custom template authors, the consequence is that new template should be installed into `PREFIX/share/jupyter/voila/templates` and not `PREFIX/share/jupyter/voila/template`.

maartenbreddels:
I agree it should be plural now. But I think we may want to have more changes, and I'm not sure we should break the template system twice in a short time.

SylvainCorlay:
We are at 0.0.10, I think we can break it now, but not after we release 0.1.0.

PR #161, opened 2019-05-27
Add CONTRIBUTING.md

### LINKED ISSUE #134: Improve development installation
I still haven't been able to get Voila installed locally for development. It's hard for me to tell if this is because I'm following instructions wrong because there aren't any develop installation instructions that I could find.

For example, if I do `pip install -e .` then I get the following output before the installation hangs:

```
pip install -e .
Obtaining file:///home/choldgraf/github_local/voila
^CERROR: Operation cancelled by user
^[[A^[[A(dev) choldgraf@barbecue ~/github_local/voila (master) $ pip install -v -e .
Created temporary directory: /tmp/pip-ephem-wheel-cache-sjwrv7j1
Created temporary directory: /tmp/pip-req-tracker-242owfut
Created requirements tracker '/tmp/pip-req-tracker-242owfut'
Created temporary directory: /tmp/pip-install-3akv39hm
Obtaining file:///home/choldgraf/github_local/voila
  Added file:///home/choldgraf/github_local/voila to build tracker '/tmp/pip-req-tracker-242owfut'
    Running setup.py (path:/home/choldgraf/github_local/voila/setup.py) egg_info for package from file:///home/choldgraf/github_local/voila
    Running command python setup.py egg_info
    running egg_info
    running jsdeps
    6.7.0
    6.7.0
    Installing build dependencies with npm.  This may take a while...

    > @QuantStack/voila@1.0.0 prepare /home/choldgraf/github_local/voila/js
    > npm run build:lib && webpack


    > @QuantStack/voila@1.0.0 build:lib /home/choldgraf/github_local/voila/js
    > babel src --out-dir lib

    Successfully compiled 3 files with Babel.
```

We should add developer docs so that people can get started!

jtpio:
Fixes #134.

This is the set of commands I've used so far and they have been enough for regular development.

@choldgraf @SylvainCorlay @maartenbreddels feel free to jump in and add anything missing.
I'll do a second pass on it a bit later today / this week.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,25 +38,47 @@ Then reload the browser tab.
 
 ## Extensions
 
+### Server extension
+
 To manually enable the classic notebook server extension:
 
 ```bash
-jupyter serverextension enable voila --sys-prefix [--py | --user]
+jupyter serverextension enable voila --sys-prefix
 ```
 
 For Jupyter Server:
 
 ```bash
-jupyter extension enable voila --sys-prefix [--py | --user]
+jupyter extension enable voila --sys-prefix
 ```
 
 This makes voila available as a server extension: [http://localhost:8888/voila/tree](http://localhost:8888/voila/tree).
 
+### Notebook extension
+
 To install the notebook extension:
 
 ```bash
-jupyter nbextension install voila --sys-prefix [--py | --user]
-jupyter nbextension enable voila --sys-prefix [--py | --user]
+jupyter nbextension install voila --sys-prefix
+jupyter nbextension enable voila --sys-prefix
+```
+
+### JupyterLab extension
+
+Node.js is required and can be installed with conda:
+
+```bash
+conda install -c conda-forge nodejs
+```
+
+To install the JupyterLab extension locally:
+
+```bash
+jupyter labextension install @jupyter-widgets/jupyterlab-manager
+jupyter labextension install ./packages/jupyterlab-voila
+
+# start in watch mode to pick up changes automatically
+jupyter lab --watch
 ```
 
 ## Running the examples
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,13 @@ Then reload the browser tab.
 To manually enable the classic notebook server extension:
 
 ```bash
-jupyter serverextension enable voila.server_extension --sys-prefix [--py | --user]
+jupyter serverextension enable voila --sys-prefix [--py | --user]
+```
+
+For Jupyter Server:
+
+```bash
+jupyter extension enable voila --sys-prefix [--py | --user]
 ```
 
 This makes voila available as a server extension: [http://localhost:8888/voila/tree](http://localhost:8888/voila/tree).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,7 @@ python -m pip install -e .
 When making changes to the frontend side of voila, open a new terminal window and run:
 
 ```bash
+cd js/
 npm run watch
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,23 @@ npm run watch
 
 Then reload the browser tab.
 
+## Extensions
+
+To manually enable the classic notebook server extension:
+
+```bash
+jupyter serverextension enable voila.server_extension --sys-prefix [--py | --user]
+```
+
+This makes voila available as a server extension: [http://localhost:8888/voila/tree](http://localhost:8888/voila/tree).
+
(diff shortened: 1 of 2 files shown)


choldgraf:
I'll take another pass at getting this set up for myself this week. I still haven't been able to get Voila working in dev mode :-/

jtpio:
Second pass done. This should be ready to go.

@choldgraf did you give it another try?

choldgraf:
will give this a shot today

choldgraf:
Some notes:

- Install instructions worked this time w/ `pip install -e .` dunno if that was a change on my end or your end but 👍 
- I got an error about not having the right version of pygments installed when I tried to run `voila`. running `pip install -U Pygments` got it to work. Perhaps that should be added in here?

jtpio:
>     * I got an error about not having the right version of pygments installed when I tried to run `voila`. running `pip install -U Pygments` got it to work. Perhaps that should be added in here?

It should normally be installed automatically (via `jupyterlab_pygments`)

choldgraf:
Yeah - I think I just hadn't updated it recently which is why I had an outdated version. Maybe just add a "conda update XXX" for the required conda packages? Though I agree most developer types could have figured it out

choldgraf:
Ah one more point here after experimenting a bit more. It doesn't seem like this installs the `voila` server extension (at least, I'm getting a 500 error when I run jupyter lab and hit `/voila`. Is that right?

SylvainCorlay:
> I got an error about not having the right version of pygments installed when I tried to run voila. running pip install -U Pygments got it to work. Perhaps that should be added in here?

Both the conda recipe and setup.py for jupyterlab_pygments require `pygments>=2.4.1` so it is weird that you could get it with an older version. I don't think we should specify anything though.

SylvainCorlay:
> Ah one more point here after experimenting a bit more. It doesn't seem like this installs the voila server extension (at least, I'm getting a 500 error when I run jupyter lab and hit /voila. Is that right?

Indeed, development installations don't install data files. Jupyter server extensions rely on a configuration directory system in `share`, where the assets for the voila jupyterlab extension are placed.

In the case of a development installation, the extension can be installed and enabled with:

 - for the classic notebook server

```
jupyter serverextension install --py --sys-prefix --symlink voila
jupyter serverextension enable --py --sys-prefix voila
```

the `--symlink` is not supported on windows.

 - for `jupyter_server`

```
jupyter extension install --py --sys-prefix --symlink voila
jupyter extension enable --py --sys-prefix voila
```

maartenbreddels:
I think we can fix all that by adding a custom devel command in setup.py, and symlinks on windows 10 seem to work fine, but requires admin mode (I think). Maybe a 'good first issue' to open?

SylvainCorlay:
> I think we can fix all that by adding a custom devel command in setup.py, and symlinks on windows 10 seem to work fine, but requires admin mode (I think). Maybe a 'good first issue' to open?

Yes, that would make sense.

In the mean time, we could add these instructions for those who want to use the server extensions in development mode.

jtpio:
> ```
> jupyter serverextension install --py --sys-prefix --symlink voila
> ```

Isn't the `jupyter serverextension enable` command enough? (no `install` subcommand for server extensions)

We can also add instructions for the notebook extension (which adds the "Voila" button).
And when https://github.com/QuantStack/voila/pull/40 is merged instructions for the JupyterLab extension.

yuvipanda:
If you're using this with jupyter_server (instead of notebook), you'll need to run `jupyter extension enable` instead of `jupyter serverextension enable`

jtpio:
Updated based on https://github.com/QuantStack/voila/pull/182 and @yuvipanda's comment.

CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+To manually enable the classic notebook server extension:
+
+```bash
+jupyter serverextension enable voila --sys-prefix [--py | --user]
+```
+
+For Jupyter Server:
+
+```bash
+jupyter extension enable voila --sys-prefix [--py | --user]
```
SylvainCorlay: remove `--user`.

SylvainCorlay: It seems to me that `--py` is always required?

jtpio: `--py` throws the following for me:

```
KeyError: 'The Python module voila does not include any valid server extensions'
```

jtpio: Full command: `jupyter serverextension enable voila --sys-prefix --py`

SylvainCorlay: indeed we don't have an entry point for a jupyter server extension in `__init__.py`. We should probably just remove the `--py`.

choldgraf:
Woot, thanks for this @jtpio

PR #163, opened 2019-05-28
Remove gridstack template

SylvainCorlay:
The gridstack template is now installed as a separate package.

```
pip install voila-gridstack
```


timkpaine:
I have my own template, is there some standardization you want to impose for the pypi projects? Alternatively I can put it on GH then transfer it to you

maartenbreddels:
We now have:
 * voila-gridstack
 * voila-vuetify 
So `voila-<template-name>` for the package name (with dash), and let setuptools do any dash to underscore conversion. Sound good?

timkpaine:
Yep that works

SylvainCorlay:
Thanks.

FYI @timkpaine we just merged a PR changing the custom templates directory from

`PREFIX/share/jupyter/voila/template`

to

`PREFIX/share/jupyter/voila/templates`

which will be part of 0.0.11.

Other then that, your current template should still work in the next version.

timkpaine:
https://github.com/timkpaine/voila-plain

timkpaine:
I just copied the code from the grid stack one and replaced the templates with my own, thanks!

timkpaine:
I added you guys as PyPi maintainers in case the name is too generic and need to take it over in the future.

maartenbreddels:
Great, thanks!
Note/warning: we will most likely break the current template system in the future.

SylvainCorlay:
(To enable progressive rendering and imorove the nbconvert template system.)

PR #168, opened 2019-05-31
adding information about customizing voila and static files

### LINKED ISSUE #140: Document how users (or template developers) can add static files
It's a bit unclear how custom CSS/JS can be controlled by either users or those who create their own templates. From the command line, it seems like `--static` provides some way to give a path to a directory w/ these assets, but I'm not sure how to name those files or make sure they actually get loaded by a template.

I think it'd be helpful in the documentation to have:

1. A guide explaining how to include custom CSS / JS with a template (e.g. if I want a template that links a CSS file in the template file itself how would I do that)
2. A guide explaining if it's possible, and if so how to use the `--static` parameter to control the static assets when `voila` is run

choldgraf:
This adds a few pieces of information:

* How to define a new location of static files and using them in a template
* How to configure Voila for use with the Jupyter Server

Would be good if somebody can confirm this information is correct :-)

closes #140

### DIFF
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -135,4 +135,51 @@ a Jupyter notebook by using the name of the folder in the ``--template`` paramet
 
     voila mynotebook.ipynb --template=mytemplate
 
+
 The result should be a Voila dashboard with your custom modifications made!
+
+
+Adding your own static files
+============================
+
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
This looks great! Merging.

PR #170, opened 2019-06-01
Add example using matplotlib in an output widget

pbugnion:
One of the great things about voila and widgets in general is that they allow a gradual transition from a development workflow to building an app. In my experience, during development, users will use matplotlib. They will only switch to a more interactive library like bqplot or plotly later.

I think it's useful to show an example of using matplotlib with voila. I've basically reproduced an example from the Plotly documentation, switching to matplotlib.

The app looks like:

Screenshot 2019-06-01 at 11 02 58

If you don't think this should be merged into voila itself, happy to spin this off as a separate example.

### DIFF
--- a/notebooks/matplotlib.ipynb
+++ b/notebooks/matplotlib.ipynb
@@ -1,0 +1,256 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 72,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import matplotlib.pyplot as plt\n",
+    "import ipywidgets as widgets\n",
+    "from IPython.display import HTML\n",
+    "\n",
+    "plt.ioff()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 80,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<style>\n",
+       ".app-sidebar p {\n",
+       "    margin-bottom: 1em;\n",
+       "    line-height: 1.7;\n",
+       "}\n",
+       "\n",
+       ".app-sidebar a {\n",
+       "    color: #106ba3;\n",
+       "}\n",
+       "\n",
+       ".app-sidebar a:hover {\n",
+       "    text-decoration: underline;\n",
+       "}\n",
+       "</style>\n",
+       "    "
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "execution_count": 80,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Quick note: the ipympl widget works perfectly in voila.

This would be a great example to show thatvwe can have interactive matplotlib in voila.

Will push out another example.

pbugnion:
Ah -- re-implementing this with ipympl sounds great. Happy to give this a go tomorrow or next week.

SylvainCorlay:
Using a couple of examples from the ipympl gallery:

```python
%matplotlib widget
import ipympl

import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 2 * np.pi, 500)
y1 = np.sin(x)
y2 = np.sin(3 * x)

fig, ax = plt.subplots()
ax.fill(x, y1, 'b', x, y2, 'r', alpha=0.3)
plt.show()
```

and

```python
import numpy as np
import matplotlib.pyplot as plt

plt.style.use('ggplot')

fig, axes = plt.subplots(ncols=2, nrows=2)
ax1, ax2, ax3, ax4 = axes.ravel()

# scatter plot (Note: `plt.scatter` doesn't use default colors)
x, y = np.random.normal(size=(2, 200))
ax1.plot(x, y, 'o')

# sinusoidal lines with colors from default color cycle
L = 2*np.pi
x = np.linspace(0, L)
ncolors = len(plt.rcParams['axes.prop_cycle'])
shift = np.linspace(0, L, ncolors, endpoint=False)
for s in shift:
    ax2.plot(x, np.sin(x + s), '-')
ax2.margins(0)

# bar graphs
x = np.arange(5)
y1, y2 = np.random.randint(1, 25, size=(2, 5))
width = 0.25
ax3.bar(x, y1, width)
ax3.bar(x + width, y2, width,
        color=list(plt.rcParams['axes.prop_cycle'])[2]['color'])
ax3.set_xticks(x + width)
ax3.set_xticklabels(['a', 'b', 'c', 'd', 'e'])

# circles with colors from default color cycle
for i, color in enumerate(plt.rcParams['axes.prop_cycle']):
    xy = np.random.normal(size=2)
    ax4.add_patch(plt.Circle(xy, radius=0.3, color=color['color']))
ax4.axis('equal')
ax4.margins(0)

plt.show()
```

ipympl

SylvainCorlay:
(Note that this screencast was done with a terrible internet connection on a train).

jeffyjefflabs:
If you're curious, I've been playing with how the different matplotlib APIs and back-ends work inside of button on_click callbacks when rendered in voila.  [Examples here](https://github.com/nbgallery/dashboards/tree/master/voila).

jtpio:
Thanks @pbugnion. 

Do you think this should be added to the examples notebooks in this repo, or kept in the gallery? 

https://github.com/voila-gallery/voila-gallery-country-indicators

maartenbreddels:
Closing this as stale.

PR #182, opened 2019-06-04
add top-level serverextension and installation instructions

ivanov:
Here's what the output looks like with these modest changes
```
$ jupyter serverextension enable voila
Enabling: voila
- Writing config: /home/pi/.jupyter
    - Validating...
      voila 0.0.13 OK
```

and 

```
$ jupyter serverextension list
config dir: /home/pi/.jupyter
    voila  enabled
    - Validating...
      voila 0.0.13 OK
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,16 @@ Voila can also be used as a notebook server extension, both with the
 [notebook](https://github.com/jupyter/notebook) server or with
 [jupyter_server](https://github.com/jupyter/jupyter_server).
 
+To install the notebook server extension, run
+
+```
+jupyter serverextension enable voila.server_extension
+```
+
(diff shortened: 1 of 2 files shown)


ivanov:
this is a smaller scope version than #161, but makes the serverextension work with just a top-level `enable voila`, instead of having to use `enable voila.serverextension`

SylvainCorlay:
> `$ jupyter serverextension enable voila`

I think that we should probably recommend the `--sys-prefix` option.

Otherwise, people may be surprised that this is not conda-environment specific or that enabling the extension requires more privileges than installing the package.

ivanov:
👍 alright, wrote in `--sys-prefix` as the only preferred way of installing.

PR #183, opened 2019-06-04
Expose phosphor

### LINKED ISSUE #116: Missing phosphor widgets js
We separated the jupyterlab ipywidget from our phosphor bindings from our [perspective](https://github.com/jpmorganchase/perspective/pull/436) library, which has the effect of introducing an explicit dependency on phosphor which seems to be failing. Any insights on how we might rectify this?

image

SylvainCorlay:
Fixes #116

Cherry-picking @timkpaine's commit.


timkpaine:
Thanks!

PR #186, opened 2019-06-05
don't 500 on cell execution errors

ivanov:
Sometimes, code execution will error (for example, missing dependencies), and it's rough to get 500 errors in the browser and have to look in the console where the server was started to see what the error was. This code leaves in-place the error presented in the notebook when executing that particular cell, and continues to log in on the server-side, but without causing a server 500 error.

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I recall that the nbexecute preprocessor had some options to allow continuation of execution if a cell fails.

I wonder how this could work with this.

vidartf:
If the cell has a tag `raises-exception` (or something like that) it will allow any exceptions to pass, I think

PR #191, opened 2019-06-05
Only use VoilaHandler for .ipynb files

### LINKED ISSUE #190: Only handle .ipynb files
For now the `VoilaHandler` can only process notebook files (`.ipynb`).

However the handler is currently using `path_regex` from `jupyter_server.base.handlers`, which catches all files:

https://github.com/QuantStack/voila/blob/cadc6e801a0a4537ee216fdfe0b67682e054d9ba/voila/app.py#L37

As a result, some non .ipynb files can hit the `VoilaHandler`:

```
WARNING:tornado.general:404 GET /voila/render/b06871f281fee6b241d60582ae9369b9.ttf (127.0.0.1): No such file or directory: b06871f281fee6b241d60582ae9369b9.ttf
```

It would make sense to use a custom `path_regex` to only catch .ipynb files (at least for now).

martinRenou:
This should fix #190

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -34,15 +34,15 @@ from traitlets import Unicode, Integer, Bool, Dict, List, default
 
 from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
 from jupyter_server.services.kernels.handlers import KernelHandler, ZMQChannelsHandler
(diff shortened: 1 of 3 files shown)


martinRenou:
Is it fine for merging?

PR #199, opened 2019-06-06
Expose dir content to the VoilaHandler template

### LINKED ISSUE #197: VoilaHandler: provide directory content
It would be nice to have access to the path content from the `VoilaHandler`. In some templates we could display the tree in a sidebar next to the rendered Notebook.

martinRenou:
This allows one to make a voila template with a tree view of the current directory:

This should fix #197

voila_tree

As discussed with @maartenbreddels, there might be better ways of doing this.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -59,11 +59,17 @@ class VoilaHandler(JupyterHandler):
             return
         self.cwd = os.path.dirname(notebook_path)
 
+        # Get path content
+        dir_path, notebook_name = os.path.split(notebook_path)
+        contents = self.contents_manager.get(dir_path)
+
         # render notebook to html
         resources = {
             'base_url': self.base_url,
             'nbextensions': nbextensions,
-            'theme': self.voila_configuration.theme
+            'theme': self.voila_configuration.theme,
+            'contents': contents,
+            'notebook_name': notebook_name
         }
 
         # include potential extra resources


SylvainCorlay:
I need to think on whether it is an issue to expose the content of the current directory.

SylvainCorlay:
Grumbles.

maartenbreddels:
Yes indeed, it does pose a security issue. If a single notebook is shown, the content should not be exposed. We were setting up testing for this.

martinRenou:
Yes. I'm removing the gridstack dependency in the tests and activating the appveyor tests before.

jeffyjefflabs:
Love the sidebar idea.  Could you achieve the same effect with a tree template instead?  The notebook could be in an iframe or loaded with jquery or something.

martinRenou:
@jeffyjefflabs If we go for the "Only one handler" idea it will be possible to do it properly with a template :)

martinRenou:
> I need to think on whether it is an issue to expose the content of the current directory.

> Yes indeed, it does pose a security issue. If a single notebook is shown, the content should not be exposed. We were setting up testing for this.

Thinking more about it. I think it is not a security issue. It only exposes the directory content to the Jinja template. The default template does not use it.

Only if you use a custom template that makes use of this variable you will see the directory content. But that is completely up to the dashboard developer.

maartenbreddels:
I'd still like to see this land in the handler, but it would preferably use the same code as in treehandler.py (the allowed_content function).

PR #200, opened 2019-06-06
Adds styling and a header to the voila tree

DanielAristidou:
Adds css styling and a simple header to the voila tree. 

Css is wrapped in a <style> within the tree.html


SylvainCorlay:
Thanks @DanielaArstidou!

Issue #188, opened 2019-06-05
Template name in Notebook metadata


martinRenou:
It would be nice to be able to specify which template to use with a given Notebook, inside of the Notebook metadata.

For example, this would allow the voila-gallery to use different templates. cc. @pbugnion

vidartf:
cc @jasongrout

timkpaine:
This is related to the issue about specifying the template in the url

martinRenou:
#105

PR #201, opened 2019-06-06
Switch templates via query string parameters

jtpio:
A proof of concept for switching templates on the fly via the notebook metadata and query string parameters.

Related to https://github.com/QuantStack/voila/issues/188 and https://github.com/QuantStack/voila/issues/105.

- ~Rename the cli parameter `template` to `templates` to pass a list of available templates (instead of a single one)~
- ~Fetch `template` from the notebook metadata~
- Fetch `template` from the query string parameters
- (tests are not fixed yet)

One thing I'm not super happy with is the need to sort the `nbconvert_template_paths` list before passing it to the nbconvert exporter.

Also this approach might be completely invalidated if / when the template mechanism is changed for the next release.

template-url

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -68,12 +68,13 @@ class VoilaHandler(JupyterHandler):
         # choose the template
         template_name = (
             self.get_query_argument('template', '') or
-            self.voila_configuration.template or
-            'default'
+            self.voila_configuration.template
         )
 
-        # prioritize paths based on the template name
-        templates = sorted(self.nbconvert_template_paths, key=lambda p: -(template_name in p))
+        templates = self.nbconvert_template_paths
+        if self.voila_configuration.template:
+            # prioritize paths based on the template name if specified
+            templates = sorted(self.nbconvert_template_paths, key=lambda p: -(template_name in p))
 
         exporter = VoilaExporter(
             template_path=templates,
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -65,8 +65,18 @@ class VoilaHandler(JupyterHandler):
             'theme': self.voila_configuration.theme
         }
 
+        # choose the template
+        template_name = (
+            self.get_query_argument('template', '') or
+            self.voila_configuration.template or
+            'default'
+        )
+
+        # prioritize paths based on the template name
+        templates = sorted(self.nbconvert_template_paths, key=lambda p: -(template_name in p))
+
         exporter = VoilaExporter(
-            template_path=self.nbconvert_template_paths,
+            template_path=templates,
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
Thanks @jtpio!

I wonder if we could drop the list of allowed templates and just let users select *any* installed template from the url params.

The `--template=foobar` command-line argument would just select the default template that is used when none is specified with the url parameter.

martinRenou:
> I wonder if we could drop the list of allowed templates and just let users select any installed template from the url params.

:+1:

jtpio:
Yes absolutely.

jtpio:
Pushed new changes to:

- Drop the `--templates=["gridstack", "default"]` option and instead use all available templates
- Keep `--template` option to default to a specific template
- Assume there is always the `default` template installed

notebooks/basics.ipynb
```diff
@@ -82,6 +82,9 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.7.3"
+  },
+  "voila": {
+   "template": "default"
```
jtpio: We might want to think about a schema or a better structure before using the notebook metadata for voila.

jtpio: out of scope for now

voila/handler.py
```diff
@@ -65,8 +65,19 @@ def get(self, path=None):
...
+        # choose the template
+        template_name = (
+            self.get_query_argument('template', '') or
+            self.voila_configuration.template
+        )
+
+        templates = self.nbconvert_template_paths
+        if self.voila_configuration.template:
+            # prioritize paths based on the template name if specified
+            templates = sorted(self.nbconvert_template_paths, key=lambda p: -(template_name in p))
```
martinRenou: This `sorted` call feels bad to me... I suppose we should collect the right template paths for each request. See #258

martinRenou: I suppose something cleaner and safer would be `templates = collect_template_paths(..., template_name=...)`

jtpio: Agree it's not very elegant.
Would need to check how nbconvert handles templates that make use of inheritance.

cailiang9:
It is cool but not a proper place to specify notebook specific template parameters.

maartenbreddels:
> Allowing specify template from common users is unsafe.

I think both options should be possible, but I agree that a user specifying a template should be possible to disable.

maartenbreddels:
Closes #188

jtpio:
Fixed by https://github.com/voila-dashboards/voila/pull/637

PR #202, opened 2019-06-06
Ignore static dir

martinRenou:


SylvainCorlay:
I don't think we can ignore the static directory because the `extension.js` file is actually not generated.

martinRenou:
I had some font files in the static dir, is it supposed to be committed too?

SylvainCorlay:
Hum no. are they re-generated if you delete them and rebuild voila?

SylvainCorlay:
Closing in the mean time so that it does not get in in the mean time.

martinRenou:
> Hum no. are they re-generated if you delete them and rebuild voila?

They are not, apparently. I'm not sure where it was coming from then.

PR #206, opened 2019-06-06
[WIP] Adds the ability to remove cells based on a tag

### LINKED ISSUE #171: Make it possible to hide markdown cells
I've found that I use markdown cells for two purposes:

1. To display user-friendly content alongside the code+outputs
2. To display *developer*-friendly reminders about the code's logic etc

In some documents, you want *both* of these things when authoring the notebook, but only the *user-friendly* content when running in a dashboard mode. Currently, I believe that Voila will display all Markdown cells, but it'd be great if there were some control over this.

One way to do this: Perhaps Voila could make use of a cell tag in order to control whether the markdown. Something like `voila_hide` could be used to turn markdown visibility on and off.

choldgraf:
This will exclude a cell from the HTML output if the tag `remove_cell` is attached to the cell metadata.

**To do**

- [ ] Decide if `remove_cell` is the right tag to use
- [ ] Add tests

closes #171

### DIFF
--- a/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
@@ -54,3 +54,15 @@ requirejs(
 {{ super() }}
 </html>
 {% endblock footer %}
+
+{%- block codecell %}
+{% if 'remove_cell' not in cell.metadata.tags %}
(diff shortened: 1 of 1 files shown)


choldgraf:
closing this because it turns out this is natively possible in nbconvert, and we should use that if we can

PR #207, opened 2019-06-07
Remove gridstack from test dependencies

martinRenou:


maartenbreddels:
My only concern here is that it is not obvious to contributors that they need to install this. If you make it also a module (test_template.py), you can try to import this in the test. If it fails to import, you can use `pytest.mark.skip` to skip the test. I think it is ok for local running that we skip this since the CI will test it.

martinRenou:
Either that or we document somewhere that you need to install the test_template before trying to run the tests

maartenbreddels:
maybe you can add it to the assert msg?
```
assert <expression>, "<reason>, or maybe you forgot to do (cd tests\test_template; pip install .)"
```
I'd rather not have people have to read the documentation before being able to contribute.

martinRenou:
You mean doing the assert in a fixture? I'm not sure we want to do an assert in each test that uses the test_template

maartenbreddels:
No, when a contributor wants to run the tests I think it should give a meaningful error, without having to dive into the docs. Basically, pip uninstall the template, see what error you get and make see if you can give the user/contributor a hint on how to fix it.

martinRenou:
I suppose the current way it fails is just fine, it says:
```
E           ValueError: No template sub-directory with name 'test_template' found in the following paths:
E           	/home/martin/.local/share/jupyter/voila/templates
E           	/home/martin/miniconda3/envs/voila/share/jupyter/voila/templates
E           	/usr/local/share/jupyter/voila/templates
E           	/usr/share/jupyter/voila/templates
```

SylvainCorlay:
> I suppose the current way it fails is just fine.

+1 on this. Will go ahead an merge.

Note: eventually we should document how to run tests locally - or make the installation of `test_template` automatic.

PR #217, opened 2019-06-11
JupyterLab extension

### LINKED ISSUE #39: JupyterLab extension
Same idea as in https://github.com/QuantStack/voila/issues/36 and https://github.com/QuantStack/voila/pull/37, but for JupyterLab. And as @maartenbreddels mentioned:

- add a command to the palette / menu
- add a button to the toolbar

jtpio:
Fixes #39.

Replaces #40 (keeps Maarten's commit).

Add a JupyterLab extension to render a notebook with voila.

### TODO

- [x] Bootstrap the extension
- [x] Add a command to the command palette, view menu and notebook toolbar
- [x] Expose two commands: view in a JupyterLab panel (IFrame) and open in a new browser tab, using the server extension endpoint
- [x] Show the notebook name in the preview tab
- [x] Use `optional` instead of `requires` for dependencies
- [x] Use voila icon for the toolbar button and the preview title
- [x] Prettier git commit hook
- [ ] Render voila preview on save
- [ ] Handle current JupyterLab theme in the voila preview (dark / light)
- [ ] Add dev setup to CONTRIBUTING.md

jupyterlab-extension

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ share/jupyter/voila/templates/default/static/*voila.js
 share/jupyter/voila/templates/default/static/*[woff|woff2|eot|svg]
 
 js/lib/
+
+packages/jupyterlab-voila/lib/
--- a/packages/jupyterlab-voila/package.json
+++ b/packages/jupyterlab-voila/package.json
@@ -25,33 +25,49 @@
   },
   "scripts": {
     "build": "tsc",
+    "prettier": "prettier --write '**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}'",
     "clean": "rimraf lib",
     "watch": "tsc -w",
     "prepare": "npm run clean && npm run build"
   },
+  "dependencies": {
+    "@jupyterlab/application": "^0.19.0",
+    "@jupyterlab/apputils": "^0.19.0",
+    "@jupyterlab/docregistry": "^0.19.0",
+    "@jupyterlab/fileeditor": "^0.19.0",
+    "@jupyterlab/mainmenu": "^0.8.1",
+    "@jupyterlab/notebook": "^0.19.0"
+  },
   "peerDependencies": {
     "@jupyterlab/application": "^0.19.0",
     "@jupyterlab/apputils": "^0.19.0",
-    "@jupyterlab/notebook": "^0.19.0",
-    "@jupyterlab/mainmenu": "^0.19.0",
     "@jupyterlab/docregistry": "^0.19.0",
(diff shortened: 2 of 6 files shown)


SylvainCorlay:
Thanks!

Do you think it would make sense to have the package currently in `js`  also in packages?

jtpio:
Grouping the two together sounds like a good idea, even though one is for the voila frontend (`js`) and the other one is for the JLab extension (`packages/jupyterlab-voila`).

Another option would be to bring the `jupyterlab-voila` folder one level up next to `js`.

SylvainCorlay:
Yep, for the the ja package naming (under `packages` or elsewhere), I don't know if there is a pattern we should use.

jtpio:
And the name `packages` is mostly for `lerna` projects, so we can rename it to something else.

wolfv:
just a nitpick but i think a monochrome icon would look better in that icon bar.

jtpio:
It would fit with the other existing icons yes.

Another option is to use the full text `[voilà]` icon in monochrome.

jtpio:
Re-rendering the preview on save:

jlab-render-save

I think the main problem with this feature will be the number of kernels being started every time the notebook is saved (could be an opt-in option instead)

jtpio:
We merged this one early so we can iterate on the remaining points in follow-up PRs.

jtpio:
Release coming soon!

wolfv:
adding instruction on how to dev-install it in the readme would be nice (for people like me) ;)

jtpio:
There are some instructions in the contributing guide: https://github.com/QuantStack/voila/blob/master/CONTRIBUTING.md#jupyterlab-extension

wolfv:
Nice! didn't see it, sorry.

jtpio:
No problem!

Magnus512:
Can we save the formated output in PDF file or something like that, if we are not working with interactive widgets ?

jtpio:
@Magnus512 one way to do it for now could be to use the "Open with Voila in New Browser Tab" command, and then use the web browser to print to PDF?

image

Magnus512:
> @Magnus512 one way to do it for now could be to use the "Open with Voila in New Browser Tab" command, and then use the web browser to print to PDF?
> 
> image

I know this, I was thinking in something more ammm "Beautiful", in order to get a final document for presentation, anyway thanks for your reply :)

jtpio:
As a side note, JupyterLab 1.0 added a "Print" entry in the File Menu which can be used to print notebooks in a nice way.
Maybe we could add a similar functionality to the voila preview toolbar.

PR #218, opened 2019-06-11
Adding prelaunch hook for kernel instances

timkpaine:
I will have an example of hooking this into auth in my [dashboard](https://github.com/timkpaine/dashboard) repo, but for now here is a simple example of pre-processing a notebook prior to launch, which adds an extra title to the page.

```python3
import voila.app

v = voila.app.Voila()
v.notebook_path = './notebooks/bqplot.ipynb'


def prelaunch_add_header(req, notebook, **kwargs):
    from nbformat import NotebookNode
    new_node = NotebookNode()
    new_node.cell_type = 'markdown'
    new_node.metadata = {}
    new_node.source = '# An added title!'
    notebook.cells.insert(0, new_node)

v.prelaunch_hook = prelaunch_add_header

v.initialize()
v.start()
```

Which yields:

Screen Shot 2019-06-11 at 9 04 04 AM


maartenbreddels:
Hi Tim, I like the idea, I wonder if this isn't better solved with a template?

timkpaine:
@maartenbreddels the tough part is making https://github.com/QuantStack/voila/issues/216 and https://github.com/QuantStack/voila/issues/239 and https://github.com/QuantStack/voila/issues/105 play nicely together. I agree that my simple example here is better handled by a template (and indeed, I have a template that does exactly this). But if you can specify a template with a url parameter (which I think is a good idea), you'll still want some hook in (for example, I will want to put things like `FOR INTERNAL USE ONLY` regardless of the template my users use). 

IMO we should try to build in a few of the same hooks that exist in JLab
- pre execute (equivalent of presave hook in jupyter contents managers)
- The authentication for who can launch a kernel should match who can launch a lab instance

probably more to follow

leogout:
Hey, I would love to hear from this. I've been wondering how to authenticate a user before he gets to a voila rendered notebook. I would like to experiment and give feedback on what's possible to do with these hooks, especially with OAuth2.
Do you think it would be possible to redirect a user to a third party login page back and forth with this solution ? I'm having a hard time getting used to the voila codebase so I can't tell if it's feasible or not...

maartenbreddels:
> IMO we should try to build in a few of the same hooks that exist in JLab

Could you give a few pointers to where that is in the jlab code?

timkpaine:
Prelaunch hook is actually from IPython, you can specify lines to execute at startup
https://ipython.readthedocs.io/en/stable/config/options/kernel.html

All of these are useable via jupyter_server, so that's definitely the right approach

timkpaine:
I just wanted to comment here, I've been using this HEAVILY in my fork. Its super useful as a per-user hook into the notebook, letting you inject arbitrary code cells which is super nice for doing validation and authentication.

maartenbreddels:
Thanks for letting us know, we'll have to see how this mixes with the progressive loading and follow up plans, but I understand its usefulness.

timkpaine:
Another good use is logging, I can grab the notebook name and inject a cell that sends remote log info about who is executing what notebook

mgmarino:
I'd just like to +1 this.  Given the current implementation, this looks like it's a nice way to handle granular authorization.

That is, on a production system, we would like to put the voila app behind e.g. an OAuth authentication service (like e.g. [here](https://aws.amazon.com/de/blogs/aws/built-in-authentication-in-alb/)), which would pass on information (like user, scopes, etc.) in the request headers to voila.  This information could be used to decide whether or not to load the notebook (ideally using information *from* the notebook, perhaps an access list stored in metadata?) and raise a `tornado.web.HTTPError(404, 'file not found')` if access is not allowed:.


This approach is somewhat different than that suggested by @leogout since the authentication would be handled elsewhere.

timkpaine:
@mgmarino we do something very similar, thats why the prelaunch hook has access to the handler. For us, it remote logs the user accessing the server, attempts to authenticate, raises a 401 with a redirect to authenticate if there are issues, then injects a few cells to log some other miscellaneous info.

timkpaine:
note that this code is a bit stale now, our internal fork puts the prelaunch hook here:
https://github.com/voila-dashboards/voila/blob/56656b7da48c7b094d4c78b016b121d1f07dd10f/voila/handler.py#L71

timkpaine:
@maartenbreddels @SylvainCorlay you can also pretty trivally use this to embed arguments. just install a variable in the kernel from handler.request.arguments in your hook

timkpaine:
for example:

```python3
def prelaunch_hook(handler, notebook, cwd):
    args = handler.request.arguments
    cell = NotebookNode()
    cell.metadata = {}
    cell.execution_count = 1
    cell.outputs = []
    cell.source = 'ARGS={}'.format(json.dumps(args))'
```
now `ARGS` is available in the kernel. Some other ideas include:
- headers: `handler.request.headers.get_all()`
- remote_ip: `handler.request.remote_ip`
- full url

maartenbreddels:
Hey Tim, I want to let you know that I still like the idea, but I'm still not sure of the consequences for voila long term, e.g. for caching, kernel-forking. Not forgotten about it.

voila/app.py
```diff
@@ -241,6 +241,11 @@ class Voila(Application):
                                  cannot be determined reliably by the Jupyter notebook server (proxified
                                  or containerized setups for example)."""))
 
+    prelaunch_hook = Any(default_value=None, allow_none=True,
```
martinRenou: Instead of `Any` here, maybe we should use `Instance(Callable)`.

`Callable` coming from the builtin `typing` module https://docs.python.org/3/library/typing.html#typing.Callable

timkpaine: changed

timkpaine:
@martinRenou theres a Callable traitlet added a while back but not in the docs still

timkpaine:
https://github.com/ipython/traitlets/issues/554

https://github.com/ipython/traitlets/issues/321

KKammoun:
Hi Ti, this looks very useful! How can I add a prelaunch hook when calling Voila from the Anaconda prompt? The python snippet you gave does not work for me from Jupyter or from Spyder console and I don't see anything about this method of starting Voila on the docs.

timkpaine:
reimplemented in https://github.com/voila-dashboards/voila/pull/724

PR #238, opened 2019-06-13
[WIP] Lab app

vidartf:
This is a work in progress on adding a jupyterlab plugin based way of adding dependencies and bundling them.

Noteworthy points:
- No method yet for a notebook to specify its dependencies (probably better in a separate issue). @jasongrout
- Currently adds lab shell to DOM as a hidden node, ideally we'd specify the shell ourselves to avoid putting any nodes with an `id` attribute on the page.
- Currently works with all examples that only use core widgets.
- There are some issues with duplication of dependencies due to lab/widgets using per-releases. Please do not consider merging before these are released in stable versions.
- Currently adds itself as a parallel bundle to the standard one, and overwriting some templates. This would need to be sorted out before merging, but allows for testing and comparison while working on it.


SylvainCorlay:
@vidartf thanks! This is really exciting.

Quick git-foo thingy: we had to change the git history of voila to get rid of a large file that was in the history and being downloaded with git clone, and it seems that you are getting some inconsistency in your PR because of it.

One way to fix it:

1) Fix your local master. Assuming `origin` is your fork and `upstream` is the main repo:

- git checkout master
- git fetch upstream
- git reset upstream/master --hard

2) Fix your branch

- git fetch upstream
- git checkout lab-app
- git reset upstream/master                               #  no --hard!
- git status (see your local changes)
- stage and commit changes
- git push origin lab-app -f

jtpio:
Posting here for reference.

In the [Voila weekly call yesterday](https://voila-dashboards.github.io/2020/07/13/weekly.html) we briefly discussed the possibility of using the refactored extension system in JupyterLab (planned for 3.0). So the lab bundles could be loaded in a similar way as with the classic notebook bundles (the way it is at the moment).

PR #240, opened 2019-06-13
Pass traitlets.config to ExecutePreprocessor

### LINKED ISSUE #125: Can user specify custom timeout limit?
```
[Voila] ERROR | Timeout waiting for execute reply (30s).
```

I don't see any option in `voila --help-all` that can allow me to change the 30s limit. My usage is `voila somenotebook.ipynb`

azjps:
Allow specifying viola `--ExecutePreprocessor.timeout=N` to change timeout of cell executions to N seconds (default is 30).

Closes #125


maartenbreddels:
I feel the name `exporter_config` now is off. We should maybe rename it to `config` or be more explicit and name it `traitlet_config`, since the handler is a non-traitlet class (https://traitlets.readthedocs.io/en/stable/config.html).

jeffyjefflabs:
Curious about the status of this - it would close #286 as well.

maartenbreddels:
Thank for your patience, looks good!

PR #241, opened 2019-06-14
Add instruction to install the test_template

jtpio:
Follow-up for https://github.com/QuantStack/voila/pull/207.

Add a quick note on how to install the `test_template` locally.

Since we now have `CONTRIBUTING.md` we can put it there until we find a way to install it automatically.


jtpio:
cc @martinRenou

Issue #231, opened 2019-06-13
Running the tests locally doesn't clean up kernels


jtpio:
Running the tests locally spins up a lot of kernels but the kernels are not properly terminated at the end of the run.

### Steps

1. Run the tests: `python -m pytest`
2. List voila processes: `ps auxf | grep voila`
3. Sample output:

```
jtp      19860  0.5  0.3 534676 58080 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_emfi2uwh/kernel-ca0eea8b-ae74-4d1e-9663-531e98ee1c32.json
jtp      19874  0.5  0.3 534676 58120 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_nyn7s5ca/kernel-ec998d36-9c87-4e62-ae00-abf8558f3655.json
jtp      19888  0.5  0.3 534676 58036 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_3fhelo7d/kernel-e213cc11-3397-405b-8527-b3a8a1fe7c96.json
jtp      19902  0.5  0.3 534676 58088 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_w08vytrr/kernel-14835642-0f4d-4757-a36e-1f4351f89663.json
jtp      19950  0.5  0.3 534420 58108 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_550qo66o/kernel-0d2582a4-6000-42fd-932a-3d2f965d4695.json
jtp      19992  0.6  0.3 534676 58052 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_18o_lt7l/kernel-fde143c2-2bb2-4b70-9479-f2f7604bf216.json
jtp      20006  0.6  0.3 534676 58096 ?        Ssl  10:42   0:00 /home/jtp/miniconda/envs/voila/bin/python -m ipykernel_launcher -f /tmp/voila_ohi7_wla/kernel-422f0318-116b-4236-8096-69cc60299979.json
[...]

PR #242, opened 2019-06-14
Shut down kernels after running the tests

jtpio:
Fixes https://github.com/QuantStack/voila/issues/231.

Two changes:

- Add a `stop` method to `VoilaApp` to shut down the kernels
- Call that method when in the tests before clearing the `voila_app` instance

It's probably simple enough for now to fix the issue, but we might want to reconsider this approach later on especially regarding `pytest-tornado` (as pointed out by @martinRenou in #231)

### DIFF
--- a/tests/app/conftest.py
+++ b/tests/app/conftest.py
@@ -41,6 +41,7 @@ def voila_app(voila_args, voila_config):
     voila_config(voila_app)
     voila_app.start()
     yield voila_app
+    voila_app.stop()
     voila_app.clear_instance()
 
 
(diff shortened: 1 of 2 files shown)


martinRenou:
Mmmmh. Tests are failing on my side. But I would not trust my conda environment this time :P Better trust the CI

jtpio:
Do the tests run in a new fresh environment, after running `python -m pip install ".[test]"` (without the `-e`) and the `test_template`?

maartenbreddels:
LGTM

maartenbreddels:
@martinRenou I'm assuming that is the case indeed :)

PR #243, opened 2019-06-14
Launch browser after listening

jtpio:
Launch the web browser only after the server is running (started to listen).

This prevents from opening a new tab when there is already an instance of voila running and another one is started afterwards using the same port (default for example).


maartenbreddels:
Thanks, indeed annoying behaviour!

PR #249, opened 2019-06-15
Fix support for Jupyter widget display in output widget

SylvainCorlay:
cc @jtpio 

Note: this should also resolve the issue that @cekees and @lheagy experienced with the `ipympl` widget and `@interact`.


SylvainCorlay:
OK, I could check that it worked with ipympl.

jtpio:
I can confirm it works also with the example posted in https://github.com/QuantStack/voila/pull/106#issue-269239380

Thanks!

PR #254, opened 2019-06-17
Use pytest-tornado instead of pytest-tornado5

jtpio:
We should be good to go with `pytest-tornado` now.


maartenbreddels:
Strange errors, I had to restart CI for a different PR a while ago, I hope this is not an issues in the testing.

jtpio:
Looking green now.

PR #262, opened 2019-06-19
Save the notebook before showing the Voila preview

### LINKED ISSUE #257: We should save the notebook when a user clicks the voila button
In the classical notebook, you can click the voila button, but it will not show the same thing as the notebook, I guess we should save it? I'm not sure what we do with the lab extension.

jtpio:
Fixes #257.

Automatically save the notebook before showing the voila preview.

## TODO

- [x] JupyterLab extension
- [x] Classic Notebook extension 

jlab-save-render

cc @wolfv

### DIFF
--- a/packages/jupyterlab-voila/src/index.ts
+++ b/packages/jupyterlab-voila/src/index.ts
@@ -104,6 +104,7 @@ const extension: JupyterLabPlugin<void> = {
         if (!current) {
           return;
         }
+        await current.context.save();
         const voilaPath = current.context.path;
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I can imagine some people would like a warning before saving, but in terms of usibility, I prefer not to. If we want to support this, we should have a dialog that says "don't ask again", but I am ok with this as it is. Very happy actually! :)

jtpio:
Good, then we can merge as it is and consider adding a warning later on if needed (with an option to disable it or "don't ask again").

PR #266, opened 2019-06-20
updates the docs logo

choldgraf:
This updates the logo in the top-left of the RTD page. The logo was added to the index page, but it will disappear when people leave that page. Here's what it looks like:

image

Another option would be to add in a white background and make it a PNG, but this PR just deals w/ the SVG that's already there

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,7 +44,7 @@ version = '.'.join(map(str, _release['version_info'][:2]))
 release = _release['__version__']
 language = None
 
-html_logo = 'quantstack-white.svg'
+html_logo = 'voila.svg'
 
 exclude_patterns = []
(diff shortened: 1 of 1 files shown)


choldgraf:
I think that travis build failure isn't related to this, can somebody restart it?

jtpio:
Thanks @choldgraf!

Build was restarted, but yes the failure looked unrelated to this.

SylvainCorlay:
Hey, thanks for the PR!

In all other projects, we have been using the organization logo and not the
project logo in the corner. At the moment it is also a good way to get
credit for QuantStack.

We may want to change these options if this moves to ProjectJupyter.

S.

choldgraf:
ah - I didn't realize this was intentional...

PR #268, opened 2019-06-25
JLab extension: render the preview on save

jtpio:
### Changes

- Add `plugin.json` to store the settings for the voila preview extension.
- Add the `renderOnSave` option to render the preview when the notebook is saved. Defaults to `false`.

image

### Follow-up

- Add a "Render on Save" checkbox in the toolbar?
- Make the voila preview panel a document widget extension?

### DIFF
--- a/packages/jupyterlab-voila/src/preview.tsx
+++ b/packages/jupyterlab-voila/src/preview.tsx
@@ -43,6 +43,7 @@ export class VoilaPreview extends MainAreaWidget<IFrame> {
     const renderOnSaveCheckbox = ReactWidget.create(
       <label>
         <input
+          style={{ verticalAlign: "middle" }}
           name="renderOnSave"
           type="checkbox"
           defaultChecked={renderOnSave}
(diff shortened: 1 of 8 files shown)


jtpio:
Added the "Render on Save" checkbox to the toolbar:

jlab-render-on-save

packages/jupyterlab-voila/src/preview.tsx
```diff
@@ -0,0 +1,106 @@
...
+            this.renderOnSave = event.target.checked;
+          }}
+        />
+        Render on Save
+      </label>
+    );
+    renderOnSaveCheckbox.addClass("jp-VoilaPreview-renderOnSave");
+
+    this.toolbar.addItem("reload", reloadButton);
+    this.toolbar.addItem("renderOnSave", renderOnSaveCheckbox);
```
maartenbreddels: jlab/phosphorjs supports react components?

maartenbreddels:
LGTM, do you wanna squash and merge?

packages/jupyterlab-voila/src/preview.tsx
jtpio: Yes (for jlab), most of the toolbar components are actually built with React (using the `ReactWidget` utility under the hood)

jtpio:
I think the commits are relatively small and self-contained for a normal merge (without squashing / losing history)?

Issue #141, opened 2019-05-16
require.js is loaded after plotly needs it


leogout:
I tried to use plotly in a notebook with the following code (from plotly's doc):
```python
import plotly.graph_objs as go
from plotly.offline import init_notebook_mode, iplot
import numpy as np

init_notebook_mode()

N = 500
random_x = np.linspace(0, 1, N)
random_y = np.random.randn(N)

trace = go.Scatter(
    x = random_x,
    y = random_y
)

iplot([trace], filename='basic-line')
```
It runs well in my local jupyterlab instance:
![image](https://user-images.githubusercontent.com/9843962/57858487-724d4100-77f1-11e9-8c93-00c2986afaaf.png)

But when I try to launch it with voila it gives me a blank output cell with the error:
![image](https://user-images.githubusercontent.com/9843962/57858041-98261600-77f0-11e9-9918-2d7766654418.png)

When I inspected the template I found that plotly tries to use require.js before it gets imported (in the base template it is at the [bottom of the file](https://github.com/QuantStack/voila/blob/master/share/jupyter/voila/template/default/nbconvert_templates/base.tpl#L38))

How can I get my graph to work with voila ?

leogout:
I just tried to put require.js in the head of the template (not ideal I know) but it works and my plot renders. I'll use a custom template for now but I would love to hear your thoughts about this.

leogout:
If someone else stumble upon this issue, what I did to fix it is:
Copy in place the folder `venv\share\jupyter\voila\template\default` with the name `plotly`.
Edit the `plotly\nbconvert_templates\base.tpl` file that you just copied accordingly to your needs. Personnaly, I needed to load `require.js` at the top of the template.
Then, run your notebook with: `voila your_notebook.ipynb --template=plotly`

scottlittle:
Thanks for sharing this. What was the explicit code that you used for loading `require.js` in your `base.tpl` file?

leogout:
In the template I overrided, I used this code:
```jinja
{%- extends 'basic.tpl' -%}
{% from 'mathjax.tpl' import mathjax %}

{%- block header -%}
<!DOCTYPE html>
<html>
<head>
{%- block html_head -%}
<meta charset="utf-8" />
{% set nb_title = nb.metadata.get('title', '') or resources['metadata']['name'] %}
<title>Voila: {{nb_title}}</title>

[...]

PR #275, opened 2019-07-01
requirejs should be loaded in the header with plotly

alexisduque:
this fix https://github.com/QuantStack/voila/issues/141


SylvainCorlay:
Thanks!

maartenbreddels:
Do you have an explanation for why this works?

alexisduque:
@maartenbreddels as explained in #141 plotly tries to use require.js before it gets imported since it is loaded in the footer in the template

PR #282, opened 2019-07-03
Drop HTML manager

SylvainCorlay:


vidartf:
Just for us following the repo: What was the motivation for this?

SylvainCorlay:
Thanks. We were moving fast for the fixup release. The html manager was just imported for the require loader, but it also loads some css (the labvariables polyfill) which breaks the voila theming.

PR #285, opened 2019-07-03
Don't include map files in distributions

xhochy:
This reduces the size of the `sdist` from `4.5M` to `1.3M` for me.

These files are quite useful during development but not needed by endusers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -254,6 +254,7 @@ def get_data_files():
     # Add all the templates
     for (dirpath, dirnames, filenames) in os.walk('share/jupyter/voila/templates/'):
         if filenames:
+            filenames = [f for f in filenames if not f.endswith('.map')]
             data_files.append((dirpath, [os.path.join(dirpath, filename) for filename in filenames]))
     return data_files
 


SylvainCorlay:
Wow I did not know map files were so big!

vidartf:
Source maps tend to be on the big side. There is one argument for distributing them normally, which would to get clearer errors copy-pasted by users from their browser console, but I'm not sure how much that is worth :)

SylvainCorlay:
Rather than deleting them, should we not produce them at all in "production
mode"?

(For publishing the npm package and the pypi package).

S.

xhochy:
An alternative approach would also be to distribute them in a separate package: https://github.com/conda-forge/mlflow-feedstock/pull/9

maartenbreddels:
> Rather than deleting them, should we not produce them at all in "production mode"?

You may still accidentally release them if they are generated (and didn't clean up).

I think we still want to have source maps at this moment to have reasonble error messages.

jtpio:
> There is one argument for distributing them normally, which would to get clearer errors copy-pasted by users from their browser console, but I'm not sure how much that is worth :)

This is generally useful, but since the voila frontend is rather minimal, it might not be worth shipping them?

SylvainCorlay:
@jtpio @maartenbreddels I am thinking of getting this in for 0.2.0. I like the reduction of the bundle size.

vidartf:
>  I like the reduction of the bundle size.

Just to make sure: You are talking about the sdist size right?

maartenbreddels:
> Just to make sure: You are talking about the sdist size right?

sdist and wheel (and thus conda package) I guess, what else could it be?

vidartf:
> 
> 
> > Just to make sure: You are talking about the sdist size right?
> 
> sdist and wheel (and thus conda package) I guess, what else could it be?

The published npm package? It's been so long I've lost all context here 😄

maartenbreddels:
I'd say we merge this, agree @SylvainCorlay?. I prefer this solution to not generating the map files since it might be useful for local debugging, and if it gets accidentally gets generated, we don't accidentally distribute it.

maartenbreddels:
Do we want to distribute this in `voila-sourcemap` maybe? That we release for each version in tandem?

maartenbreddels:
Let do a release with this first, it's easy to revert or change.

PR #290, opened 2019-07-08
Remove reference to jquery

### LINKED ISSUE #288: $ undefined
Now that jquery has left the page, we get many warning, that $ is not defined, from:
```
<script type="text/javascript">
var element = $('#3f949d58-54e6-4a0e-905e-e8382b9e53e4');
</script>
```
Which is something nbconvert outputs. Can we simply get rid of this in the voila template? I have no idea what this should do.

jtpio:
Fixes #288.


SylvainCorlay:
Thx!

PR #292, opened 2019-07-08
Move js/ (voila standalone frontend) to packages/

jtpio:
Move the `js` folder to `packages`, so both the standalone frontend and the JupyterLab extension sit next to each other.

A follow-up for this change would be to slightly rework `setup.py` to also run `npm pack` for the JupyterLab extension and install it automatically as `data_file`.


jtpio:
Test failure looks relevant.

`tests/app/image_inlining_test.py` passes locally when executed alone (`python -m pytest -k "test_image_inlining"`)

But it doesn't pass when started with `python -m pytest` and running after `tests/app/execute_test.py` (which errors with `..Task was destroyed but it is pending`)

maartenbreddels:
It also seems to be failing in master, so might be an external dependency, jupyter_client is what @SylvainCorlay suggested, but the inlining might be nbconvert?

SylvainCorlay:
There was an nbconvert release?

maartenbreddels:
Not that i know of or checked, but I doubt the inlining error has something to do with jupyter_client, although not impossible.

jtpio:
Rebased to check if the tests would pass.

maartenbreddels:
should we directly move it to the proper name? `packages/voila`?

jtpio:
`packages/voila` sounds like a better name indeed.

jtpio:
Closing in favor of #786.

PR #293, opened 2019-07-08
Test pinning jupyter_client

SylvainCorlay:
To determine whether the test breakage was due to the 5.3.0 release.


jtpio:
I restarted one of the tests (random timeout), and they are now passing.

SylvainCorlay:
Thanks. We have investigated the issue with @JohanMabille  and `jupyter_client` 5.3 has some issues with the closing of the App. It should not impact voila's behavior so I don't think that we should merge this.

PR #294, opened 2019-07-09
Add instructions to install dependencies of example notebooks

mkcor:
Hello,

I made it clear that you need dependencies (as documented in `environment.yml`) to be able to run the demo notebooks (which live in `./notebooks/`).


mkcor:
Or maybe this should be merged with https://github.com/QuantStack/voila/blob/master/CONTRIBUTING.md#running-the-examples ...?

SylvainCorlay:
ignoring the test failure which is all due to the jupyter_client 5.3.0 catastrophy. Merging!

PR #295, opened 2019-07-09
Increase ulimit for testing

SylvainCorlay:
Should fix the tests on OS X.


SylvainCorlay:
@JohanMabille we need to investigate if jupyter_client 5.3.1 properly closes all sockets.

It is possible that the failure to close socket is the cause to the test failures on master.

Issue #224, opened 2019-06-12
500: Internal Server Error


amrrs:
Constantly getting this error - tried updating `nbconvert`, renaming `filename` nothing changed

```
[I 15:33:00.226 NotebookApp] Executing notebook with kernel:
[E 15:33:05.055 NotebookApp] Uncaught exception GET /voila/render/packt/Pycon_Ticket.ipynb (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/voila/render/packt/Pycon_Ticket.ipynb', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.7/site-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/usr/local/lib/python3.7/site-packages/tornado/gen.py", line 1133, in run
        value = future.result()
      File "/usr/local/lib/python3.7/site-packages/tornado/gen.py", line 1147, in run
        yielded = self.gen.send(value)
      File "/usr/local/lib/python3.7/site-packages/voila/handler.py", line 72, in get
        contents_manager=self.contents_manager  # for the image inlining
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/templateexporter.py", line 255, in __init__
        super(TemplateExporter, self).__init__(config=config, **kw)
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/exporter.py", line 110, in __init__
        self._init_preprocessors()
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/exporter.py", line 264, in _init_preprocessors
        self.register_preprocessor(preprocessor, enabled=True)
      File "/usr/local/lib/python3.7/site-packages/nbconvert/exporters/exporter.py", line 225, in register_preprocessor
        preprocessor_cls = import_item(preprocessor)
      File "/usr/local/lib/python3.7/site-packages/traitlets/utils/importstring.py", line 34, in import_item
        module = __import__(package, fromlist=[obj])
      File "/usr/local/lib/python3.7/site-packages/voila/csspreprocessor.py", line 13, in <module>
        from jupyterlab_pygments import JupyterStyle
      File "/usr/local/lib/python3.7/site-packages/jupyterlab_pygments/__init__.py", line 9, in <module>
        from .style import JupyterStyle
[...]

PR #297, opened 2019-07-11
Add pygments to install_requires

jtpio:
Add an explicit requirement for the transitive `pygments` dependency (from [jupyterlab_pygment](https://github.com/jupyterlab/jupyterlab_pygments/blob/b6033176746b1b8405c2abe842f42d88b296fc1a/setup.py#L22)).

This error seems to occur quite frequently when installing voila via pip.

Related to #224, https://github.com/voila-gallery/gallery/issues/58, https://github.com/voila-gallery/gallery/issues/54.


SylvainCorlay:
Thanks!

PR #298, opened 2019-07-11
Let exporter inherit from SlidesExporter to expose more resources

mkcor:
Hello,

With this change, we are paving the way for the reveal template. Without it, using the reveal template yields a 500 Internal Server Error.

Steps to reproduce:

* Copy directory `share/jupyter/voila/template/reveal` from https://github.com/QuantStack/voila/pull/54/files into `PREFIX/share/jupyter/voila/templates/`;
* Get `notebooks/rise_example.ipynb` from https://github.com/QuantStack/voila/pull/54/files and (from where the notebook lives) run
```
voila rise_example.ipynb --template=reveal
```

We now have
```py
ipdb> VoilaExporter.mro()                                                                                                                  
[<class 'voila.exporter.VoilaExporter'>, <class 'nbconvert.exporters.slides.SlidesExporter'>, <class 'nbconvert.exporters.html.HTMLExporter'>, <class 'nbconvert.exporters.templateexporter.TemplateExporter'>, <class 'nbconvert.exporters.exporter.Exporter'>, <class 'traitlets.config.configurable.LoggingConfigurable'>, <class 'traitlets.config.configurable.Configurable'>, <class 'traitlets.traitlets.HasTraits'>, <class 'traitlets.traitlets.HasDescriptors'>, <class 'object'>]
```
instead of
```py
ipdb> VoilaExporter.mro()                                                                                                                  
[<class 'voila.exporter.VoilaExporter'>, <class 'nbconvert.exporters.html.HTMLExporter'>, <class 'nbconvert.exporters.templateexporter.TemplateExporter'>, <class 'nbconvert.exporters.exporter.Exporter'>, <class 'traitlets.config.configurable.LoggingConfigurable'>, <class 'traitlets.config.configurable.Configurable'>, <class 'traitlets.traitlets.HasTraits'>, <class 'traitlets.traitlets.HasDescriptors'>, <class 'object'>]
```

/cc @maartenbreddels

PR #301, opened 2019-07-12
Passing custom resources to voila configuration

mkcor:
Hello,

This PR overwrites #298 (previously submitted). Example use:
```
voila reveal.ipynb --template=reveal --reveal_transition=fade
```
The config additions are almost copied-pasted from https://github.com/jupyter/nbconvert/blob/76061f8df164a742fc039b2574508bd8cf7d96a1/nbconvert/exporters/slides.py#L105-L130 -- should I add a comment about it?

/cc @maartenbreddels

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -12,8 +12,6 @@ import tornado.web
 
 from jupyter_server.base.handlers import JupyterHandler
 
-from nbconvert.exporters.slides import prepare
-
 from .execute import executenb
 from .exporter import VoilaExporter
 
@@ -80,7 +78,7 @@ class VoilaHandler(JupyterHandler):
         resources.update(resources_reveal)
 
         # include potential extra resources
-        extra_resources = self.voila_configuration.extra_resources
+        extra_resources = self.voila_configuration.resources
         if extra_resources:
             for k in extra_resources.keys():
                 if k in resources.keys():
@@ -114,10 +112,6 @@ class VoilaHandler(JupyterHandler):
                 or not exporter.exclude_input                   # keep cell if input not excluded
             )
         result.cells = list(filter(filter_empty_code_cells, result.cells))
-
-        # add convenience metadata on cells for reveal template
-        if self.voila_configuration.template == 'reveal':
-            result = prepare(result)
 
         html, resources = exporter.from_notebook_node(result, resources=resources)
 
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -6,25 +6,15 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
-import collections
 import os
 
 import tornado.web
 
 from jupyter_server.base.handlers import JupyterHandler
+from jupyter_server.config_manager import recursive_update
 
 from .execute import executenb
 from .exporter import VoilaExporter
-
-
-def update_nested_dict(d, u):
-    """Update (nested) dictionaries: update keys down to any depth."""
-    for k, v in u.items():
-        if isinstance(v, collections.Mapping):
-            d[k] = update_nested_dict(d.get(k, {}), v)
-        else:
-            d[k] = v
-    return d
 
 
 class VoilaHandler(JupyterHandler):
@@ -91,7 +81,7 @@ class VoilaHandler(JupyterHandler):
         # include potential extra resources
         extra_resources = self.voila_configuration.resources
         if extra_resources:
-            resources = update_nested_dict(resources, extra_resources)
+            recursive_update(resources, extra_resources)
 
         exporter = VoilaExporter(
             template_path=self.nbconvert_template_paths,
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
+import collections
 import os
 
 import tornado.web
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
I like the idea of allowing to specify resources through the `VoilaConfiguration` object. An issue here is that this adds reveal-specific configurables to voila-core, which we cannot do for all possible templates.

Maybe a first way to do this would have the `VoilaConfiguration` have a `resources` `Dict` trait, which would not be reveal-specific. This may make setting options with the command line a bit difficult though.

mkcor:
@SylvainCorlay we are on the same wavelength... It's cumbersome to write JSON at the command line, but maybe great docs could make up for this inconvenience (I guess users will end up copying-pasting...).

Ok, so I'll implement the `Dict` trait.

SylvainCorlay:
I wonder if we could hack traitlets to have a `Bunch` syntax in the command line so that we can do

```
voila --foo.bar=baz
```

for any valid name `bar`.

mkcor:
Oh, I hadn't seen your comment about `Bunch` (which I hadn't heard of before). Please see 4144ed6.

I can see three additional improvements which would belong in this PR.

* Modularization (abstraction): replace this [if block](https://github.com/QuantStack/voila/pull/301/files#diff-da3df63b00696d9c775762da2f14fa41R84) by a function call, say `resources.dict_merge(extra_resources)` along the lines of these [suggestions](https://gist.github.com/angstwad/bf22d1822c38a92ec0a9); is it a good idea to start a `utils.py` submodule in voila?
* Validation: ideally we would want to check the validity of the values passed at the CLI and fall back on our defaults (when present) if a value isn't valid... 
* Instructions: I could add a "Coming soon!" subsection in the README to showcase that extra resources can be configured (even if the reveal template isn't published yet).

voila/handler.py
```diff
@@ -67,6 +67,32 @@ def get(self, path=None):
...
+                    if isinstance(extra_resources[k], dict):
+                        for kk in extra_resources[k].keys():
+                            # update or add depth-2 resources
+                            resources[k][kk] = extra_resources[k][kk]
+                        else:
+                            # update depth-1 (top-level) resources
+                            resources[k] = extra_resources[k]
+                else:
+                    # add depth-1 (top-level) resources
+                    resources[k] = extra_resources[k]
```
SylvainCorlay: Maybe we can do a sort of recursive extend.

From the SO answer here: 
https://stackoverflow.com/questions/3232943/update-value-of-a-nested-dictionary-of-varying-depth

Python 2

```python
import collections

def update(d, u):
    for k, v in u.iteritems():
        if isinstance(v, collections.Mapping):
            d[k] = update(d.get(k, {}), v)
        else:
            d[k] = v
    return d
```

Python 3:

```python
import collections

def update(d, u):
    for k, v in u.items():
        if isinstance(v, collections.Mapping):
            d[k] = update(d.get(k, {}), v)
        else:
            d[k] = v
    return d
```

SylvainCorlay:
This is looking good to me. I left an inline comment about the sort of recursive `dict.extend` that you are doing.

Small note: I rebased the PR on master and add a commit removing the `prepare` reveal-specific stage. Also, `extra_resources` is directly called `resources`.

voila/handler.py
mkcor: Thanks, @SylvainCorlay !

I have named this utility function `update_nested_dict()` to be more explicit and not to step on `dict.update()`'s toes. I guess we don't need to support Python 2... am I right?

SylvainCorlay:
This looks good to me.

@maartenbreddels let me know if you have any comment on this before I merge. I would like to have a release including this so that @mkcor can iterate on the reveal template work.

voila/handler.py
```diff
@@ -16,6 +17,16 @@
...
 
 
+def update_nested_dict(d, u):
+    """Update (nested) dictionaries: update keys down to any depth."""
+    for k, v in u.items():
+        if isinstance(v, collections.Mapping):
+            d[k] = update_nested_dict(d.get(k, {}), v)
+        else:
+            d[k] = v
+    return d
```
maartenbreddels: Can we use https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/config_manager.py#L19 ?

SylvainCorlay: Indeed I did not think of the config manager merging mechanism.

@maartenbreddels can you push to this branch making that change?

mkcor: @maartenbreddels Thank you! A helper function to update dictionaries recursively had to exist 'nearby' indeed...

@SylvainCorlay Done via 6ea5bc7 (function call changed because the config manager's function is in-place).

maartenbreddels: Great! I happen to know because I worked on that module :)

maartenbreddels:
LGTM, the failure was a timeout in nbconvert (we might want to increase that limit), restarted the travis job.

voila/handler.py
maartenbreddels: We still have py2 support, but your last commit fixed this.

SylvainCorlay:
OK, I pushed a commit removing the `resources_reveal` dictionary, which is specific to the *reveal* template.

(Since we can't elect any template-specific things to be part of the main voila package).

maartenbreddels:
Good point. I think if templates want to have specific resources, we could add that to conf.json, at least for defaults.

SylvainCorlay:
So the idea is that the reveal template should work without the resource dictionary providing `theme`, `transition` and `scroll`. In the voila template file, using `reveal.get('theme', 'simple')` instead of `reveal.theme`.

mkcor:
Example use (update):
```
voila reveal.ipynb --template=reveal --VoilaConfiguration.resources="{'reveal': {'transition': 'fade'}}"
```

PR #305, opened 2019-07-15
Monochrome voila icon for the notebook toolbar

jtpio:
The grey and orange default icon wouldn't render nice when using the Dark theme.

Switching to light and dark monochrome icons that follow the style of the existing Toolbar items.

image

image

### DIFF
--- a/packages/jupyterlab-voila/style/index.css
+++ b/packages/jupyterlab-voila/style/index.css
@@ -1,3 +1,5 @@
+@import "variables.css";
+
 .jp-VoilaIcon {
-  background-image: url("./voila.svg");
+  background-image: var(--jp-icon-voila);
 }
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Seeing the same failure on the same setup (osx py37) in other PR's, not sure if it is a fluke (the timeout limit too small) or consistent failures.

jtpio:
Yes I've seen those too. Restarted the job and it passed.

maartenbreddels:
Ok, good to know, so we should fix that. Do you want me to take a look at it?

jtpio:
Sure. A first guess for now is that it could be related to the way kernels are shut down (`stop` method) and how `pytest-tornado` manages the event loop.

jtpio:
@maartenbreddels I think this is ready for review / merge

maartenbreddels:
Work here 👍
I do miss the orange now that I see it for real.
@SylvainCorlay are you ok with the change of orange to monochrome, if yes, you can merge it.

SylvainCorlay:
Looks all great. Merging. We will update things when we adopt the new design.

PR #307, opened 2019-07-15
Add --debug flag

jtpio:
Add a `--debug` flag to make it more convenient to start voila in debug mode.

### Usage

```
voila --debug
```


maartenbreddels:
Cool, I didn't know this was possible! Failure is unrelated btw.

jtpio:
Yes. And it should also automatically be made available when https://github.com/QuantStack/voila/pull/270 lands (along with other flags).

PR #309, opened 2019-07-17
Handle non-existing kernels

### LINKED ISSUE #259: Work with non-existing kernels
The notebook is quite forgiving when the kernel name does not match the kernels installed, while voila give a 500 (http) error, I guess we can copy the logic of the classical notebook or lab on how to deal with that.

jtpio:
Fixes #259. It should also provide a simple solution for https://github.com/QuantStack/voila/issues/267.

This PR makes use of the `KernelSpecManager` to first check that the requested kernel name (from the notebook metadata) is available. If it isn't then it will default to the default kernel name.

I also got hit by that recently while using `xeus-python` and decided to give it a shot.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -64,6 +64,7 @@ class VoilaHandler(JupyterHandler):
                 if kernel["spec"]["language"].lower() == kernel_language
             ]
             if matches:
+                matches.sort(key=lambda name: all_kernel_specs[name]["spec"]["display_name"])
                 kernel_name = matches[0]
 
         # Launch kernel and execute notebook
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -132,7 +132,7 @@ class VoilaHandler(JupyterHandler):
 
     @tornado.gen.coroutine
     def find_kernel_name_for_language(self, kernel_language, kernel_specs=None):
-        """Finds a best matching kernel name given a kernel languages.
+        """Finds a best matching kernel name given a kernel language.
 
         If multiple kernels matches are found, we try to return the same kernel name each time.
         """
@@ -147,4 +147,4 @@ class VoilaHandler(JupyterHandler):
             matches.sort(key=lambda name: kernel_specs[name]["spec"]["display_name"])
             raise tornado.gen.Return(matches[0])  # TODO py2: replace by return
         else:
-            raise tornado.web.HTTPError(500, 'No Jupyter kernel for language %r not found' % kernel_language)
+            raise tornado.web.HTTPError(500, 'No Jupyter kernel for language %r found' % kernel_language)
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -128,7 +128,7 @@ class VoilaHandler(JupyterHandler):
         notebook.metadata.kernelspec.name = kernel_name
         notebook.metadata.kernelspec.display_name = all_kernel_specs[kernel_name]['spec']['display_name']
         notebook.metadata.kernelspec.language = all_kernel_specs[kernel_name]['spec']['language']
-        return notebook
+        raise tornado.gen.Return(notebook)  # TODO py2: replace by return
 
     @tornado.gen.coroutine
     def find_kernel_name_for_language(self, kernel_language, kernel_specs=None):
@@ -145,6 +145,6 @@ class VoilaHandler(JupyterHandler):
         if matches:
             # Sort by display name to get the same kernel each time.
             matches.sort(key=lambda name: kernel_specs[name]["spec"]["display_name"])
-            return matches[0]
+            raise tornado.gen.Return(matches[0])  # TODO py2: replace by return
         else:
             raise tornado.web.HTTPError(500, 'No Jupyter kernel for language %r not found' % kernel_language)
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -52,13 +52,19 @@ class VoilaHandler(JupyterHandler):
             raise tornado.web.HTTPError(404, 'file not found')
 
         # Fetch kernel name from the notebook metadata
-        default_kernel_name = self.kernel_manager.default_kernel_name
-        kernel_name = notebook.metadata.get('kernelspec', {}).get('name', default_kernel_name)
+        kernelspec = notebook.metadata.get('kernelspec', {})
+        kernel_name = kernelspec.get('name', self.kernel_manager.default_kernel_name)
 
-        # Use the default kernel name if the kernel from the metadata is not in the kernel specs
+        # Find a spec matching the language if the kernel name does not exist in the kernelspecs
         all_kernel_specs = yield tornado.gen.maybe_future(self.kernel_spec_manager.get_all_specs())
         if kernel_name not in all_kernel_specs:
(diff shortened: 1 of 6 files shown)


jtpio:
Another approach would be to make use of the `language` field from the `kernelspec` metadata to find a spec that matches the same language.

SylvainCorlay:
What does the classic notebook do? It appears to be quite liberal with kernelspecs.

jtpio:
It looks like much of the logic for dealing with missing specs in the classic notebook is actually handled by the frontend.

This is what it appears to be doing:

1. The frontend fetches the list of `kernelspecs`. When a spec is not found it logs it in the dev tools console: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/kernelselector.js#L226-L227

2. It then defaults to the first match using the `language` field, and emits the `spec_changed.Kernel` event: https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/static/notebook/js/kernelselector.js#L249

3. The `spec_changed.Kernel` event is then consumed here: https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/notebook.js#L345-L357

4. A new session is started using the name from 2.: https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/static/notebook/js/notebook.js#L356

I might have missed some details but this is basically the outcome of stepping through the code in the dev tools.

SylvainCorlay:
Thanks for looking into this. What do you think is the cleanest thing to do?

It seems that jupyterlab pops up a dialog to select a kernel.

Or such a logic could be in the backend?

jtpio:
Finding a match based on the `language` field would certainly help with many of the common issues related to:

- having `"name": "python2"` in the notebook metadata
- conda named kernels: `"name": "conda-env-foo-py"`

And in some way it follows the logic from the classic notebook.

For the case where there is no match for a given `language`, for example trying to execute a C++ notebook on a Voila setup that only has Python kernels installed, this would still return a 500.

jtpio:
The last commits implement finding a spec matching the `language` field from the metadata, as well as sorting the matches by display name (cf [notebook implementation](https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/static/notebook/js/kernelselector.js#L215-L223)).

If there is no `language` in the metadata, it defaults to `python`.

voila/handler.py
```diff
@@ -51,7 +51,20 @@ def get(self, path=None):
...
+        all_kernel_specs = yield tornado.gen.maybe_future(self.kernel_spec_manager.get_all_specs())
+        if kernel_name not in all_kernel_specs:
+            kernel_language = kernelspec.get('language', 'python').lower()
+            matches = [
+                name for name, kernel in all_kernel_specs.items()
+                if kernel["spec"]["language"].lower() == kernel_language
+            ]
+            if matches:
+                matches.sort(key=lambda name: all_kernel_specs[name]["spec"]["display_name"])
+                kernel_name = matches[0]
```
maartenbreddels: I wonder if we have 0 matches, if we can have a custom template with a proper error msg and http error code: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

If there are 1 or more matches, I think we should log a warning, so users are aware of it (I can imagine someone expecting it to run with a conda-env-kernel, which cannot be found, leading to surprises).

voila/handler.py
```diff
@@ -51,7 +51,20 @@ def get(self, path=None):
             raise tornado.web.HTTPError(404, 'file not found')
 
         # Fetch kernel name from the notebook metadata
-        kernel_name = notebook.metadata.get('kernelspec', {}).get('name', self.kernel_manager.default_kernel_name)
+        kernelspec = notebook.metadata.get('kernelspec', {})
+        kernel_name = kernelspec.get('name', self.kernel_manager.default_kernel_name)
+
+        # Find a spec matching the language if the kernel name does not exist in the kernelspecs
+        all_kernel_specs = yield tornado.gen.maybe_future(self.kernel_spec_manager.get_all_specs())
```
maartenbreddels: Why is `yield tornado.gen.maybe_future` needed? The current implementation gives a plain dict, or can other implementations return a future?

voila/handler.py
jtpio: Agreed, a warning would definitely be useful to let the users know the kernel being used is not the one they expect.

Having a custom template with a proper error message is also a good idea, but maybe for a follow-up PR?

voila/handler.py
jtpio: The kernel spec manager could be a `GatewayKernelSpecManager`, in which case the `maybe_future` would be needed: https://github.com/jupyter/notebook/blob/e498de6775b3de57f5ff827e562c179b17893064/notebook/gateway/managers.py#L521

jtpio:
Not sure it was discussed before, but do we need to keep supporting Python 2?

maartenbreddels:
Not really, but we have it now, and it doesn't lead yet to any major hurdles. I'd say the incentive to keep supporting is decaying exponentially though.

voila/handler.py
```diff
@@ -98,3 +92,59 @@ def filter_empty_code_cells(cell):
...
+        matches = [
+            name for name, kernel in kernel_specs.items()
+            if kernel["spec"]["language"].lower() == kernel_language
+        ]
+        if matches:
+            # Sort by display name to get the same kernel each time.
+            matches.sort(key=lambda name: kernel_specs[name]["spec"]["display_name"])
+            raise tornado.gen.Return(matches[0])  # TODO py2: replace by return
+        else:
+            raise tornado.web.HTTPError(500, 'No Jupyter kernel for language %r not found' % kernel_language)
```
jtpio: Looks like a double negation in the error message. Maybe something like this?

```suggestion
            raise tornado.web.HTTPError(500, 'Jupyter kernel for language %r not found' % kernel_language)
```

voila/handler.py
```diff
@@ -98,3 +92,59 @@ def filter_empty_code_cells(cell):
...
+            self.log.warning('Could not find a kernel named %r, will use  %r', missing_kernel_name, kernel_name)
+        # We make sure the notebook's kernelspec is correct
+        notebook.metadata.kernelspec.name = kernel_name
+        notebook.metadata.kernelspec.display_name = all_kernel_specs[kernel_name]['spec']['display_name']
+        notebook.metadata.kernelspec.language = all_kernel_specs[kernel_name]['spec']['language']
+        raise tornado.gen.Return(notebook)  # TODO py2: replace by return
+
+    @tornado.gen.coroutine
+    def find_kernel_name_for_language(self, kernel_language, kernel_specs=None):
+        """Finds a best matching kernel name given a kernel languages.
```
jtpio: Minor typo:

```suggestion
        """Finds a best matching kernel name given a kernel language.
```

jtpio:
Right, the hurdles are quite minimal for now.

jtpio:
@maartenbreddels this looks good to me.

Returning a 500 when there is no matching kernel is definitely useful.
The other case (showing a warning when 1+ matches found) could be added in a different PR.

PR #314, opened 2019-07-19
Clear output before executing the notebook

### LINKED ISSUE #214: Clear cell output before execute?
Would it make sense for voila to clear all cell outputs before executing the notebook?  If there's an exception part of the way through execution, the rendered notebook will contain stale outputs from wherever it was executed previously, and widgets defined after the point of the exception don't render.  It's potentially confusing to the end user.  I suppose if you're using standalone voila in a production environment, you'd probably clear outputs in the notebook before saving it, but you might not think to do that with the server extension. 

Short example below.  Run both cells in normal Jupyter first, then save and launch in voila.

```
# cell 1
import datetime
print(datetime.datetime.now())
raise RuntimeError

# cell 2
import ipywidgets as widgets
print(datetime.datetime.now())
widgets.Button(description='hello')
```

jeffyjefflabs:
If there is an exception during execution, output from a previous execution could be displayed in the voila dashboard.  Fixes #214.

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
+from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError, ExecutePreprocessor
 
 from ipykernel.jsonutil import json_clean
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks @jeffyjefflabs. This looks good to me!

PR #316, opened 2019-07-21
Voila manager constructor

SylvainCorlay:
This came up as I am updating thebelab for jupyterlab 1.0. (Trying to make the code bases converge.)

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -3211,17 +3211,6 @@
       "integrity": "sha512-d3OEjQV4ROpoflsnUA8HozoIR504TFxNivYEUi6uwz0IYhBkTDXGuWlNdMtybRt3nqVx/L6XqMt0FxkXuWKZhw==",
       "dev": true
     },
-    "ajv": {
-      "version": "5.1.6",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.1.6.tgz",
-      "integrity": "sha1-Sy8aGd7Ok9V6whYDfj6XkcfdFWQ=",
-      "dev": true,
-      "requires": {
-        "co": "^4.6.0",
-        "json-schema-traverse": "^0.3.0",
-        "json-stable-stringify": "^1.0.1"
-      }
-    },
     "ajv-errors": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/ajv-errors/-/ajv-errors-1.0.1.tgz",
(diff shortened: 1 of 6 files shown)


SylvainCorlay:
Double tested with command line app, jupyter server and notebook. Merging.

maartenbreddels:
I think we should not have changed the API, but make these optional arguments. This now breaks other templates. What do you think?

SylvainCorlay:
> This now breaks other templates.

Are we using the voila `WidgetManager` explicitly in templates?

SylvainCorlay:
> Are we using the voila WidgetManager explicitly in templates?

I guess we are for templates that don't use main.js...

maartenbreddels:
Yes, vuetify templates are using that API, and the React versions will do so as well.

It's easy to fix, but would be nice if we can keep the backward compatible in the future.

SylvainCorlay:
What do you think is the way forward. This change (taking out the context and all) is also required if we want to move to typescript. This also makes it possible to use the same widget manager for thebelab (I am working on this).

Should we consider that we are moving towards 0.2 (including the progressive rendering)?

maartenbreddels:
I am ok with this breaking change, as long as we don't have to do it often, and we consider this part of the public API.

What I like about having this as an API, is that people can have some custom js (like we do with the vuetify template), without having to create a new bundle.

SylvainCorlay:
> I am ok with this breaking change, as long as we don't have to do it often, and we consider this part of the public API.

Yes, I meant we take the occasion of the 0.2 release (which should have the progressive rendering) to introduce that kind of changes?

Or I could make the arguments as optional (and fill them in if not provided), while deprecating the possibility to not provide then...

maartenbreddels:
> Or I could make the arguments as optional (and fill them in if not provided)

That was my initial idea as well, but the first argument has changed meaning, so maybe indeed a 0.2 would be a good idea.

Working on the template refactor, after which the progressive rendering can start.

jtpio:
This also affects the [voila-adminlte](https://github.com/martinRenou/voila-adminlte) template:

https://github.com/martinRenou/voila-adminlte/blob/5b54301b040296a494af520ccbf2efd153f00688/share/jupyter/voila/templates/adminlte/nbconvert_templates/util.js#L23

cc @martinRenou

maartenbreddels:
Maybe it would be nice to have the 2nd and 3rd argument with default, so we don't have to copy-paste too much? Although the session object is also a large chunk of code.

SylvainCorlay:
Absolutely.

Setting defaults before calling the base class constructor prevents conversion to typescript.

maartenbreddels:
Ok, I didn't realize this is already released, which means none of the demos with custom templates that use this API work (cc @martinRenou )

maartenbreddels:
> prevents conversion to typescript.

Is the plan to move to typescript?

vidartf:
Just FYI: The thing that typescript complains about, is a JS issue as well, namely if you use `this` before calling `super` in the constructor: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super#Description

PR #317, opened 2019-07-22
feat: make the markdown rendered configurable in VoilaExporter

### LINKED ISSUE #234: Support specifying a custom markdown renderer in Voila
We need to make the markdown renderer a configurable option for the VoilaExporter so that third party can provide their own implementation and create custom handling of e.g. links in markdown cells.

maartenbreddels:
Example configuration (e.g. in voila.json) could be:

```
'VoilaExporter': {
    'markdown_renderer_class': 'my.package.MyMarkdownRenderer'
}
```
Closes #234

We could argue that we want this to be in nbconvert, but we can have it in voila first, upstream it later without having compatibility issues.

### DIFF
--- a/voila/exporter.py
+++ b/voila/exporter.py
@@ -33,16 +33,18 @@ class VoilaMarkdownRenderer(IPythonRenderer):
 class VoilaExporter(HTMLExporter):
     """Custom HTMLExporter that inlines the images using VoilaMarkdownRenderer"""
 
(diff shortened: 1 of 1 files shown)


cquah:
Thanks @maartenbreddels ! @aschlaep for awareness

SylvainCorlay:
This looks good to me. Thanks for tackling this!

PR #321, opened 2019-07-24
test: set timeout of pytest_tornado to 20 seconds to avoid false CI failured

### LINKED ISSUE #320: Unittest sometimes fail (timeout)
This happens consistently on travis+OSX+py37. Maybe this is due to a specific package only being installed with a particular version on that setup.
Looking at:
https://travis-ci.org/QuantStack/voila/jobs/560894161
it seems it has to do with `pytest-tornado`. 
I first thought it might be related to https://github.com/jupyter/nbconvert/issues/1067 but that seems unlikely given the error msg.

maartenbreddels:
Closes #320

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,5 +26,5 @@ install:
 before_script:
   - flake8 voila tests setup.py
 script:
-  - VOILA_TEST_DEBUG=1 py.test tests/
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I had to restart again, but had a different failure, it should at least give us fewer issues with CI, or get the real underlying error message.

PR #322, opened 2019-07-24
Fix Windows develop install

### LINKED ISSUE #192: Dev install cannot find npm
When trying to run `pip install -e .` on Windows, I get the following error:

```
C:\dev\voila [master ≡ +3 ~3 -0 !]> pip install -e .
Obtaining file:///C:/dev/voila
    Complete output from command python setup.py egg_info:
    running egg_info
    running css
    running jsdeps
    `npm` unavailable.  If you're running this command using sudo, make sure `npm` is available to sudo
    rebuilding js and css failed
    missing files: ['C:\\dev\\voila\\share\\jupyter\\voila\\templates\\default\\static\\voila.js']
      File "<string>", line 1, in <module>
      File "C:\dev\voila\setup.py", line 290, in <module>
        setup(**setup_args)
      File "C:\Miniconda3\lib\site-packages\setuptools\__init__.py", line 143, in setup
        return distutils.core.setup(**attrs)
      File "C:\Miniconda3\lib\distutils\core.py", line 148, in setup
        dist.run_commands()
      File "C:\Miniconda3\lib\distutils\dist.py", line 966, in run_commands
        self.run_command(cmd)
      File "C:\Miniconda3\lib\distutils\dist.py", line 985, in run_command
        cmd_obj.run()
      File "C:\dev\voila\setup.py", line 211, in run
        return command.run(self)
      File "C:\dev\voila\setup.py", line 55, in run
        raise e
      File "C:\dev\voila\setup.py", line 48, in run
        self.distribution.run_command('jsdeps')
      File "C:\Miniconda3\lib\distutils\dist.py", line 985, in run_command
        cmd_obj.run()
      File "C:\dev\voila\setup.py", line 128, in run
        raise ValueError(msg)
    ValueError: Missing file: C:\dev\voila\share\jupyter\voila\templates\default\static\voila.js
    npm is required to build a development version

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\dev\voila\
```

vidartf:
Fixes #192.

Reuses certain parts from [jupyter-packaging](https://github.com/jupyter/jupyter-packaging), with some small improvements that I will also contribute back.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,6 +123,7 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None):
                     return name
     return None
 
+
 # TODO: remove this function once we can depend on jupyter_packing, see:
 #  https://github.com/QuantStack/voila/pull/322
 # `run` function copied from jupyter_packaging under the following license:
--- a/setup.py
+++ b/setup.py
@@ -122,7 +122,6 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None):
     return None
 
 
-
 # `run` function copied from jupyter_packaging under the following license:
 # -------------------------------------------------------------------------
 #
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,8 @@ def update_package_data(distribution):
     build_py.finalize_options()
 
 
+# TODO: remove this function once we drop Python2, see:
+#  https://github.com/QuantStack/voila/pull/322
 # `shutils.which` function copied verbatim from the Python-3.3 source.
 def which(cmd, mode=os.F_OK | os.X_OK, path=None):
     """Given a command, mode, and a PATH string, return the path which
@@ -121,7 +123,8 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None):
                     return name
     return None
 
-
+# TODO: remove this function once we can depend on jupyter_packing, see:
+#  https://github.com/QuantStack/voila/pull/322
 # `run` function copied from jupyter_packaging under the following license:
 # -------------------------------------------------------------------------
 #
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ from io import BytesIO
 from subprocess import check_call, CalledProcessError
 
 import os
+import shlex
 import sys
 
 try:
@@ -25,11 +26,6 @@ log.set_verbosity(log.DEBUG)
 here = os.path.dirname(os.path.abspath(__file__))
 node_root = os.path.join(here, 'js')
 is_repo = os.path.exists(os.path.join(here, '.git'))
-
-npm_path = os.pathsep.join([
-    os.path.join(node_root, 'node_modules', '.bin'),
-    os.environ.get('PATH', os.defpath),
-])
 
 
 def in_read_the_docs():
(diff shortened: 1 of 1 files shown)


vidartf:
Works now. Note that the existing wrapping of `egg_info` causes the JS build to be run twice during install. I'd recommend using MANIFEST.in to pick up the datafiles instead, so you would only have to do a single build: https://setuptools.readthedocs.io/en/latest/setuptools.html#generating-source-distributions

maartenbreddels:
Sorry for the flake8, we should have something in the docs for pre-commit hooks.

maartenbreddels:
Whow, I find this a bit harder to digest than the solution in ipyvolume, what was wrong with that solution?

vidartf:
> Whow, I find this a bit harder to digest than the solution in ipyvolume, what was wrong with that solution?

Quoting from #192:

> The main two points are probably:
    - Using which to find the executable.
    - Using shell=True or not.

vidartf:
Regarding hard to digest, note that the `which` function is just a polyfill for Python 2.

maartenbreddels:
I still find https://github.com/maartenbreddels/ipyvolume/commit/a434ce71135e95844b1169655fe1d873a2e857a8 a better solution that is used throughout most of the widget packages. If that works, and is less lines of code, why should we the which/run solution instead.

These massive setup.py's scare people (including me), which is not good for contributions I think.

vidartf:
If you want to do it your way, don't ask me to open a PR ;)

maartenbreddels:
My point was, this little piece of code works for ipyvolume, since you are on windows, it's better if you test it is really working, sorry for being such as pain.

vidartf:
Ok, thanks for defusing the tension. Let me clarify my position a bit then as well:

In the long term, I think the solution for all of this should be good:
- Once we can reliably assume all users to be on `pip>=10`, we can have all python packages bundling JS have a [pyproject.toml](https://www.python.org/dev/peps/pep-0518/) where they require `jupyter_packaging`.
- `jupyter_packaging` is meant to avoid all JS building packages copy-pasting the same code in all its various, branching versions. This will put all fixes in one spot, and clean up setup.py code (just import needed functionality). It will also allow us to track what the reasoning is for each piece of code by more clearly following the commit log and issue discussions.

While we are waiting for that glorious future, I'm slowly trying to make as many packages as possible converge towards the code in `jupyter_packaging`. The main motivation being:
- Easier upgrade once we are able to switch.
- Trying to make sure everyone benefits from the fixes that are in `jupyter_packaging`.

Some of the solutions that are in `jupyter_packaging` are a little obscure on the surface. Some of them turn out to have a very good reason once you dig into it, some of them don't (but they rarely cause any harm). I simply went with the solution from jupyter packaging because I believe it has had more eyes on it, and been tested by more people. E.g. I believe the `shell=True` argument is critical in certain cases on Windows in order to get the correct env.

maartenbreddels:
Ok, that sounds good. Would you mind opening an issue or putting in some comments when we can remove this code, and maybe even refer to this PR in the code?

maartenbreddels:
@vidartf let me know if you agree with the comments I put in, if those are correct, then we can merge this.

vidartf:
LGTM, but flake is complaining ;)

PR #270, opened 2019-06-25
[WIP] Voila as an ExtensionApp

Zsailer:
I went ahead and started refactoring Voila to use the [`ExtensionApp`](https://github.com/jupyter/jupyter_server/pull/48) in jupyter_server. It greatly simplifies the main application. 

Using this API, Voila can be initialized, configured and launched from the command line, and you can generate configuration files using the `--generate-config` flag.

It can also be loaded as a Jupyter server extension, allowing you to serve Voila side-by-side with other client applications running on one server. 

One thing to note is that all static files are now served behind the url namespace `static/voila/` (this is automatically set up by `ExtensionApp`).

This should be working. Let me know if you have issues with testing it out.  

@SylvainCorlay @maartenbreddels

Issue #107, opened 2019-04-10
Voila-render source files


SylvainCorlay:
This is a usecase that is not handled by the voila, and was supported by @pbugnion's `ipywidgets_server`.

To handle that case, we need to figure out how to

 - decide which kernelspec to use for a given source file (should it be infered from the file extensions?).
 - should voila actually insert the content of the file into a single-cell notebook, or handle script file completely differently from notebooks?

choldgraf:
A potentially wacky solution would be to use jupytext (for python files) to convert into notebooks (and optionally populate cells in a modular fashion using jupytext syntax) before serving

maartenbreddels:
In the context of the material design widgets, I see also a use case where 'cell output' does not exist. Only widgets, with a particular 'id', on where that widget should end up in the template.

For the source file, I think the simplest approach would be to see it as a single cell and fully rely on display (or the last expression) for visualization. Using display also allows for markdown etc.

choldgraf:
Have y'all explored the idea of using cell metadata for this? That's what I use in jupyter book and it works pretty nicely (e.g.: https://jupyter.org/jupyter-book/features/hiding.html#hiding-code-cells-and-displaying-a-button-to-show-them)

maartenbreddels:
I think jupytext support would be nice, but I think we should also have pure Python support, or C++ even. We could rely on using `display(widget)`, and simply see the source file as 1 cell.

maartenbreddels:
I think if we want to support this, we may want to reconsider the extension. What if we have both `foobar.py` and `foobar.ipynb` in a directory, what should render at `/voila/render/foobar`?

maartenbreddels:
I think this is good material for the dashboarding workshops!

choldgraf:
amazing! very cool :-)

mwouts:
@choldgraf , it's good to see you here! I see that you contribute to every fun project :smile:. And thanks for mentioning Jupytext!

@SylvainCorlay, is it correct that you would like
- a `jupyter_server_config.d/jupytext.json` file that activates Jupytext's server extension on the `ServerApp`
- and that the server extension should set a contents manager that derives from the current app's content manager class (e.g. `LargeFileManager` in `jupyter_server.services.contents.largefilemanager` rather than from `LargeFileManager` in `notebook.services.contents.largefilemanager`) ?

SylvainCorlay:
> @SylvainCorlay, is it correct that you would like

That is correct!

[...]

PR #328, opened 2019-07-26
Feat support py files via jupytext

maartenbreddels:
This partly solves #107 because jupytext works with jupyter server out of the box, so we can do:
```
$ jupytext --to py notebooks/basics.ipynb
$ jupyter server --ServerApp.token='' --ServerApp.open_browser=False
$ open http://localhost:8889/voila/render/notebooks/basics.py
```
And it just works.
For voila standalone support we just have to wait for #270 and get this for free.

If we don't want to wait for this we could add cmd line option to voila to force it to use jupytext, e.g.:
```
$ voila --VoilaApp.force_jupytext
```
Another option would be to support py files natively out of the box (I'll do a separate PR for that).

### DIFF
--- a/voila/paths.py
+++ b/voila/paths.py
@@ -16,7 +16,7 @@ STATIC_ROOT = os.path.join(ROOT, 'static')
 # if the directory above us contains the following paths, it means we are installed in dev mode (pip install -e .)
 DEV_MODE = os.path.exists(os.path.join(ROOT, '../setup.py')) and os.path.exists(os.path.join(ROOT, '../share'))
 
-notebook_path_regex = r'(.*\.ipynb)'
+notebook_path_regex = r'(.*\.(?:ipynb|py))'
 
 
 def collect_template_paths(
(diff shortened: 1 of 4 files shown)


voila/paths.py
```diff
@@ -16,7 +16,7 @@
 # if the directory above us contains the following paths, it means we are installed in dev mode (pip install -e .)
 DEV_MODE = os.path.exists(os.path.join(ROOT, '../setup.py')) and os.path.exists(os.path.join(ROOT, '../share'))
 
-notebook_path_regex = r'(.*\.ipynb)'
+notebook_path_regex = r'(.*\.(?:ipynb|py))'
```
jtpio: Wondering if we could get support for other languages for free here (Julia, R and so on)?

maartenbreddels: Indeed, I've added julia in #330, because I could test it. I didn't feel comfortable adding more. I also wonder if we should have a `.*` regex?

jtpio: This regex was added because voila could only handle `.ipynb` files.

But if scripts (in different languages) are going to be supported, it probably make sense to switch back to `.*`. Otherwise we would have to maintain a white list of supported extensions.

maartenbreddels:
My preference now is to rely on #330 (merge that PR, close this PR), and document that you can use jupytext instead.

PR #330, opened 2019-07-29
Feat: support script files

maartenbreddels:
Depends on #309  (alternative to #328)

Now we can do:
```
$ voila tests/notebooks/print.jl
$ voila tests/notebooks/print.py
```

It works by injecting the script code into an in-memory single-cell notebook, which means the last expression is printed out. If for some reasons one needs more output, you can use `display(expression)`, or use jupytext (which should also be supported with this PR).

### DIFF
--- a/tests/app/tree_test.py
+++ b/tests/app/tree_test.py
@@ -19,4 +19,4 @@ def test_tree(http_client, base_url):
     text = response.body.decode('utf-8')
     assert 'print.ipynb' in text, 'tree handler should render ipynb files'
     assert 'print.xcpp' in text, 'tree handler should render xcpp files (due to extension_language_mapping)'
-    assert 'print.py' not in text, 'treeh handler should not render .py files (due to extension_language_mapping)'
+    assert 'print.py' not in text, 'tree handler should not render .py files (due to extension_language_mapping)'
--- a/tests/app/tree_test.py
+++ b/tests/app/tree_test.py
@@ -1,0 +1,22 @@
+# test tree rendering
+import pytest
+
+
+@pytest.fixture
+def voila_args(notebook_directory, voila_args_extra):
+    return ['--VoilaTest.root_dir=%r' % notebook_directory, '--VoilaTest.log_level=DEBUG'] + voila_args_extra
+
+
+@pytest.fixture
+def voila_args_extra():
+    return ['--VoilaConfiguration.extension_language_mapping={".xcpp": "C++11"}']
+
+
+@pytest.mark.gen_test
+def test_tree(http_client, base_url):
+    response = yield http_client.fetch(base_url)
+    assert response.code == 200
+    text = response.body.decode('utf-8')
+    assert 'print.ipynb' in text, 'tree handler should render ipynb files'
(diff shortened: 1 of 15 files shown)


maartenbreddels:
Example usage:
```
voila --VoilaConfiguration.extension_language_mapping='{".py": "python", ".xcpp": "C++11"}' \
  --VoilaConfiguration.language_kernel_mapping='{"python": "xpython"}' tests/notebooks
```
Now you can access:
 * http://localhost:8866/voila/render/print.py (using xeus-python kernel)
 * http://localhost:8866/voila/render/print.xcpp (uses xeus-cling's C++11 kernel)

The tree handler and the voila handler (both the standalone and server extension) will serve and show `.ipynb` and all the extensions listed in this configuration option.

The final step would be to document this

We might also want an `extension_kernel_mapping` configuration, to go directly from `.py` -> `<kernel_name>`, but maybe that is overkill.

tests/app/tree_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@pytest.mark.gen_test
+def test_tree(http_client, base_url):
+    response = yield http_client.fetch(base_url)
+    assert response.code == 200
+    text = response.body.decode('utf-8')
+    assert 'print.ipynb' in text, 'tree handler should render ipynb files'
+    assert 'print.xcpp' in text, 'tree handler should render xcpp files (due to extension_language_mapping)'
+    assert 'print.py' not in text, 'treeh handler should not render .py files (due to extension_language_mapping)'
```
jtpio:
```suggestion
    assert 'print.py' not in text, 'tree handler should not render .py files (due to extension_language_mapping)'
```

jtpio:
> We might also want an `extension_kernel_mapping` configuration, to go directly from `.py` -> `<kernel_name>`, but maybe that is overkill.

That would remove one level of indirection, but I guess having both `extension_language_mapping` and `language_kernel_mapping` gives enough flexibility. Also we would need to handle conflicts or precedence if the 3 options are used together (in a conf file for example).

jtpio:
We can also consider adding aliases for these options if they tend to be used often.

voila/handler.py
```diff
@@ -140,17 +145,36 @@ def fix_notebook(self, notebook):
...
+        spec = all_kernel_specs[kernel_name]
+        notebook = nbformat.v4.new_notebook(
+            metadata={
+                'kernelspec': {
+                    'display_name': spec['spec']['display_name'],
+                    'language': spec['spec']['language'],
+                    'name': kernel_name
+                }
+            },
+            cells=[nbformat.v4.new_code_cell(model['content'])],
```
jtpio: Maybe later after adding support for the progressive rendering (https://github.com/QuantStack/voila/pull/133) there could be more than one cell created, from cell delimiters such as `%%` or `# In[1]:`.

maartenbreddels: Yes, this could again be a configurable, mapping from file extension to regex. I'd say not now right?

jtpio: Indeed let's do that later (we can open an issue to track it).

PR #358, opened 2019-08-21
fix: support nested output widgets in voila executor

### LINKED ISSUE #355: control output widget from interact
Minimal example to cause error:
```
#cell 1
import ipywidgets as widgets
from ipywidgets import interact
from IPython.display import clear_output
output1 = widgets.Output()
output1

#cell 2
def toggle_text(val):
    with output1:
        print(val)
interact(toggle_text, val=widgets.ToggleButton(description='Toggle text'))
```

```
...
File "/opt/anaconda3/lib/python3.7/site-packages/voila/execute.py", line 79, in set_state
    del self.executor.output_hook[self.msg_id]
KeyError: 'e7108564-b1a38beaf451e1c7095357b2'
```

I can fix this problem by changing https://github.com/jkerfs/voila/blob/master/voila/execute.py#L79
to 
```
if self.msg_id in self.executor.output_hook:
    del self.executor.output_hook[self.msg_id]
```

Is anyone familiar with why this would be happening?  Am I using interact and output widget wrong or should we include that output hook check shown above to resolve it?

maartenbreddels:
Fixes #355

The issue boiled down to having nest output widgets like:
```python
import ipywidgets as widgets
output1 = widgets.Output()
output2 = widgets.Output()
with output1:
    print('in 1')
    with output2:
        print('in 2')
    print('also in 1')
```

Voila would not keep a stack of output hooks, but only the last one. This caused it to fail after the output1 context was left.

The solution is to keep a stack of output hooks, tests are included.

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -5,6 +5,7 @@
 #                                                                           #
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
+import collections
 
 from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError, ExecutePreprocessor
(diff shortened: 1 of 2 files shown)


voila/execute.py
```diff
@@ -115,13 +138,15 @@ def handle_comm_msg(self, outs, msg, cell_index):
             if 'state' in data:
                 state = data['state']
                 comm_id = msg['content']['comm_id']
+                print('set_state', comm_id, state)
```
jtpio: Looks like this print can be removed.

jtpio:
Thanks @maartenbreddels.

Tested locally with the example above and the nested outputs are handled properly.

PR #361, opened 2019-08-21
feat: serve static files similar to the notebook

maartenbreddels:
but limited by whitelist and blacklist rules.

# Issue

When using the following in a Jupyter notebook:
```python
import ipywidgets
html = ipywidgets.HTML(value='<img src="./quantstack.svg"></img>')
html
```
The Jupyter notebook server will happily serve the svg file.

However, in voila, we consider this a security risk, since we do not assume the users are trusted (we don't want to give users access to our files). However, in voila, we'd like this to work as well.

# Solution 
This PR solves this by having whitelist and blacklist regexes that protect the content that is served. Each request for a file needs to match *at least one whitelist regex and NO blacklist regex* 

E.g. to serve all files, except those starting with private, or notebook files:
```
voila mynotebook.ipynb --VoilaConfiguration.file_whitelist="['.*']" --VoilaConfiguration.file_blacklist="['private.*', '.*\.(ipynb)']"
```

To serve only images:
```
voila tests/ --VoilaConfiguration.file_whitelist=ile_whitelist="['.*\.(png|jpg|gif|svg)']"
```

# Uncertain
I think that we should not always redirect in the handler.py, since it tells users that a file exists (this gives an attacker information which files exists). I think we should check the rules in two places.

Should the default be as it is now (images, and notebooks/py files always blacklisted)?

TODO:
 * [ ] tests
 * [ ] docs

### DIFF
--- a/voila/static_file_handler.py
+++ b/voila/static_file_handler.py
@@ -7,6 +7,7 @@
 #############################################################################
 
 import os
+import re
 
 import tornado.web
 
@@ -37,3 +38,21 @@ class MultiStaticFileHandler(tornado.web.StaticFileHandler):
                 self.root = root  # make sure all the other methods in the base class know how to find the file
                 break
         return abspath
+
+
+class WhiteListFileHandler(tornado.web.StaticFileHandler):
+    def initialize(self, whitelist=[], blacklist=[], **kwargs):
+        self.whitelist = whitelist
+        self.blacklist = blacklist
+        super(WhiteListFileHandler, self).initialize(**kwargs)
+
+    def get_absolute_path(self, root, path):
+        # StaticFileHandler.get always calls this method first, so we use this as the
+        # place to check the path. Note that now the path seperator is os dependent (\\ on windows)
+        whitelisted = any(re.fullmatch(pattern, path) for pattern in self.whitelist)
+        blacklisted = any(re.fullmatch(pattern, path) for pattern in self.blacklist)
+        if not whitelisted:
+            raise tornado.web.HTTPError(403, 'File not whitelisted')
+        if blacklisted:
+            raise tornado.web.HTTPError(403, 'File blacklisted')
+        return super(WhiteListFileHandler, self).get_absolute_path(root, path)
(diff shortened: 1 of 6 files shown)


SylvainCorlay:
Thanks for this @maartenbreddels this is not what I had in mind with the inlining but I think that it is a good solution at least for now. Happy to merge this.

voila/static_file_handler.py
```diff
@@ -37,3 +38,21 @@ def get_absolute_path(self, root, path):
...
+
+class WhiteListFileHandler(tornado.web.StaticFileHandler):
+    def initialize(self, whitelist=[], blacklist=[], **kwargs):
+        self.whitelist = whitelist
+        self.blacklist = blacklist
+        super(WhiteListFileHandler, self).initialize(**kwargs)
+
+    def get_absolute_path(self, root, path):
+        # StaticFileHandler.get always calls this method first, so we use this as the
+        # place to check the path. Note that now the path seperator is os dependent (\\ on windows)
```
jasongrout: separator

voila/static_file_handler.py
```diff
@@ -37,3 +38,21 @@ def get_absolute_path(self, root, path):
...
+        whitelisted = any(re.fullmatch(pattern, path) for pattern in self.whitelist)
+        blacklisted = any(re.fullmatch(pattern, path) for pattern in self.blacklist)
+        if not whitelisted:
+            raise tornado.web.HTTPError(403, 'File not whitelisted')
```
jasongrout: I think these probably should be 404s to be indistinguishable from the file not existing. If you decide on 403s, though, perhaps a more generic message is better, like the default 403 message, or just now message.

jasongrout:
> I think that we should not always redirect in the handler.py, since it tells users that a file exists (this gives an attacker information which files exists). I think we should check the rules in two places.

I think it is fine to unconditionally redirect, since that gives no more information.

> Should the default be as it is now (images, and notebooks/py files always blacklisted)?

I think the default should be secure, i.e., no files served.

PR #364, opened 2019-08-26
Run executenb on a ThreadPoolExecutor

### LINKED ISSUE #363: Performance: nbconvert blocks tornado ioloop
During a review of #133 with @jtpio we observed that the progressive loading (i.e. the spinner) didn't work for firefox, or in chrome with the dev console open.

The issue was that tornado enters the voila handler, writes the chunks that are flushed out, but never returns control to the tornado IO loop.

Firefox (or chrome in dev mode with cache disabled), got some chunks of HTML, and tried to fetch js and css resources. But since the IO loop is blocked, the request only gets handled after all the chunks are pushed to the client/browser. The result is that the browser can only show the content after the js before the body tag is downloaded, making the progressive rendering fail. Chrome happens to cache the files, so there it worked.

This was solved by 33db3546a0691feb765686aa669137a6deddb6af which yields after writing and flushing each chunk. 

However, we realised another issue. Although we give control back to tornado's IO loop, nbconvert does not, it does a busy wait/polling on each cell execution. This means that if a cell takes 10 minutes to execute, the whole tornado server will be unresponsive.

There are two solutions possible:
 * Have the nbconvert cell execution in a thread (no worries about the GIL!).
 * Make nbconvert work async.

The first one is probably the easiest to implement.

PS: Note that this issue is not related to #133, it only revealed it.

jtpio:
Fixes #363 

This is using `run_on_executor` as suggested in https://github.com/QuantStack/voila/issues/363#issuecomment-524354712, and a [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html).

async-exec-voila

cc @maartenbreddels

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -7,8 +7,10 @@
 #############################################################################
 
 import os
+from concurrent.futures import ThreadPoolExecutor
 
 import tornado.web
+from tornado.concurrent import run_on_executor
 
 from jupyter_server.base.handlers import JupyterHandler
 from jupyter_server.config_manager import recursive_update
@@ -20,6 +22,8 @@ from .exporter import VoilaExporter
 
 
 class VoilaHandler(JupyterHandler):
+
+    executor = ThreadPoolExecutor()
 
     def initialize(self, **kwargs):
         self.notebook_path = kwargs.pop('notebook_path', [])    # should it be []
@@ -58,7 +62,7 @@ class VoilaHandler(JupyterHandler):
         cwd = os.path.dirname(notebook_path)
         kernel_id = yield tornado.gen.maybe_future(self.kernel_manager.start_kernel(kernel_name=notebook.metadata.kernelspec.name, path=cwd))
         km = self.kernel_manager.get_kernel(kernel_id)
-        result = executenb(notebook, km=km, cwd=cwd, config=self.traitlet_config)
+        result = yield self.execute_notebook(notebook, km=km, cwd=cwd, config=self.traitlet_config)
 
         # render notebook to html
         resources = {
@@ -101,6 +105,11 @@ class VoilaHandler(JupyterHandler):
 
     def redirect_to_file(self, path):
         self.redirect(url_path_join(self.base_url, 'voila', 'files', path))
+
+    @run_on_executor
+    def execute_notebook(self, notebook, km, cwd, config):
+        """Execute a notebook asynchronously on the executor"""
+        return executenb(notebook, km=km, cwd=cwd, config=config)
 
     @tornado.gen.coroutine
     def load_notebook(self, path):


jtpio:
Also since it is using `concurrent.futures`, this is not compatible with Python 2.

voila/handler.py
```diff
@@ -21,6 +23,8 @@
 
 class VoilaHandler(JupyterHandler):
 
+    executor = ThreadPoolExecutor()
```
jtpio: From 3.5+, if not specified, `max_workers` defaults to "to the number of processors on the machine, multiplied by 5".

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

maartenbreddels:
I'm afraid it is not that simple. This will fix it for the current situation, but the progressive part will not map to this solution.
I'm not sure a ThreadPoolExecutor is the best solution, since we need to await each html snippet, and all calls to the kernel should come from the same thread. 

Out of interest, could you benchmark this, say run:
```
$  time ab -n 10 -c 10 http://localhost:8866/ # parallel
$  time ab -n 10 -c 1 http://localhost:8866/ # serial
```

jtpio:
Yes for the case of the progressive rendering it looks slightly more complicated I agree. Which is why this PR is only focused on the current situation.

jtpio:
Here are the results of the tests:

```
$  time ab -n 10 -c 10 http://localhost:8866/ # parallel

This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient).....done


Server Software:        TornadoServer/6.0.3
Server Hostname:        localhost
Server Port:            8866

Document Path:          /
Document Length:        23216 bytes

Concurrency Level:      10
Time taken for tests:   4.292 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      234850 bytes
HTML transferred:       232160 bytes
Requests per second:    2.33 [#/sec] (mean)
Time per request:       4292.363 [ms] (mean)
Time per request:       429.236 [ms] (mean, across all concurrent requests)
Transfer rate:          53.43 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       0
Processing:  1507 2455 435.0   2695    2785
Waiting:     1507 2455 435.1   2695    2785
Total:       1508 2455 435.0   2695    2785

Percentage of the requests served within a certain time (ms)
  50%   2695
  66%   2695
  75%   2695
  80%   2695
  90%   2785
  95%   2785
  98%   2785
  99%   2785
 100%   2785 (longest request)
ab -n 10 -c 10 http://localhost:8866/  0,01s user 0,01s system 0% cpu 4,308 total
```

```
$ time ab -n 10 -c 1 http://localhost:8866/ # serial

This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient).....done


Server Software:        TornadoServer/6.0.3
Server Hostname:        localhost
Server Port:            8866

Document Path:          /
Document Length:        23216 bytes

Concurrency Level:      1
Time taken for tests:   12.574 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      234850 bytes
HTML transferred:       232160 bytes
Requests per second:    0.80 [#/sec] (mean)
Time per request:       1257.365 [ms] (mean)
Time per request:       1257.365 [ms] (mean, across all concurrent requests)
Transfer rate:          18.24 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  1195 1257  50.1   1247    1336
Waiting:     1195 1257  50.1   1247    1336
Total:       1195 1257  50.1   1247    1336

Percentage of the requests served within a certain time (ms)
  50%   1247
  66%   1287
  75%   1289
  80%   1326
  90%   1336
  95%   1336
  98%   1336
  99%   1336
 100%   1336 (longest request)
ab -n 10 -c 1 http://localhost:8866/  0,00s user 0,00s system 0% cpu 12,580 total
```

jtpio:
Also since it was running on Python 3.7, the `ThreadPoolExecutor` default its `max_workers` to 40 (8 cpus * 5).

```python
import multiprocessing
multiprocessing.cpu_count()
```

jtpio:
Bumping the number of concurrent connections to `20` with:

```
ab -n 100 -c 20 http://localhost:8866/
```

And starting to get a few issues with the number of open file descriptors as well as:

```
File "zmq/backend/cython/socket.pyx", line 547, in zmq.backend.cython.socket.Socket.bind
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: Address already in use

```

It would be interesting to revive the discussion on having a pool of kernels / forking kernels: https://github.com/QuantStack/voila/issues/30

maartenbreddels:
Indeed, I see this as well (the file descriptors), it's useful to rub these benchmarks.

What I am really surprised about is the horrible performance, why only ~2 requests/second?

jtpio:
Also the kernels are not automatically shut down when running these benchmarks. So memory consumption goes up quickly as well.

jtpio:
> Indeed, I see this as well (the file descriptors), it's useful to rub these benchmarks.

We will probably want to have a sane default for `max_workers` then to limit the concurrency.

timkpaine:
Remember threadpool is still GIL limited. Any reason to not use a process pool?

maartenbreddels:
The kernels are in a separate process, the GIL should not be a problem. We only use the threads to escape that nbconvert is blocking, and not working with asyncio.
It may still be interesting to benchmark it though, because I believe (still interpreting the profiler results) I still see a lot of GIL.

jtpio:
> What I am really surprised about is the horrible performance, why only ~2 requests/second?

The previous benchmarks were run with the `basics.ipynb` notebook.

Using a notebook that only has one markdown cell:

```
$ time ab -n 10 -c 10 http://localhost:8866/

This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient).....done


Server Software:        TornadoServer/6.0.3
Server Hostname:        localhost
Server Port:            8866

Document Path:          /
Document Length:        6369 bytes

Concurrency Level:      10
Time taken for tests:   2.198 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      66370 bytes
HTML transferred:       63690 bytes
Requests per second:    4.55 [#/sec] (mean)
Time per request:       2197.551 [ms] (mean)
Time per request:       219.755 [ms] (mean, across all concurrent requests)
Transfer rate:          29.49 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   713 1204 267.4   1417    1484
Waiting:      713 1204 267.5   1417    1484
Total:        713 1204 267.4   1417    1484

Percentage of the requests served within a certain time (ms)
  50%   1417
  66%   1417
  75%   1417
  80%   1418
  90%   1484
  95%   1484
  98%   1484
  99%   1484
 100%   1484 (longest request)
ab -n 10 -c 10 http://localhost:8866/  0,01s user 0,01s system 0% cpu 2,214 total

```

timkpaine:
Ah yeah I see didn't look at the code, processes won't get you much. You can probably defer more blocking work to the pool though, are some of those kernel management lines above blocking? Also there are no python2 issues yet, concurrent futures works fine.

jtpio:
>  Also there are no python2 issues yet, concurrent futures works fine.

Still needs the backport though? https://pypi.org/project/futures/

timkpaine:
@jtpio I've used the backport extensively in production, it works fine

timkpaine:
Threadpool works best on blocking but Gil-less calls. Those executes aren't Gil less so you shouldn't get too much benefit

maartenbreddels:
> Using a notebook that only has one markdown cell:

I've added 20 cells with `pass` in the notebook (on my local version, not committed), which gives me horrible performance, do you see the same? Could you run `ab` on that?

jtpio:
@maartenbreddels with the "20 pass":

```
$ time ab -n 10 -c 10 http://localhost:8866/
This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient).....done


Server Software:        TornadoServer/6.0.3
Server Hostname:        localhost
Server Port:            8866

Document Path:          /
Document Length:        6116 bytes

Concurrency Level:      10
Time taken for tests:   2.761 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      63840 bytes
HTML transferred:       61160 bytes
Requests per second:    3.62 [#/sec] (mean)
Time per request:       2761.052 [ms] (mean)
Time per request:       276.105 [ms] (mean, across all concurrent requests)
Transfer rate:          22.58 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       0
Processing:   847 1530 369.8   1770    1914
Waiting:      847 1530 369.9   1770    1914
Total:        847 1530 369.8   1770    1914

Percentage of the requests served within a certain time (ms)
  50%   1770
  66%   1773
  75%   1773
  80%   1911
  90%   1914
  95%   1914
  98%   1914
  99%   1914
 100%   1914 (longest request)
ab -n 10 -c 10 http://localhost:8866/  0,02s user 0,00s system 0% cpu 2,776 total

```

SylvainCorlay:
Closing as per in-person discussions on how we are going to tackle this issue.

vidartf:
> Closing as per in-person discussions on how we are going to tackle this issue.

How? 😺

SylvainCorlay:
> How? :smiley_cat:

So basically, we will not be using threads but integrate into the tornado even loop. The blocking nbexecute call will be replaced by an async function that we not block the event loop while code is running.

maartenbreddels:
The progressive rendering will not work with this solution. We plan to use asyncio, get rid of py27 support, move to minimally python35. Feel free to weigh in.

vidartf:
> We plan to use asyncio, get rid of py27 support, move to minimally python35.

❤️

mpacer:
I realize this is closed, but is there an issue tracking the progress on the proposed plan to drop py27 support and move to minimal python 3.5 & switching nbexecute to an async generator?

SylvainCorlay:
@mpacer unfortunately, async generators were only added with Python 3.7, and we can't require Python 3.7 yet :disappointed:

The ultimate goal is to make jupyter_client fully async for both the management of the kernel lifecycle and for execution requests, etc, and to use an async generator for executing the cells.

Since making jupyter_client fully async will take a bit of time, @maartenbreddels  worked around the issue in https://github.com/voila-dashboards/voila/pull/374 by using a thread executor to make the execution of cells non-blocking.

jf---:
> Also the kernels are not automatically shut down when running these benchmarks. So memory consumption goes up quickly as well

I think this is still an achilles heel of the project - by reloading the page rapidly many kernels are spun up, which take time to get reaped. Watching the memory consumption on a server while continuously refreshing the page looks uncomfortable like a a DDoS attack by hitting reload 20x.
Perhaps a straightfwd way of sidestapping this issue is throttling reload. I recall @maartenbreddels made cool `debounce` decorator. That approach to avoid spinning up too many kernels all too easily might be of interest.

PR #368, opened 2019-08-28
Docs update for v0.1.10

maartenbreddels:


RELEASE.md
```diff
@@ -22,7 +22,10 @@ Make sure the `dist/` folder is empty.
 1. Update [voila/_version.py](./voila/_version.py) and [environment.yml](./environment.yml) with the new version number (see and [example diff](https://github.com/QuantStack/voila/commit/5c6fd8dd3ea71412ae9c20c25248453d22a3b60a))
 2. `python setup.py sdist bdist_wheel`
 3. Double check the size of the bundles in the `dist/` folder
-4. `twine upload dist/*`
+4. Run the tests
```
jtpio: As an alternative, making sure the last commit on master passes on Travis?

maartenbreddels: But that can miss some things that go right on travis, like conda dependencies etc (like we had with pygments)

docs/source/customize.rst
```diff
@@ -183,3 +183,36 @@ Lab session, use the following command when starting the server::
 
 When users run ``voila`` by hitting the ``voila/`` endpoint, this configuration
 will be used.
+
+Serving static files
+====================
+
+Unlike the Jupyter lab or classical notebook server, ``voila`` does not serve
```
jtpio: I think this is the "official" spelling?

```suggestion
Unlike JupyterLab or the classic notebook server, ``voila`` does not serve
```

RELEASE.md
jtpio: Right, but running locally could also have some side effects (leaking environments).

maartenbreddels: Yes, both should pass, and I use a new conda env just to be sure.

jtpio:
Thanks!

PR #372, opened 2019-08-30
Load template-related config from JSON file.

mkcor:
Check this out, @maartenbreddels!

It's basically a cleaner version of #326 (which I will close). Still missing a test or two though...

I have used `os.path.dirname` to go one directory up.

Here, we are not simply reading whatever `conf.json` file we may find in a list of template paths, we are ensuring it corresponds to the template being used.

Demo style:

* if no `conf.json` file lives in `PREFIX/share/jupyter/voila/templates/reveal`, then running
```
$ voila reveal.ipynb --template=reveal
```
will launch a slideshow where transitions slide (because `'slide'` is the default, as set in the reveal template);

* if this `conf.json` [file](https://github.com/QuantStack/voila/pull/326#issuecomment-526417088) lives in `PREFIX/share/jupyter/voila/templates/reveal`, then running
```
$ voila reveal.ipynb --template=reveal
```
will launch a slideshow where transitions zoom, because this config file is well-formed and says `"transition": "zoom"`;

* if this `conf.json` [file](https://github.com/QuantStack/voila/pull/326#issuecomment-526417088) lives in `PREFIX/share/jupyter/voila/templates/reveal`, then running
```
$ voila reveal.ipynb --template=reveal --VoilaConfiguration.resources="{'reveal': {'transition': 'fade'}}"
```
will launch a slideshow where transitions fade, because CLI config has preference; if `conf.json` specified a non-default resource (e.g., `"scroll": true`), it wouldn't be taken into account, unfortunately, because traitlets' Config [merge](https://github.com/ipython/traitlets/blob/41551bc8b30ccc28af738e93615e3408cb94d5d3/traitlets/config/loader.py#L190) is not recursive...

If we really wanted to merge all template-related configs recursively, always ensuring CLI priority, we would need to have the convoluted approach I had in #326.

The little contortions I had to do in `handler.py` come from covering all cases, e.g., for
```
$ voila reveal.ipynb --template=reveal --VoilaConfiguration.resources="{'reveal': {'transition': 'fade'}}"
```
to run regardless, with or without a config file somewhere. And all the config merging and passing around happening in `app.py` leads to the fact that: a) we need to get to `self.voila_configuration.config.VoilaConfiguration.resources` to find those extra resources (so painfully retrieved and combined) and b) `self.voila_configuration.config.VoilaConfiguration.resources` ends up being of type `traitlets.config.loader.LazyConfigValue` when neither a file nor CLI config passes extra resources (maybe I could prevent this by setting `self.voila_configuration.config.VoilaConfiguration.resources = {}` ahead of time in `app.py`).

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -40,10 +40,9 @@ from traitlets import Unicode, Integer, Bool, Dict, List, default
 from jupyter_server.services.kernels.kernelmanager import MappingKernelManager
 from jupyter_server.services.kernels.handlers import KernelHandler, ZMQChannelsHandler
 from jupyter_server.services.contents.largefilemanager import LargeFileManager
-from jupyter_server.base.handlers import path_regex
+from jupyter_server.base.handlers import FileFindHandler, path_regex
 from jupyter_server.utils import url_path_join
 from jupyter_server.services.config import ConfigManager
-from jupyter_server.base.handlers import FileFindHandler
 
 from jupyter_client.kernelspec import KernelSpecManager
 
@@ -341,8 +340,6 @@ class Voila(Application):
 
         # then we load the config
         self.load_config_file('voila', path=self.config_file_paths)
-        # but that cli config has preference, so we overwrite with that
-        self.update_config(self.cli_config)
         # common configuration options between the server extension and the application
         self.voila_configuration = VoilaConfiguration(parent=self)
         self.setup_template_dirs()
@@ -480,13 +477,15 @@ class Voila(Application):
             self.log.debug('serving directory: %r', self.root_dir)
             handlers.extend([
                 (self.server_url, VoilaTreeHandler, tree_handler_conf),
-                (url_path_join(self.server_url, r'/voila/tree' + path_regex), VoilaTreeHandler, tree_handler_conf),
-                (url_path_join(self.server_url, r'/voila/render/(.*)'), VoilaHandler,
-                    {
-                        'nbconvert_template_paths': self.nbconvert_template_paths,
-                        'config': self.config,
-                        'voila_configuration': self.voila_configuration
-                    }),
+                (url_path_join(self.server_url, r'/voila/tree' + path_regex),
+                 VoilaTreeHandler, tree_handler_conf),
+                (url_path_join(self.server_url, r'/voila/render/(.*)'),
+                 VoilaHandler,
+                 {
+                     'nbconvert_template_paths': self.nbconvert_template_paths,
+                     'config': self.config,
+                     'voila_configuration': self.voila_configuration
+                 }),
             ])
 
         self.app.add_handlers('.*$', handlers)
--- a/voila/app.py
+++ b/voila/app.py
@@ -357,21 +357,19 @@ class Voila(Application):
             template_conf_dir = [os.path.join(k, '..') for k in self.nbconvert_template_paths]
             conf_paths = [os.path.join(d, 'conf.json') for d in template_conf_dir]
             for p in conf_paths:
-                # see if config path corresponds to template in use
-                if os.path.basename(os.path.dirname(p)) == self.voila_configuration.template:
-                    # see if config file exists
-                    if os.path.exists(p):
-                        # load the template-related config
-                        loader = JSONFileConfigLoader('conf.json', template_conf_dir)
-                        conf = loader.load_config()
-                        # check that config file is meant for template in use
-                        assert conf.traitlet_configuration['base_template'] == self.voila_configuration.template
-                        # create sub-config as Config instance for merging (see next)
-                        conf.VoilaConfiguration = Config(conf.traitlet_configuration)
-                        # merge config from file with otherwise voila config, ensuring CLI config priority
-                        conf.merge(self.voila_configuration.config)
-                        # pass merged config to overall voila config
-                        self.voila_configuration.config.VoilaConfiguration = conf.VoilaConfiguration
+                # see if config file exists
+                if os.path.exists(p):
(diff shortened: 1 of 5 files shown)


mkcor:
@SylvainCorlay (following up on your earlier question, if I understand it better) with this hack, do you expect more than one way to pass resources at the CLI, for example something like
```bash
--VoilaConfiguration.resources="{'config': {'traitlet_configuration': {'resources': {'reveal': {'transition': 'fade'}}}}}"
```
in addition to 
```bash
--VoilaConfiguration.resources="{'reveal': {'transition': 'fade'}}"
```
?

It's not trivial to find the alternative structure which would work, due to the convoluted config merge I'm doing here: https://github.com/QuantStack/voila/pull/372/files#diff-024301d95616622c19f49b8cd78d33c3R363-R367

mkcor:
@maartenbreddels could you please review, so I can wrap up this week? Thanks!

voila/app.py
```diff
@@ -348,6 +346,25 @@ def setup_template_dirs(self):
                 self.static_paths,
                 self.template_paths,
                 self.voila_configuration.template)
+            # look for possible template-related config files
+            template_conf_dir = [os.path.dirname(k) for k in self.nbconvert_template_paths]
```
maartenbreddels: This will not give the proper path if k ends with a slash

voila/app.py
```diff
@@ -348,6 +346,25 @@ def setup_template_dirs(self):
...
+            conf_paths = [os.path.join(d, 'conf.json') for d in template_conf_dir]
+            for p in conf_paths:
+                # see if config path corresponds to template in use
+                if os.path.basename(os.path.dirname(p)) == self.voila_configuration.template:
```
maartenbreddels: I don't think we should check this, actually, we'd like to 'inherit' all the configurations of the base template.

voila/app.py
```diff
@@ -348,6 +346,25 @@ def setup_template_dirs(self):
...
+                        # load the template-related config
+                        loader = JSONFileConfigLoader('conf.json', template_conf_dir)
+                        conf = loader.load_config()
+                        # check that config file is meant for template in use
+                        assert conf.traitlet_configuration['base_template'] == self.voila_configuration.template
+                        # create sub-config as Config instance for merging (see next)
+                        conf.VoilaConfiguration = Config(conf.traitlet_configuration)
+                        # merge config from file with otherwise voila config, ensuring CLI config priority
+                        conf.merge(self.voila_configuration.config)
+                        # pass merged config to overall voila config
```
maartenbreddels: I think instead of relying on the JSONFileConfigLoader, it would be easier to load the json using the json library, use the merge functionality of jupyter_server, and pass the resulting merged dict to the Config constructor. What do you think?

voila/app.py
maartenbreddels:
```suggestion
            template_conf_dir = [os.path.join(k, '..') for k in self.nbconvert_template_paths]
```

voila/app.py
mkcor: Right, it would be more straightforward, more readable to use plain JSON... And then we wouldn't have to sacrifice this precious recursive merge. You have convinced me!

voila/app.py
mkcor: Right, thank you! I definitely have to remove the check line 354 then.

voila/app.py
mkcor: Agreed. Fixed in fe5dd7f633dc1202b00add745d28df6a662f9781.

mkcor: Agreed. Fixed in fe5dd7f633dc1202b00add745d28df6a662f9781.

voila/app.py
mkcor: Addressed in 52f49db618347494846b21e16047d30aa935f36b (reminiscent #326 and sparing a couple lines of code).

SylvainCorlay:
@mkcor the test failure is not related to your PR and was in master. It should be fixed it you rebase on the current master.

SylvainCorlay:
This looks good to me.

voila/handler.py
```diff
@@ -69,7 +69,9 @@ def get(self, path=None):
         }
 
         # include potential extra resources
-        extra_resources = self.voila_configuration.resources
+        extra_resources = self.voila_configuration.config.VoilaConfiguration.resources
+        if not isinstance(extra_resources, dict):
+            extra_resources = extra_resources.to_dict()
```
SylvainCorlay: I don't understand this change.

mkcor: Wait, I should be more specific in my answer: when `PREFIX/share/jupyter/voila/templates/reveal/conf.json` does not exist *and no resource gets passed at the CLI*, `extra_resources` is a `traitlets.config.loader.LazyConfigValue` object.

Now, you may ask, but it's only a `traitlets.config.loader.LazyConfigValue` object because of the change you made right above, line 72! And that's correct, because `self.voila_configuration.resources` is always a dict. The reason I'm going for `self.voila_configuration.config.VoilaConfiguration.resources` is to support the case *with* config file, because `self.voila_configuration.resources` contains only resources passed at the CLI, not those in the config file. This in turns stems from the very specific config merge we are doing here: https://github.com/voila-dashboards/voila/pull/372/files#diff-024301d95616622c19f49b8cd78d33c3R372

mkcor:
> @mkcor the test failure is not related to your PR and was in master. It should be fixed it you rebase on the current master.

Thank you! Done.

SylvainCorlay:
Thanks @mkcor !

PR #389, opened 2019-09-27
Fix exporter for templates not using the progressive rendering

jtpio:
Fixes the following issue reported in the `gridstack` repo: https://github.com/voila-dashboards/voila-gridstack/issues/12

This should also apply to all the templates that are not using the progressive rendering and derive from `base.tpl`.

cc @maartenbreddels


jtpio:
Ideally we should have tests for that too. Or make the current `test_template` execute cells.

PR #391, opened 2019-10-02
Hide empty cells using CSS

### LINKED ISSUE #387: Extra padding in empty cells
Cells that contain code but no output, still take up some space in the output. Looks like there might be a 5px padding?

Example Notebook:
image

Example Output:
image

Extra Padding:
image

Expected behaviour is that the two lines should not have a gap between them.

martinRenou:
Should fix #387

### DIFF
--- a/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
@@ -34,7 +34,14 @@
 {%- endblock html_head_js -%}
 
 {%- block html_head_css -%}
+  <style>
+    /*Hide empty cells*/
+    .jp-mod-noOutputs.jp-mod-noInput {
+      display: none;
+    }
+  </style>
 {%- endblock html_head_css -%}
+
 {%- endblock html_head -%}
 </head>
 {%- endblock header -%}
--- a/share/jupyter/voila/templates/default/nbconvert_templates/lab.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/lab.tpl
@@ -2,9 +2,12 @@
 
 {% block codecell %}
 {%- if not cell.outputs -%}
-{%- set extra_class="jp-mod-noOutputs" -%}
+{%- set no_output_class="jp-mod-noOutputs" -%}
 {%- endif -%}
-<div class="jp-Cell jp-CodeCell jp-Notebook-cell {{ extra_class }}">
+{%- if not resources.global_content_filter.include_input -%}
+{%- set no_input_class="jp-mod-noInput" -%}
+{%- endif -%}
+<div class="jp-Cell jp-CodeCell jp-Notebook-cell {{ no_output_class }} {{ no_input_class }}">
 {{ super() }}
 </div>
 {%- endblock codecell %}
--- a/share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
@@ -7,6 +7,8 @@
 
 
 {%- block html_head_css -%}
+{{ super() }}
+
 <link rel="stylesheet" type="text/css" href="{{resources.base_url}}voila/static/index.css">
 
 {% if resources.theme == 'dark' %}


voila/handler.py
```diff
@@ -140,7 +140,8 @@ def _jinja_cell_generator(self, nb, kernel_id):
             result = executenb(nb, km=km, cwd=self.cwd, config=self.traitlet_config)
             cell = result.cells[0]  # keep a reference to the executed cell
             nb.cells = all_cells  # restore notebook in case we access it from the template
-            # we don't filter empty cells, since we do not know how many empty code cells we will have
+            # We don't filter empty cells, since we do not know how many empty code cells we will have
+            # We'll do the filtering in CSS
```
maartenbreddels: I think we can/should now remove all reference to stripping these cells in the code.

martinRenou: I disagree. We strip cells in the `_jinja_notebook_execute` method, but not in `_jinja_cell_generator` method, which is pretty heterogeneous so this comment is important.

Also, I think it should not be fixed in CSS but in Python. Only that we did not find a way to do that correctly...

vidartf:
Are there not some cases where a cell can initially have no outputs, but some later code (asynchronously) adds an output to the cell? This would not normally be an issue with nbconvert, but if we are doing progressive rendering, this might be an issue. Is it possible to use the `:empty` CSS selector here?

martinRenou:
Good point, I agree using `:empty` or `:blank` would be better, because it would handle this case.

vidartf:
:blank is still in the future: https://caniuse.com/#feat=mdn-css_selectors_blank

martinRenou:
I've got an alternate solution, I'm going to open another PR

martinRenou:
Alternate solution: #405

martinRenou:
@vidartf 

> Are there not some cases where a cell can initially have no outputs, but some later code (asynchronously) adds an output to the cell?

Do you already have a specific use case in mind? I'm trying to figure out a scenario where this could happen.

maartenbreddels:
display supports a display_id, so that you can change outputs of previous cells. e.g.:
https://nbviewer.jupyter.org/github/jupyter/nbconvert/blob/master/nbconvert/preprocessors/tests/files/Clear%20Output.ipynb

We don't support that actually with voila progressive rendering. I wonder if we should explicitly not support that, or if we should fix this somehow.

Note that if you output an empty Output widget, it will always work (since the cell contains output in the form of a widget)

martinRenou:
> display supports a display_id, so that you can change outputs of previous cells

I also thought about this, but for having a `display_id` you need to display something first no? With a first `display` call

vidartf:
> I also thought about this, but for having a display_id you need to display something first no? With a first display call

Yes, my question basically boils down to: If you do a a `display` call without any data, does that still make a difference for `cell.outputs`?

martinRenou:
> Yes, my question basically boils down to: If you do a a display call without any data, does that still make a difference for cell.outputs?

I actually tried and it seems you can not use `display` with no data. I thought of making an output placeholder with `display(None)` and it actually displays the string `"None"`, so there is an output.

vidartf:
This seems to work: `display()`

The interesting thing is that if I try this:

> handle = display.display(display_id=True)

> handle.update('world')

It doesn't actually work in neither classic notebook nor lab. No errors are raised, but the update isn't actually displayed. So I don't think it is a very common pattern (but strictly speaking, I would have expected that it should work).

vidartf:
(note, that I'm happy with the code in this PR as-is, since neither classic nor lab supports it, just making a note for the future about it)

martinRenou:
I do agree.

I think I prefer the alternate solution though #405. It's basically the same behavior, only that instead of hiding empty cells we simply don't send them.

martinRenou:
@maartenbreddels would you like to have another look at this one?

SylvainCorlay:
I am fine with this solution.

SylvainCorlay:
@maartenbreddels I had a small comment on this one that martin is fixing now. I think this should get in before the release.

maartenbreddels:
I think so too, failure is unrelated. I'll wait a bit more before making a release, in case we need to merge more.

martinRenou:
Done

PR #393, opened 2019-10-02
Use Voila logo for the spinner

martinRenou:
spinner


jtpio:
That looks nice.

Maybe we want to minify / compress the svg file too ?

martinRenou:
We can try, but it looks pretty small already

SylvainCorlay:
This is really cool!

The improvement of the spinner SVG should probably be in a different PR. (Removing the green circle under the yellow will remove the rendering artifact)

SylvainCorlay:
A way to prevent the loading delay may be to inline the SVG in the HTML.

SylvainCorlay:
One simple change to make that will reduce the size would be to remove all the attributes prefixed with `inkscape` in the SVG.

jtpio:
Inkscape has an option to save as "Optimized SVG (*.svg)", which should take care of all these.

martinRenou:
I also removed unused defined colors, whitespaces and new lines, saving ~100kb

jtpio:
Does this affect the perceived loading delay?

share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
```diff
@@ -15,6 +15,23 @@
...
 
+<style>
+  .spinner {
+      -webkit-animation: rotation 2s infinite linear;
+    -moz-animation: rotation 2s infinite linear;
+    -o-animation: rotation 2s infinite linear;
+    animation: rotation 2s infinite linear;
+      transform-origin: 50% 50%;
+      -webkit-transform-origin: 50% 50%;
+      -moz-transform-origin: 50% 50%;
```
SylvainCorlay: I am not sure you still need the `-moz`, `-o` and `-webkit` versions of css properties.

martinRenou: Fixed :) Thanks

martinRenou:
> Does this affect the perceived loading delay?

It does actually, the logo arrived the same time as the text when I tried. But it's only local...

SylvainCorlay:
The way to fix this is to inline the svg in the HTML.

This will also make it possible to set the colors of the parts of the spinner with CSS.

share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
```diff
@@ -42,6 +54,14 @@ a.anchor-link {
         color: var(--jp-content-font-color1);
         font-family: sans-serif;
     }
+
+    .color1{
```
SylvainCorlay: `voila-spinner-color1`, `voila-spinner-color2`...

SylvainCorlay:
This looks great and I am happy to get this in as it is now.

Leaving it open to give a chance to @maartenbreddels to chime in.

martinRenou:
Some CSS fixes coming

jtpio:
Anything more to add / fix before merging it? (so it can go in the next release)

SylvainCorlay:
We should credit Jasmine for her contribution!

maartenbreddels:
Perfect, especially inlining the svg indeed, that should render instantly.

martinRenou:
@jasminefrh I added your contribution :) Thanks a lot for your help!

PR #396, opened 2019-10-03
Speed up progressive rendering

### LINKED ISSUE #395: Progressive rendering is super slow
Since progressive rendering has been added, rendering a Notebook in voila became super slow...

Taking a Notebook that prints "hello" 30 times:
note

I assume we all agree this Notebook should take less than a second to be executed, starting time of the kernel included.

Without the progressive rendering (using a dev version that simply removes this feature):
without

With the progressive rendering:
with

We should either improve this feature, or make it optional

martinRenou:
Should fix #395 

Instead of using the `preprocess` method (which is for preprocessing an entire Notebook) for each cell, I use `preprocess_cell` for each cell

speedup

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -107,6 +107,14 @@ class VoilaExecutePreprocessor(ExecutePreprocessor):
             result = (nb, resources)
         return result
 
+    def preprocess_cell(self, cell, resources, cell_index, store_history=True):
+        try:
+            result = super(VoilaExecutePreprocessor, self).preprocess_cell(cell, resources, cell_index, store_history)
+        except CellExecutionError as e:
+            self.log.error(e)
+            result = [cell]
+        return result
+
     def register_output_hook(self, msg_id, hook):
         # mimics
         # https://jupyterlab.github.io/jupyterlab/services/interfaces/kernel.ikernelconnection.html#registermessagehook
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -94,9 +94,12 @@ class OutputWidget:
 
 class VoilaExecutePreprocessor(ExecutePreprocessor):
     """Execute, but respect the output widget behaviour"""
-    def preprocess(self, nb, resources, km=None):
+    def __init__(self, **kwargs):
+        super(VoilaExecutePreprocessor, self).__init__(**kwargs)
         self.output_hook_stack = collections.defaultdict(list)  # maps to list of hooks, where the last is used
         self.output_objects = {}
+
+    def preprocess(self, nb, resources, km=None):
         try:
             result = super(VoilaExecutePreprocessor, self).preprocess(nb, resources=resources, km=km)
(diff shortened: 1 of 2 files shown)


martinRenou:
@maartenbreddels what do you think? It improves perfomances but the logic was also wrong before (processing each cell like it's a single Notebook)

martinRenou:
Part of the slowness was coming from the fact that internally nbconvert creates new sockets for connecting with the kernel at each `preprocess` call (that is to say for each cell...)

SylvainCorlay:
Reviewed this in person with @martinRenou and this looks great.

martinRenou:
I'm wondering if the `preprocess` implemented in the Voila executor has extra logic that we should have. Maybe the Voila executor should reimplement `preprocess_cell`

martinRenou:
I now reimplement `preprocess_cell` in the Voila executor so that it has a similar behavior to `preprocess`

voila/handler.py
```diff
@@ -133,15 +135,14 @@ def _jinja_cell_generator(self, nb, kernel_id):
...
-        for cell in all_cells:
-            # we execute one cell at a time
-            nb.cells = [cell]  # reuse the same notebook
-            result = executenb(nb, km=km, cwd=self.cwd, config=self.traitlet_config)
-            cell = result.cells[0]  # keep a reference to the executed cell
-            nb.cells = all_cells  # restore notebook in case we access it from the template
-            # we don't filter empty cells, since we do not know how many empty code cells we will have
-            yield cell
+        nb, resources = ClearOutputPreprocessor().preprocess(nb, {'metadata': {'path': self.cwd}})
+        ep = VoilaExecutePreprocessor(config=self.traitlet_config)
```
maartenbreddels: We are missing https://github.com/voila-dashboards/voila/commit/c21b216e227656521da419a82815f001e510b4b0  (see https://github.com/voila-dashboards/voila/blame/master/voila/execute.py#L159)

martinRenou: Mmh... Isn't it line `138`?

voila/execute.py
```diff
@@ -107,6 +107,14 @@ def preprocess(self, nb, resources, km=None):
...
         return result
 
+    def preprocess_cell(self, cell, resources, cell_index, store_history=True):
+        try:
+            result = super(VoilaExecutePreprocessor, self).preprocess_cell(cell, resources, cell_index, store_history)
+        except CellExecutionError as e:
+            self.log.error(e)
+            result = [cell]
+        return result
+
```
maartenbreddels: Why is this needed?

martinRenou: To mimic the `VoilaExecutePreprocessor.preprocess` behavior. I change this behavior a bit in a separate PR concerning error outputs BTW.

martinRenou: The `try except` prevents a `505` server error when the Notebook contains an error

PR #398, opened 2019-10-04
Strip errors if not in debug mode

martinRenou:
errors

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -7,11 +7,45 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 import collections
+import logging
 
 from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError, ExecutePreprocessor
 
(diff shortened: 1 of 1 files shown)


vidartf:
Would it make sense to only show the exception name and message when not in debug mode (i.e. only strip the trace)?

martinRenou:
It could give information about the running code (*e.g.* When the project uses custom error names)

voila/execute.py
```diff
@@ -7,13 +7,47 @@
...
 from ipykernel.jsonutil import json_clean
 
 
+def strip_code_cell_errors(cell):
+    """Strip any error outputs and traceback from a code cell."""
+    outputs = cell['outputs']
+
+    error_outputs = [output for output in outputs if output['output_type'] == 'error']
+
+    error_message = 'There was an error when executing cell [{}]. Please run Voila with --debug for seeing the exact error.'.format(cell['execution_count'])
```
SylvainCorlay: for seeing -> to see

the exact error -> the error message

PR #404, opened 2019-10-08
Update nbconvert

martinRenou:
This will hopefully fix the CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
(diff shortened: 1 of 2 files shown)


vidartf:
As long as the filter does not remove any cell cells where the output is populated later (I'm not sure what `cell.outputs` is in that case), this is OK as well.

martinRenou:
I suppose you wanted to comment on the #405 PR :) But yeah you've got a point. We need to think more about all of this. `:empty` might be better then, but `:empty` requires not having any line returns... Which is an annoying constraint for the template.

SylvainCorlay:
Merging even though it does not fully fix the CI.

PR #405, opened 2019-10-08
WIP - Filter empty cells

martinRenou:


vidartf:
As long as the filter does not remove any cell cells where the output is populated later (I'm not sure what cell.outputs is in that case), this is OK as well.

maartenbreddels:
My preference goes to solving this in the template, and for instance not do this in the debug template.

martinRenou:
Well, we actually still do that in the template. The only difference is that we don't add empty cells in the page, instead of only making them invisible.

vidartf:
A way to get an empty output (but I guess this makes the cell be not empty, so this should still be ok):
```py
display({}, raw=True, display_id=True)
```

martinRenou:
But:
```python
handle = display({}, raw=True, display_id=True)
handle.update({'text/html': 'test'})
```
does not seem to show anything (in classic Notebook)

vidartf:
It works in lab though.

martinRenou:
> It works in lab though.

And actually it works in Voila as well. Both using master and this branch.

vidartf:
Yes, but the question is then: Does #391 work?

martinRenou:
Yes, for the same reason. `cell.outputs` is not None in the template, so we don't hide it in CSS

maartenbreddels:
As discussed, cells should always be returned, since voila has the option to show the source code. Also, filtering out cells is impossible, since we don't know a priori if a cell will not contain output, and to make the progress indicator work, we need to return as many cells as there are in the notebook.

I think that only leaves the option open for a template to have custom behavior for empty output cells, e.g. only have an if statement in the template:
```
{% if 'outputs' in cell and cell.outputs %>
```

martinRenou:
Closing in favor of #391

PR #413, opened 2019-10-10
Expose base_url to browser-open template

martinRenou:
This is for homogeneity with other template files under the "templates/" directory. It allows having the same base template accessing the same assets under the static directory.

For example, when having a `base.html` file like this:
```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="{{base_url}}voila/static/style.min.css">
  </head>

  <body>
    {% block body %}
    {% endblock %}
    
    <script src="{{base_url}}voila/static/javascript.min.js"></script>
  </body>
</html>
```

We can now make the `browser-open.html` and the `404.html` template inherit from the same `base.html` template.

This is backward incompatible for other templates out there.

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -524,7 +524,7 @@ class Voila(Application):
 
             jinja2_env = self.app.settings['jinja2_env']
             template = jinja2_env.get_template('browser-open.html')
-            fh.write(template.render(open_url=url))
+            fh.write(template.render(open_url=url, base_url=url))
 
         def target():
             return browser.open(urljoin('file:', pathname2url(open_file)), new=self.webbrowser_open_new)


martinRenou:
@maartenbreddels @SylvainCorlay @jtpio What do you think about this change?

maartenbreddels:
I would add base_url, the facts that they are the same, might not be true in the future?
But yes, I like the idea of having at least base_url available.

maartenbreddels:
> This is backward incompatible for other templates out there.

And adding it will not break other templates.

SylvainCorlay:
I am not sure about this - please hold on before merging.

martinRenou:
This is going to ease template developers' life... Being able to inherit from the same template and having access to the same assets is important IMO.

maartenbreddels:
I think we need to add a list of variables available for which templates, especially once we merge/unify all templates in one directory.

martinRenou:
> the facts that they are the same, might not be true in the future?

Mmmh.. The purpose of this change is especially because they are the same. Why do you say it might not be true in the future?

maartenbreddels:
open_url can be a more nested path right, for instance, you can start a server, and open the url to 
`{base_url}/somedir/somenotebook.ipynb`.

SylvainCorlay:
The current name corresponds to what is exposed to other jupyter frontends such as jupyterlab.

martinRenou:
I see! So it makes a lot of sense to have them separated then. I'm all for it.

maartenbreddels:
I think all templates should have base_url exposed anyway, which would make them more uniform.

jtpio:
Exposing `base_url` sounds like a good idea indeed.

martinRenou:
@SylvainCorlay Is this ready to go in?

Issue #260, opened 2019-06-18
Logout button doesn't work when deployed along with JupyterHub


romulomanzano:
Exploring Voila as a tool to expose interactive notebooks to non technical users. 

My approach has been to enable the pre-fix functionality (so we can simply use voila at the root level) to distribute the relevant links to the users. The only problem so far is that the logout button doesn't seem to work at all
![voila-logout](https://user-images.githubusercontent.com/11479305/59725601-2fb9c280-91fc-11e9-86c1-732db0bd4869.gif)

djangoliv:
It would be nice to inherit the header of jupyterhub in case it is used in jupyterhub. 
It would be possible to make a 'logout' and a 'control panel' from the voila pages.

PR #415, opened 2019-10-11
Clean templates

martinRenou:
There was a lot of unwanted noise in those templates like the "login widget". I suppose it was due to a copy/paste from JupyterLab/Notebook templates.

I also improved the code indentation.

### DIFF
--- a/share/jupyter/voila/templates/default/templates/404.html
+++ b/share/jupyter/voila/templates/default/templates/404.html
@@ -1,5 +1,5 @@
 {% extends "error.html" %}
-{% block error_detail %}
-<p>{% trans %}You are requesting a page that does not exist!{% endtrans %}</p>
-{% endblock %}
 
+{% block error_detail %}
+  <p>{% trans %}You are requesting a page that does not exist!{% endtrans %}</p>
+{% endblock %}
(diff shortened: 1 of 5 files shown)


martinRenou:
Putting this in WIP, we need to check if this login_widget is for the authentication when the token is not in the url

jtpio:
> Putting this in WIP, we need to check if this login_widget is for the authentication when the token is not in the url

:+1: 

This `login_widget` was also indirectly mentioned in https://github.com/voila-dashboards/voila/issues/260 I believe.

martinRenou:
It seems to me that those login/logout widgets are leftovers...

When you browse Voila without the token you get redirected to `http://localhost:8888/login?next=%2Fvoila`, so it is not handled by Voila.

And the issue you pointed to @jtpio is comforting me in thinking those widgets are useless.

maartenbreddels:
What is page.html actually? Is that the classic notebook template? If so, we can remove it I guess.

jtpio:
> It seems to me that those login/logout widgets are leftovers...

That was my impression too.

martinRenou:
So, are you ok with this @maartenbreddels ?

maartenbreddels:
Yeah, I'm a bit unsure about the login/logout stuff. I'm not so familiar with it, don't know if it functions at all, and I wonder if people already depend on it.

@timkpaine : you might be using the login/logout features, is it functional at all?

If it does not function, I'm ok with this PR.

martinRenou:
@maartenbreddels you added this `login_widget` part. https://github.com/voila-dashboards/voila/commit/28faacc9b03b160fd8fa920ad045f4ec0667ab67. Was it some copy-paste from somewhere else?

Issue #381, opened 2019-09-11
warning message show in R kernel


hiteck88:
Hello, 
 I am using jupyter lab R kernel, I have set options(warn=-1), so no warning is shown in the original notebook file. But it still shows in the voila. How to hide the warning message?
Thanks

![image](https://user-images.githubusercontent.com/8680807/64716415-56506800-d490-11e9-8c0a-ebca8c3f56a5.png)

SylvainCorlay:
Thanks @hiteck88! I think that we should filter out warning by defaults.

A PR fixing that would be welcome!

Issue #252, opened 2019-06-17
Warnings are not hided in voila when hided in the Jupyter


ArVar:
I've noticed, that when hiding warnings in Jupyter/IPython via:
```
import warnings
warnings.filterwarnings('ignore')
warnings.simplefilter('ignore')
```
it is working pretty well in the notebook itself but not, when run with voila.
Is there a possible way to achieve this?

It might also be related to #171 .

Best regards,
Arthur

PR #416, opened 2019-10-11
Strip code cell warnings

GeorgianaElena:
This PR strips cell warnings when voila is run without the ``--debug`` flag, similar with how it's done for errors.

I'm not sure if it's enough to fix the related issues (https://github.com/voila-dashboards/voila/issues/381, https://github.com/voila-dashboards/voila/issues/252) or if there's a better approach to this :eyes: .


martinRenou:
I guess it should be enough for fixing the issues you mentioned

voila/execute.py
```diff
@@ -35,13 +35,30 @@ def strip_code_cell_errors(cell):
...
+    # There is no 'outputs' key for markdown cells
+    if 'outputs' not in cell:
+        return cell
+
+    outputs = cell['outputs']
+
+    warning_outputs = [output for output in outputs if output['output_type'] == 'stream' and output['name'] == 'stderr']
+
+    for output in warning_outputs:
+        output['name'] = ''
```
martinRenou: Shouldn't you set the "text" key to an empty string as well?

GeorgianaElena: I thought about this, but by setting the "name" key to an empty string, the warning message doesn't get printed to any output stream and I thought it won't make a difference if I'd change/emptied the text. But I'm not sure I got the logic right :grimacing:. Am I making any sense?

martinRenou: Yes! Maybe the front-end does not show anything by default if the name is empty, I did not check what the logic is. But just to be sure I would set the text to empty string too...

Thinking about it, it might be even better to remove those outputs from the cell outputs list. What do you think?

GeorgianaElena: I agree it'll be better to remove the warning outputs entirely.

GeorgianaElena: Updated the PR! Thanks a lot for the feedback :)

voila/execute.py
```diff
@@ -35,13 +35,30 @@ def strip_code_cell_errors(cell):
...
+    # There is no 'outputs' key for markdown cells
+    if 'outputs' not in cell:
+        return cell
+
+    outputs = cell['outputs']
+
+    def is_warning(output):
+        return output['output_type'] == 'stream' and output['name'] == 'stderr'
+
+    cell['outputs'] = [output for output in outputs if not is_warning(output)]
```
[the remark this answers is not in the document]
martinRenou: I don't want to block your PR for this kind of details so this can be done later if you want this fix soon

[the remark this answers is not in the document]
GeorgianaElena: No worries, there's no hurry, I'm just excited for my first PR to Voila :heart:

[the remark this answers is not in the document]
martinRenou: Thanks a lot for your PR :) !!

martinRenou:
@SylvainCorlay @maartenbreddels @jtpio

maartenbreddels:
Hi Georgiana,

thanks for you first time contribution! I recognize that this is something you often want, but I am not sure this should be default behavior. Are there examples of where the output to stderr is actually something you would like to show to a user? Maybe not, and we should just merge this.

Anyone else with some input?

cheers,

Maarten

martinRenou:
> I recognize that this is something you often want, but I am not sure this should be default behavior

The same way we strip errors by default, I suppose we should strip warnings by default (they could leak some code, information...).
Note that when running with `--debug` you will see warning with this PR.

maartenbreddels:
But with an exception, you get feedback that something happened, now we basically send stderr to /dev/null. I am not sure that it should go under the same flag, because now you cannot hide exceptions, but show stderr. The workaround for this could be to use an output widget btw.

martinRenou:
So you mean you would rather see something like:

```python
warnings = [
    output for output in cell['outputs']
    if output['output_type'] == 'stream' and output['name'] == 'stderr'
]

for warning in warnings:
    warning['name'] = 'warning'
    warning['text'] = 'There was a warning in cell {}, run Voila with --debug if you want to see it.'
```

?

jtpio:
> The same way we strip errors by default, I suppose we should strip warnings by default (they could leak some code, information...).

I also think it makes sense to strip warnings by default.

martinRenou:
@maartenbreddels do you mind having another look a this if we're making a release? Do you think we should have an equivalent logic as with error stripping (see my previous comment)?

maartenbreddels:
Ok, I think indeed the default behaviour should be not to show the error output, since it is stderr or warnings, you don't want to see it, and it may leak information. What I am unsure about, is that `--debug` should be the only way to turn this on (since it will also show exception). But if we add extra flags for this later, this won't be a breaking change.

TLDR; let's merge this as it, I prefer to squash and merge it.

jtpio:
> What I am unsure about, is that `--debug` should be the only way to turn this on (since it will also show exception).

That could be tweaked with `--Voila.log_level` then?

maartenbreddels:
> That could be tweaked with `--Voila.log_level` then?

I think that's more for the internal logging. That may be a per template settting actually, your debug template might want to show that by default? Or should it be a per app setting?

maartenbreddels:
Linux is green, OSX takes ages, I'm going to merge this. Thanks for your first contribution @GeorgianaElena

SylvainCorlay:
Thanks @GeorgianaElena ! This should be in the release later today.

jtpio:
> I think that's more for the internal logging. That may be a per template settting actually, your debug template might want to show that by default? Or should it be a per app setting?

Well for now `--debug` is just an alias for `--Voila.log_level=logging.DEBUG`.

Issue #392, opened 2019-10-02
Autodetection of free port


JoergVanAken:
First of all: voila is just great! 

Last week I deployed a little application which the user can easily start by double clicking a windows batch file. This works fine but there is one little thing: The app cannot be started in multiple windows because the port number is fixed. I fixed this on my own by looping over a a few port numbers in the calling batch script. But when I start a new jserver a free port is autodetected. I would love to see this feature in voila, too.

jtpio:
Thanks @JoergVanAken, that would indeed be very useful.

This is something that should normally come for free once https://github.com/voila-dashboards/voila/pull/270 lands.

In the meantime, we could replicate the [logic of the notebook server](https://github.com/jupyter/notebook/blob/4a9cfa7518c911d1b89b201acd0083adc6fb680e/notebook/notebookapp.py#L1440-L1459) in voila.

PR #424, opened 2019-10-13
add auto port detection as in jupyter notebook

katsar0v:
Fixes https://github.com/voila-dashboards/voila/issues/392 :)

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -19,6 +19,8 @@ import shutil
 import signal
 import socket
 import webbrowser
+import errno
+import random
 
 try:
     from urllib.parse import urljoin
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @katsar0v for looking at this.

voila/app.py
```diff
@@ -488,9 +494,41 @@ def stop(self):
...
+                    continue
+                else:
+                    raise
+            else:
+                self.port = port
+                success = True
+                break
+
+        if not success:
+            self.log.critical(_('ERROR: the notebook server could not be started because '
```
jtpio: We probably want to say "voila" here.

katsar0v: I made a new commit :)

voila/app.py
```diff
@@ -488,9 +494,41 @@ def stop(self):
...
+                    self.log.warning(_("Permission to listen on port %i denied") % port)
+                    continue
+                else:
+                    raise
+            else:
+                self.port = port
+                success = True
+                break
+
+        if not success:
```
maartenbreddels: could be a single else. e.g.
```
for port in ...
   ...
else:
  ...
```

Issue #432, opened 2019-10-16
Switch to Azure Pipelines / GitHub Actions?


jtpio:
This topic has been discussed before in person but never phrased as an issue.

For now Voila is tested with Travis on Linux and Mac. But we should also have tests on Windows.

Traditionally AppVeyor has been the go-to option to complement Travis. However there is now `Azure Pipelines` and `GitHub Actions` that offer such option under the same interface.

cc @maartenbreddels who was also interested in moving away from Travis.

jtpio:
Fixed by https://github.com/voila-dashboards/voila/pull/445.

Tests now run on GitHub Actions as well as on Travis. We should eventually add support for Windows too, but that can be tracked separately.

PR #444, opened 2019-10-25
Setup GitHub Action for CI

jtpio:
Setup GitHub Action for testing.

This is to evaluate GitHub Action as a replacement for Travis CI, and eventually add Windows testing to Voila (xref #432).

I quickly tried `windows-latest` as well, but ran into issues with conda. And also the tests using xeus-cling will have to be disabled for Windows.

@maartenbreddels there is an official `actions/setup-python@v1` for Python, but for `conda` we still have to do things manually it seems (setting up the path and activating the environment).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,63 @@
+name: Tests
+
+on: [push]
+
+jobs:
+  run:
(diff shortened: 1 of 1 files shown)


jtpio:
Workflow file is not picked up from the fork. Opening a new PR from a branch.

PR #445, opened 2019-10-25
Setup GitHub Actions for CI

jtpio:
Setup GitHub Actions for testing.

Github Actions Status

This is to evaluate GitHub Actions as a replacement for Travis CI, and eventually add Windows testing to Voila (xref #432).

I quickly tried `windows-latest` as well, but ran into issues with conda. And also the tests using `xeus-cling` will have to be disabled for Windows.

@maartenbreddels there is an official `actions/setup-python@v1` for Python, but for `conda` we still have to do things manually it seems (setting up the path and activating the environment).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,63 @@
+name: Tests
+
+on: [push]
+
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: [2.7, 3.6, 3.7]
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Add conda to $PATH
+      run: echo ::add-path::$CONDA/condabin
+
+    - name: Update conda on Mac
+      if: matrix.os == 'macos-latest'
+      run: |
+        # sudo required?
+        sudo conda update -y -n base conda setuptools
+        
+    - name: Update conda on Linux
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        conda update -y -n base conda setuptools
+    
+    - name: Init conda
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: [2.7, 3.6, 3.7]
```
jtpio: 3.8 not available on conda yet

.github/workflows/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    - uses: actions/checkout@v1
+
+    - name: Add conda to $PATH
+      run: echo ::add-path::$CONDA/condabin
+
+    - name: Update conda on Mac
+      if: matrix.os == 'macos-latest'
+      run: |
+        # sudo required?
+        sudo conda update -y -n base conda setuptools
```
jtpio: Permissions might be a bit strict with actions. This is what we get without `sudo`:

```
NotWritableError: The current user does not have write permissions to a required path.
  path: /Users/runner/.conda/pkgs/urls.txt
  uid: 501
  gid: 20

If you feel that permissions on this path are set incorrectly, you can manually
change them by executing

  $ sudo chown 501:20 /Users/runner/.conda/pkgs/urls.txt

In general, it's not advisable to use 'sudo conda'.
```

jtpio:
And to answer your question from earlier today @maartenbreddels, it seems to be faster than on Travis :rocket:

maartenbreddels:
I see no harm in merging this to see if we prefer this over Travis.

jtpio:
Yes. Let's merge then?

PR #449, opened 2019-10-26
Fix filtering of empty code cells.

mkcor:
When running `$ voila index.ipynb --template=reveal`, the right arrow (to navigate to the next slide) would become unavailable at the first code cell. The issue did not arise when using `voila index.ipynb --template=reveal --strip_sources=False` so we suspected an issue with filtering 'empty' code cells.

We recovered the expected behaviour by removing the `filter_empty_code_cells()` call in `handler.py`. It seems safe to remove this function entirely, because filtering has been handled in the front-end with CSS since the implementation of progressive rendering.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -22,15 +22,6 @@ from .execute import executenb, VoilaExecutePreprocessor
 from .exporter import VoilaExporter
 
 
-# Filter for empty cells.
-def filter_empty_code_cells(cell, exporter):
-    return (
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Note: this only impacts non-progressively rendered notebooks which use that method.

cc @maartenbreddels @martinRenou

SylvainCorlay:
Merging!

PR #450, opened 2019-10-27
Include both jupyter_server and notebook server in README.

mkcor:
A Jupyter server is either a `notebook` server or a `jupyter_server` (as understood in https://voila.readthedocs.io/en/latest/using.html#as-a-jupyter-server-extension).


SylvainCorlay:
Thx

Issue #423, opened 2019-10-13
Add documention for GridspecLayout


jtpio:
Now that ipywdgets `7.5` is out, it would be great to have an example in the docs to show that it is possible to build a relatively advanced dashboard using existing widgets and the `GridspecLayout`.

![image](https://user-images.githubusercontent.com/591645/66713886-155db300-edb0-11e9-9ac6-7eb373736ca6.png)

Doc: https://ipywidgets.readthedocs.io/en/latest/examples/Layout%20Templates.html#Grid-layout

Blog post: https://blog.jupyter.org/introducing-templates-for-jupyter-widget-layouts-f72bcb35a662

It would also be great to have a live example in the [gallery](https://github.com/voila-dashboards/gallery)!

PR #452, opened 2019-10-28
Added GridspecLayout example

GeorgianaElena:
Added a new notebook that builds a multiplication table "game", using *GridspecLayout*.
It's my first time playing with widgets, so I'm not sure I did everything right :relaxed:

closes https://github.com/voila-dashboards/voila/issues/423

### DIFF
--- a/notebooks/gridspecLayout.ipynb
+++ b/notebooks/gridspecLayout.ipynb
@@ -1,0 +1,109 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Which multiplication table do you want to learn?"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "6f9816bc14dd4a349da63583524c590a",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "GridspecLayout(children=(Dropdown(index=1, layout=Layout(grid_area='widget001'), options=(1, 2, 3, 4, 5, 6, 7,…"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from ipywidgets import GridspecLayout, Button, BoundedIntText, Valid, Layout, Dropdown\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @GeorgianaElena, this is a fun example!

Do you think we should point to the example notebooks (currently in the `notebooks` folder) somewhere in the [documentation](https://voila.readthedocs.io)?

GeorgianaElena:
Thanks @jtpio, I really had fun working on it.

>Do you think we should point to the example notebooks (currently in the notebooks folder) somewhere in the documentation?

It seems that the example notebooks are mentioned a few times in the docs (e.g. [here](https://voila.readthedocs.io/en/stable/contribute.html#running-the-examples) and [here](https://voila.readthedocs.io/en/stable/deploy.html#setup-an-example-project)).
But I think it would be nice to have a dedicated section *"About the example notebooks"* (or some other (better) title), where we describe in a few words the notebooks and link to the section that explains [how to run them](https://voila.readthedocs.io/en/stable/contribute.html#running-the-examples).
I would put this in [*Deploying Voila*](https://voila.readthedocs.io/en/stable/deploy.html#deploying-voila), just before [set-upping an example project](https://voila.readthedocs.io/en/stable/deploy.html#setup-an-example-project) it's described.
What do you think?

jtpio:
> But I think it would be nice to have a dedicated section _"About the example notebooks"_ (or some other (better) title), where we describe in a few words the notebooks and link to the section that explains [how to run them](https://voila.readthedocs.io/en/stable/contribute.html#running-the-examples).

That would be very valuable yes :+1: 

At the moment it's a bit hidden in the contributing guide, which is probably not the first thing end users look at.

GeorgianaElena:
@jtpio, should I address this in the current PR or should I open a new one?

jtpio:
@GeorgianaElena you're right this could be done in a different PR

notebooks/gridspecLayout.ipynb
```diff
@@ -0,0 +1,109 @@
...
+    "# Which multiplication table do you want to learn?"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [
```
jtpio: I was quickly checking the other examples, and their outputs are removed from the notebook. So we should probably do the same here.

GeorgianaElena: Thanks @jtpio! I removed the cell outputs and added a notebook description. Will open a new PR for the docs.

PR #454, opened 2019-10-29
Fixup voila configuration

SylvainCorlay:
cc @mkcor @JovanVeljanoski.

in fact base_template is not part of the traitlets configuration.

### DIFF
--- a/tests/test_template/share/jupyter/voila/templates/test_template/conf.json
+++ b/tests/test_template/share/jupyter/voila/templates/test_template/conf.json
@@ -1,6 +1,5 @@
 {
   "traitlet_configuration": {
-    "base_template": "test_template",
     "resources": {
       "test_template": {
         "test_resource": "test resource from config file"
--- a/voila/app.py
+++ b/voila/app.py
@@ -364,12 +364,11 @@ class Voila(Application):
                     # load the template-related config
                     with open(p) as json_file:
                         conf = json.load(json_file)
-                    # check that config file is meant for template in use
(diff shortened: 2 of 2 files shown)


mkcor:
By 'traitlets configuration,' do you mean an object inheriting from any of https://traitlets.readthedocs.io/en/stable/config.html? If so, you are right. The `base_template` key was our own idea, as discussed (but not broadly enough) over at https://github.com/voila-dashboards/voila/pull/326#issuecomment-525682901.

In spirit, I definitely like the idea of replacing the `assert` with an `if` or a `try/except`. But I think keeping the `base_template` key in the template config file could make sense, especially in [cases](https://github.com/voila-dashboards/voila/pull/372#discussion_r322184430) anticipated by @maartenbreddels.

mkcor:
I'm wondering if we should to remove line `"base_template": "test_template",` from the test template config file... It's one thing not to require it. But we still want to be fine with it, don't we?

SylvainCorlay:
> I'm wondering if we should remove line "base_template": "test_template", from the test template config file... It's one thing not to require it. But we still want to be fine with it, don't we?

Thanks. It is in fact a top-level key for the json configuration file (and not a trait).

mkcor:
> Thanks. It is in fact a top-level key for the json configuration file (and not a trait).

Not sure what you mean by 'in fact' because we basically need to decide one way and stick with it...

SylvainCorlay:
> Not sure what you mean by 'in fact' because we basically need to decide one way and stick with it...

By *in fact* iI mean that this is how it was used *de facto* :)

PR #455, opened 2019-10-29
Test for recursion error

SylvainCorlay:


SylvainCorlay:
We could reproduce what Marianne was describing. Closing.

PR #458, opened 2019-11-01
New docs section about example notebooks

GeorgianaElena:
Following the discussion in https://github.com/voila-dashboards/voila/pull/452, I added a section in the docs about the example notebooks.

### DIFF
--- a/docs/source/using.rst
+++ b/docs/source/using.rst
@@ -89,3 +89,24 @@ When Voilà is run on a notebook, the following steps occur:
    Jupyter server.
 #. When users access the page, the widgets on the page have access to
    the underlying Jupyter kernel.
+
+The example notebooks
+=====================
+
+The `notebooks directory <https://github.com/voila-dashboards/voila/tree/master/notebooks>`__
+contains a collection of Jupyter notebooks that can be rendered using Voilà:
+
+* **basics.ipynb** - a notebook with interactions requiring a roundtrip to the kernel.
+* **bqplot.ipynb** - uses custom Jupyter widgets such as
+  `bqplot <https://github.com/bloomberg/bqplot>`__.
+* **dashboard.ipynb** - uses gridstack.js for the layout of each output.
+* **interactive.ipynb** - makes use of ipywidget's @interact.
+* **ipympl.ipynb** - contains custom interactive matplotlib figures using the
+  `ipympl <https://github.com/matplotlib/jupyter-matplotlib>`__ widget.
+* **ipyvolume.ipynb** - uses custom Jupyter widgets such as
+  `ipyvolume <https://github.com/maartenbreddels/ipyvolume>`__.
+* **xleaflet.ipynb** - a notebook that uses C++ kernel and interactive widgets
+
+These examples demonstrate different interactive HTML widgets and can be used as inspiration
+for getting started with Voilà. Checkout the `instructions <contribute.html#running-the-examples>`_
+in the contributor guide for how to run them.
--- a/docs/source/using.rst
+++ b/docs/source/using.rst
@@ -100,6 +100,9 @@ contains a collection of Jupyter notebooks that can be rendered using Voilà:
 * **bqplot.ipynb** - uses custom Jupyter widgets such as
   `bqplot <https://github.com/bloomberg/bqplot>`__.
 * **dashboard.ipynb** - uses gridstack.js for the layout of each output.
+* **gridspecLayout.ipynb** - uses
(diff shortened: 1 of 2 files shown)


GeorgianaElena:
I first thought this section should be in `Deploying Voilà`, but I believe it makes more sense to be in `Using Voilà`. What do you think?

jtpio:
> I first thought this section should be in `Deploying Voilà`, but I believe it makes more sense to be in `Using Voilà`. What do you think?

`Using Voilà` sounds like a better place indeed.

docs/source/using.rst
```diff
@@ -89,3 +89,24 @@ When Voilà is run on a notebook, the following steps occur:
...
+* **interactive.ipynb** - makes use of ipywidget's @interact.
+* **ipympl.ipynb** - contains custom interactive matplotlib figures using the
+  `ipympl <https://github.com/matplotlib/jupyter-matplotlib>`__ widget.
+* **ipyvolume.ipynb** - uses custom Jupyter widgets such as
+  `ipyvolume <https://github.com/maartenbreddels/ipyvolume>`__.
+* **xleaflet.ipynb** - a notebook that uses C++ kernel and interactive widgets
+
+These examples demonstrate different interactive HTML widgets and can be used as inspiration
+for getting started with Voilà. Checkout the `instructions <contribute.html#running-the-examples>`_
+in the contributor guide for how to run them.
```
jtpio: Maybe it would actually be better to move these instructions from the contributing guide to the user guide?

GeorgianaElena: Good idea! I moved the instructions to the user guide and left a note in the contributing guide about them. Thanks!

docs/source/using.rst
```diff
@@ -89,3 +89,36 @@ When Voilà is run on a notebook, the following steps occur:
...
+The example notebooks
+=====================
+
+The `notebooks directory <https://github.com/voila-dashboards/voila/tree/master/notebooks>`__
+contains a collection of Jupyter notebooks that can be rendered using Voilà:
+
+* **basics.ipynb** - a notebook with interactions requiring a roundtrip to the kernel.
+* **bqplot.ipynb** - uses custom Jupyter widgets such as
+  `bqplot <https://github.com/bloomberg/bqplot>`__.
+* **dashboard.ipynb** - uses gridstack.js for the layout of each output.
```
jtpio: We could also include the new `gridspecLayout.ipynb` from https://github.com/voila-dashboards/voila/pull/452.

docs/source/using.rst
```diff
@@ -89,3 +89,36 @@ When Voilà is run on a notebook, the following steps occur:
...
+* **xleaflet.ipynb** - a notebook that uses C++ kernel and interactive widgets
+
+These examples demonstrate different interactive HTML widgets and can be used as inspiration
+for getting started with Voilà.
+
+To **run the example notebooks**, a few additional libraries can be installed :
+
+.. code-block:: bash
+
+   conda install -c conda-forge ipywidgets ipyvolume bqplot scipy
```
jtpio: Since there is an `environment.yml` in the root folder (used for Binder), we could also suggest something like:

```
Or alternatively:

   conda env create
```

jtpio:
Thanks for the update! 

I left two other inline comments, and then it should be ready to go :rocket:

docs/source/using.rst
GeorgianaElena: Thanks @jtpio :sunflower:

docs/source/using.rst
GeorgianaElena: Good idea, thanks!

docs/source/using.rst
```diff
@@ -89,3 +89,45 @@ When Voilà is run on a notebook, the following steps occur:
...
+The `notebooks directory <https://github.com/voila-dashboards/voila/tree/master/notebooks>`__
+contains a collection of Jupyter notebooks that can be rendered using Voilà:
+
+* **basics.ipynb** - a notebook with interactions requiring a roundtrip to the kernel.
+* **bqplot.ipynb** - uses custom Jupyter widgets such as
+  `bqplot <https://github.com/bloomberg/bqplot>`__.
+* **dashboard.ipynb** - uses gridstack.js for the layout of each output.
+* **gridspecLayout.ipynb** - uses
+  `GridspecLayout <https://ipywidgets.readthedocs.io/en/latest/examples/Layout%20Templates.html#Grid-layout>`__
+  for the layout of different widges.
```
jtpio: Minor typo spotted:

```suggestion
  for the layout of different widgets.
```

GeorgianaElena: Thanks a lot @jtpio! Sorry it took so long to address this.

PR #462, opened 2019-11-05
Spinner update improvements

martinRenou:
- rename `voila_process` function into `updateSpinner`
- replace `Executing 1 of 6` text by `Executed 1 of 6`. Because it's actually the number of executed cell that is displayed.
- indentation improvements

### DIFF
--- a/share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
@@ -46,15 +46,15 @@ a.anchor-link {
 {{ spinner.html() }}
 
 <script>
-  var voilaCellExecuteAfter = function(cell_index, cell_count) {};
+  var voilaCellExecuteAfter = function(cellIndex, cellCount) {};
 
-  var voilaCellExecuteBefore = function(cell_index, cell_count) {
+  var voilaCellExecuteBefore = function(cellIndex, cellCount) {
     var el = document.getElementById("loading_text");
-    el.innerHTML = `Executing ${cell_index} of ${cell_count}`;
+    el.innerHTML = `Executing ${cellIndex} of ${cellCount}`;
   };
 
   // Backward compatibility
-  var voila_process = function(cell_index, cell_count) {};
+  var voila_process = function(cellIndex, cellCount) {};
 </script>
 
 <div id="rendered_cells" style="display: none">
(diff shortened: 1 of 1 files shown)


maartenbreddels:
> * rename `voila_process` function into `updateSpinner`

I agree that `voila_process` wasn't the best name, and should have been camelcasting. However, this might break other templates, and I'm not sure the name is better, since it implies a spinner. 

Also, maybe it is better to show progress before the execution of a cell. Imagine the first cell takes 1 minutes to complete, I think it's more useful to see `Executing cell 0 of 10...` in that case. It takes a bit more code to do this, but what do you think?

martinRenou:
I totally agree, I will update my PR

maartenbreddels:
any idea for a better name btw? I wanted to avoid name collisions, so a voila prefix seemed natural. 
Maybe have:
 * `voilaCellRunBefore(cellIndex, cellCount)`
* `voilaCellRunAfter(cellIndex, cellCount)`

This gives people the freedom to implement a before of after progress update (The text may need to be updated before, the progress indicator after?)

martinRenou:
I agree with those names :)

But would you like them to be renamed now or should we stay backward compatible with custom templates?

Maybe we can keep the `voila_process` call around for a bit and introduce `voilaCellRunBefore` and `voilaCellRunAfter` now?

martinRenou:
Also, the spinner text is initialized with:

```html
<h2 id="loading_text">Running {{nb_title}}...</h2>
```

Is it a good idea to expose the Notebook name?

maartenbreddels:
It is by default in the <title> tag right?

Yes, we could call both voilaCellRunAfter and voila_process for backward compat.

martinRenou:
When will we get rid of `voila_process` ?

martinRenou:
Shouldn't we initialize the spinner text with:

```html
<h2 id="loading_text">Executing 0 of {{ cell_count }}...</h2>
```

?

maartenbreddels:
Never :)
Or we need to do a check of all the templates we maintain to see if somebody uses it. It could be that nobody does actually, since nothing in voila.tpl can be reused since #208 is not merged yet.

maartenbreddels:
> ```
> <h2 id="loading_text">Executing 0 of {{ cell_count }}...</h2>
> ```

We could, but I think this should be the text set when `voilaCellBefore(0, X)` is called, the time between the initial text and this call might be short, so we could use this text. Or we can have something like `Waiting for executing first cell`. 

Another thing.. should we use 1 based indexing? I think we should since we're facing users, not programmers :)

martinRenou:
> Another thing.. should we use 1 based indexing? I think we should since we're facing users, not programmers :)

Yes, and this way we stay close to what other Notebook frontends do.

>  I think this should be the text set when voilaCellBefore(0, X)

I'm a bit confused, I don't see where this call should be put.

maartenbreddels:
> > I think this should be the text set when voilaCellBefore(0, X)
> 
> I'm a bit confused, I don't see where this call should be put.

With 1 indexing it should be 1, not 0, sorry for the confusion.

Lets switch from natural language to python/pseudocode, I expect sth like:
```
voilaCellBefore(1, nb.cells.length)
for i, cell in enumerate(cell_iterator):
  render_cell()
  voilaCellAfter(i+1, nb.cells.length)
  voila_process(i+1, nb.cells.length) // backward compat
  if i < nb.cells.length-1: # don't execute the last iteration
     voilaCellBefore(i+2, nb.cells.length)
```

martinRenou:
If we do that, the text shown by `voilaCellAfter` will not be visible long enough for the user to even see it.

martinRenou:
Done :)

share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
```diff
@@ -49,41 +49,52 @@ a.anchor-link {
...
+    <script>voilaCellExecuteBefore(1, {{ cell_count }});</script>
+    {%- for cell in cell_generator(nb, kernel_id) -%}
+      {% set cellloop = loop %}
+      {%- block any_cell scoped -%}
+        <script>voila_process({{ cellloop.index }}, {{ cell_count }});</script>
+        {{ super() }}
+        <script>
+          voilaCellExecuteAfter({{ cellloop.index }}, {{ cell_count }});
+
+          if ({{ cellloop.index }} != {{ cell_count }}) {
```
maartenbreddels: Pedantic mode: I think we should do the condition in jinja, so no code gets outputted, otherwise we get a:
```
if (3 != 3) {
```
In the output.

martinRenou:
@maartenbreddels @jtpio last change request resolved. I think this is good to go.

share/jupyter/voila/templates/default/nbconvert_templates/voila.tpl
```diff
@@ -46,41 +46,52 @@ a.anchor-link {
...
+    {% set cell_count = nb.cells|length %}
+    {#
+    Voila is using Jinja's Template.generate method to not render the whole template in one go.
+    The current implementation of Jinja will however not yield template snippets if we call a blocks' super()
+    Therefore it is important to have the cell loop in the template.
+    The issue for Jinja is: https://github.com/pallets/jinja/issues/1044
+    #}
+    <script>voilaCellExecuteBefore(1, {{ cell_count }});</script>
+    {%- for cell in cell_generator(nb, kernel_id) -%}
+      {% set cellloop = loop %}
```
jtpio: Not strictly related to this change, but it looks like `cellloop` could be renamed to `cell_loop` to improve readability?

martinRenou: Yes

jtpio: :+1:

jtpio:
This looks good, thanks @martinRenou!

Let's wait for Travis before merging.

PR #470, opened 2019-11-07
Bundle mathjax with Voilà

SylvainCorlay:
This appears to be working - but we still have an issue with the mathjax fonts because of the ` redirect_to_file`  function which redirects to the `files` directory (in command-line mode) or to the `render` directory (in the server extension mode).


maartenbreddels:
I'm in favor of this approach btw, any indication for bundle size growth?

SylvainCorlay:
This is ready to go. @maartenbreddels @aschlaep.

aschlaep:
As a followup, maybe it makes sense to add a notebook to the `notebooks` directory with some LaTeX?

PR #471, opened 2019-11-07
Add cell error instruction configurable

aschlaep:
This adds a VoilaConfiguration option to allow customization of the debug instruction when there's a cell error.  It may not always make sense to say "Please run Voila with --debug ...", for example when voila is run as a notebook/server extension.  This'll allow people to supply their own instructions to the user for how to remediate potential bugs.

If anybody has any suggestions for a better name than `cell_error_instruction`, please let me know!

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -13,6 +13,8 @@ from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError, ExecutePreprocessor
 from nbformat.v4 import output_from_msg
 
+from traitlets import Unicode
+
 from ipykernel.jsonutil import json_clean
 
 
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
LGTM.

SylvainCorlay:
cc @maartenbreddels

maartenbreddels:
Good idea @aschlaep. I think however, the trait should go into `VoilaExecutePreprocessor`, does that make sense to you? If the `strip_notebook_errors` function is made into a method, it could directly access that property.

I'm happy to make those changes for you if you want.

SylvainCorlay:
@aschlaep could you push the changes that you just showed me so that I can try out the PR and the issue with default values.

aschlaep:
I added the default message in execute.py because the default value of the traitlet in configuration.py wasn't getting passed through for some reason...

voila/execute.py
```diff
@@ -226,6 +193,40 @@ def clear_output(self, outs, msg, cell_index):
...
+        if 'outputs' not in cell:
+            return cell
+
+        outputs = cell['outputs']
+
+        error_outputs = [output for output in outputs if output['output_type'] == 'error']
+
+        instruction = 'Please run Voila with --debug to see the error message.'
+        if 'VoilaConfiguration' in self.config and 'cell_error_instruction' in self.config['VoilaConfiguration']:
+            instruction = self.config['VoilaConfiguration']['cell_error_instruction']
```
maartenbreddels: This can be removed.

voila/execute.py
```diff
@@ -226,6 +193,40 @@ def clear_output(self, outs, msg, cell_index):
...
+
+        error_message = 'There was an error when executing cell [{}]. {}'.format(cell['execution_count'], instruction)
```
maartenbreddels:
```suggestion
        error_message = 'There was an error when executing cell [{}]. {}'.format(cell['execution_count'], this.cell_error_instruction)
```

maartenbreddels:
Thanks @aschlaep !

PR #474, opened 2019-11-11
More robust cell_type check

vidartf:
Make cell type check slightly more consistent across functions.


SylvainCorlay:
LGTM. The PR needs a rebase since we merged the other PR making `strip_code_cell_errors` a method instead of a free function.

vidartf:
@SylvainCorlay Should be ok now. Feel free to squash merge to clean commits up.

PR #403, opened 2019-10-07
feat: execute cells in a thread, this will unblock the server's event loop

maartenbreddels:
BREAKING refactor: use native coroutines instead of tornado's

This is a simpler version of #374

TODO: 
 * [x] drop py27 in travis, add py35
 * [x] handle exceptions in thread.

PR #475, opened 2019-11-12
Feature: start kernel async

maartenbreddels:
In #403 we had to start the kernel before we send the first bytes to the client, leading to a large time-to-first-byte TTFB (the HTML header containing script tags that the browser can already fetch). This is because we need to start the client in the main thread (something in 0mq is not thread safe I guess, causing the websocket connection to fail).

Here we start the kernel async from the main thread, and allow the jinja thread to access that.

It is a bit involved due to the py35 restriction.

I also moved the require/script tags up, so that the browser can fetch resources, even before the kernel is started.


SylvainCorlay:
Closing as superseded by #597

PR #481, opened 2019-11-19
fixes #468 - bugfixes for libraries that use other phosphor libs present in JupyterLab

### LINKED ISSUE #468: Missing phosphor libraries - 404
Looks like [this](https://github.com/voila-dashboards/voila/blob/5798a22e49ae50c451ea33098849571148ddfc1a/js/src/manager.js#L24) isnt enough for libraries that heavily use phosphor. I'm getting 404s on 

- `@phosphor/signaling`
- `@phosphor/virtualdom`
- `@phosphor/algorithm`
- `@phosphor/commands`
- `@phosphor/domutils`

timkpaine:
defines the other phosphor libraries expected to be present in jlab


maartenbreddels:
Note to self: check bundle size increase.

I think if not too big, I'm ok with this.

jtpio:
@timkpaine have you been using this in a custom template in the meantime?

timkpaine:
@jtpio in a fork, yep. its necessary for any jlab extension that uses phosphor, e.g. https://github.com/finos/perspective

timkpaine:
need to add `@jupyterlab/coreutils` as well.

timkpaine:
So based on my analysis:
- Any extension that uses phosphor needs the set of @phosphor extensions present
- Any extension that includes a mimerender will need the `@jupyterlab` extensions

Because these packages can't be imported standalone e.g. by `require`, and are present in JLab, extensions that use them will not work with `voila` until this PR is in. There may be more deps we need to include to enable other more esoteric extensions to run.

timkpaine:
@maartenbreddels bundle size shouldn't increase much since we already have `@phosphor/widgets`, which already has a dependency on the other stuff https://github.com/phosphorjs/phosphor/blob/master/packages/widgets/package.json#L45

the `@jupyterlab/` stuff would seem worth it even with a bundle size increase, since it enables arbitrary mimerenderers to continue to function

timkpaine:
interested in any thoughts on this, otherwise I'll need to fork soon for some of my PRs.

timkpaine:
failed on unrelated error

SylvainCorlay:
I am ok with this and am happy to merge now for integration in 0.2.

ping @jtpio

timkpaine:
@SylvainCorlay thanks! The side effect of this merge is that any libraries that also have rendermimes will work out of the box (e.g. see the CC'd Plotly issue).

SylvainCorlay:
This is great. We are trying to push towards 0.2 but if we cannot, I may backport it into a 0.1.x.

timkpaine:
@SylvainCorlay the only question is the phosphor/lumino name change, it might be enough to just import as lumino but define on the window as both the lumino and phosphor name (e.g. just copy paste those lines and s/phosphor/lumino).

jtpio:
Indeed, aliases for lumino might be enough. Or something similar to the way phosphor / lumino is handled in ipywidgets 7.

PR #484, opened 2019-11-24
Pass Papermill parameters from request through to notebook

timkpaine:
This PR does the following:
- tries to get the `parameters` argument from either the query parameter or body of the request
- if `parameters` is passed and it is a dict, it calls `papermill.parameterize.parameterize_notebook` on the underlying notebook, prior to execution

### DIFF
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -328,3 +328,11 @@ To hide both the code cell and the output cell (if any) for every cell that has 
 
 You can use any tag you want but be sure to use the same tag name in the Voilà command.
 And please note that this functionality will only hide the cells in Voilà but will not prevent them from being executed.
+
+
+Integrate with Papermill for Parameterized Applications
+=======================================================
+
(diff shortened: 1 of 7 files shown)


timkpaine:
tmp

timkpaine:
@vidartf can we inspect notebooks for their parameters and make a UI for injecting them prior to launch in `phoila`

timkpaine:
@SylvainCorlay @maartenbreddels @jtpio thoughts on this? it seems like a natural thing to do, though the implementation could be different

jtpio:
Thanks @timkpaine.

That sounds like a natural thing to do indeed, and would definitely be useful.

The main question is probably whether we want voila to depend on papermill. There was some similar discussion in JupyterLab about this (papermill vs notebook standard): https://github.com/jupyterlab/jupyterlab/pull/8410

timkpaine:
@jtpio prelaunch hook + example showing how to use papermill would accomplish the same functionality without the dependency

timkpaine:
closed by https://github.com/voila-dashboards/voila/pull/724

PR #493, opened 2019-12-04
Use Python 3.5 in GitHub Actions workflow

jtpio:
New changes on master since https://github.com/voila-dashboards/voila/pull/445 was open, for example https://github.com/voila-dashboards/voila/pull/403 which drops Python 2.7.

We can update the GitHub Actions workflow to reflect that change too, since it's failing on master: https://github.com/voila-dashboards/voila/runs/333012981


maartenbreddels:
Thanks!

PR #501, opened 2019-12-13
fix: the default title in the template is always Notebook

mariobuikhuizen:
The property used as a fallback for the title of the template is
not initialized by Voila. Following nbconvert, in which this
property is set to the filename of the notebook [1], this commit
now does the same.

Related to: voila-dashboards/voila-vuetify#25

[1] https://github.com/jupyter/nbconvert/blob/6c6e5fa0c757169017fd325fad98efa8d4a368ea/nbconvert/exporters/exporter.py#L172

### DIFF
--- a/tests/app/execute_test.py
+++ b/tests/app/execute_test.py
@@ -19,7 +19,7 @@ def test_hello_world(http_client, base_url):
     assert response.code == 200
     html_text = response.body.decode('utf-8')
     assert 'Hi Voila' in html_text
-    assert 'print' not in html_text, 'by default the source code should be stripped'
+    assert 'print(' not in html_text, 'by default the source code should be stripped'
     assert 'test_template.css' not in html_text, "test_template should not be the default"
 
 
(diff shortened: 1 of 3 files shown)


timkpaine:
This means you can no longer override the name?

mariobuikhuizen:
If title is set in the notebook metadata, that is used.

`{% set nb_title = nb.metadata.get('title', '') or resources['metadata']['name'] %}`

timkpaine:
I think we were setting it in resources, which this now clobbers

maartenbreddels:
This breaks our tests, I'll fix that after the weekend.

timkpaine:
Yeah so extra resources should take precedence still https://github.com/voila-dashboards/voila/blob/master/voila/handler.py#L70

So this is backwards compatible

jtpio:
This looks good.

@maartenbreddels did you want to take another look? or can this be merged?

PR #530, opened 2020-01-28
Fix: Respect allow_errors value

### LINKED ISSUE #519: VoilaExecutePreprocessor.allow_errors always True
If a cell throws an exception, the cells after it are still being executed.  Setting allow_errors explicitly to False (which is the default anyway) doesn't affect it.

martinRenou:
This will fix #519.

It will respect the value of the `allow_errors` option.

If `allow_errors` is `False` (default), the execution will stop on the first error, and the traceback of the last error will be visible only if Voila is run with `--debug`.

If `allow_errors` is `True`, the execution does not stop on errors, and tracebacks are visible only if Voila is run with `--debug`.

### DIFF
--- a/tests/app/notebooks_test.py
+++ b/tests/app/notebooks_test.py
@@ -1,4 +1,6 @@
 # simply tests if some notebooks execute
+import sys
+
 import pytest
 
 
@@ -8,6 +10,7 @@ def voila_args(notebook_directory, voila_args_extra):
 
 
 @pytest.mark.gen_test
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="Matplotlib installation issue on MacOS, Python 3.5")
 def test_other_comms(http_client, base_url):
     response = yield http_client.fetch(base_url + '/voila/render/other_comms.ipynb')
     html_text = response.body.decode('utf-8')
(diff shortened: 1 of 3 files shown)


voila/handler.py
```diff
@@ -136,7 +142,10 @@ def _jinja_cell_generator(self, nb, kernel_id):
 
         with ep.setup_preprocessor(nb, resources, km=km):
             for cell_idx, cell in enumerate(nb.cells):
-                res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                try:
+                    res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                except:
+                    raise tornado.web.HTTPError(500, 'Execution error in cell {}'.format(cell_idx + 1))
```
martinRenou: Unfortunately, this does not show the 500 page. The page is stuck on the spinner.

tests/app/notebooks_test.py
```diff
@@ -8,6 +10,7 @@ def voila_args(notebook_directory, voila_args_extra):
 
 
 @pytest.mark.gen_test
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="Matplotlib installation issue on MacOS, Python 3.5")
```
SylvainCorlay, line 13: I think we can remove that since we now require 3.6.

SylvainCorlay:
This needs a rebase!

Also, it seems that the change to notebooks_test.py is unnecessary.

Issue #516, opened 2020-01-16
Voila cannot show Greek letter (latex format) correctly


uu1477:
When I marked widgets with description, it is working well in JupyterLab. 

![Screenshot from 2020-01-16 15-16-18](https://user-images.githubusercontent.com/20816399/72532226-4e0b9d80-3873-11ea-8ed5-83acecdbf741.png)

However, when I use Voila, it cannot show the Greek letter correctly. 
![Screenshot from 2020-01-16 15-16-29](https://user-images.githubusercontent.com/20816399/72532267-62e83100-3873-11ea-9b43-1d7ad37f7218.png)

I am using the least Voila version, which compiled from this Git repository.

jtpio:
Thanks @uu1477.

That looks related to https://github.com/voila-dashboards/voila/pull/470 and https://github.com/voila-dashboards/voila/pull/489.

maartenbreddels:
I think this is a regression, only markdown cells render to math.

maartenbreddels:
Correction: this does seem to work in master (browser cache made it not work). Could you try opening the dev console and disable your cache?

jtpio:
The `basics.ipynb` notebook still seems to be an issue on master.

![image](https://user-images.githubusercontent.com/591645/73257712-8c368480-41c4-11ea-8021-dbd55a4450e7.png)

jtpio:
But adding a cell below with the following content seems to load the font, which applies to the slider description above:

```python
x = widgets.Checkbox(value=True, description='$\Psi$ Real')
x
```

![image](https://user-images.githubusercontent.com/591645/73258374-d9ffbc80-41c5-11ea-96c5-38a22c3a52c3.png)

jgunstone:
adding here because I'm guessing it could be related... 

Math not display in Voila. 


```py
from IPython.display import Math, display, clear_output
import ipywidgets as w

out = w.Output()
with out: 
    clear_output()
    display(Math('\\frac{\\rm{kg}}{\\rm{m}^{2} \\cdot \\rm{s}^{2}}'))
    
out
```
![image](https://user-images.githubusercontent.com/21370980/182820257-fce430c6-f5a4-45df-9cb7-59b3e99ea7fa.png)


## env 

ubuntu2004

```bash
(aectemplater-dev) jovyan@PORT4028:/mnt/c/engDev/git_mf/tmp/aectemplater/ui$ conda list jupyter
[...]

PR #531, opened 2020-01-28
fix: mock v2 api of MathJax

maartenbreddels:
This is a workaround or fix for #516

There are a few problems:
 * We wait for the widgets to be build before we render mathjax, however, if widgets are inside a VBox, they may be rendered async, so they come after mathjax is done.
 * typeset in https://github.com/jupyter-widgets/ipywidgets/blob/b3708e1305248b132999322266aab07c7e9ec3dc/packages/controls/src/utils.ts#L16 (which is called by description widget in https://github.com/jupyter-widgets/ipywidgets/blob/b3708e1305248b132999322266aab07c7e9ec3dc/packages/controls/src/widget_description.ts#L62) uses the v2 API of MathJax, while we use v3 since #470 (since voila 0.1.18)

In this PR I mock the v2 API a bit, and rerender all math. I am not sure we should do this.

cc @jasongrout 
We could go mathjax v3 on all platforms, notebook, lab, widgets, or revert to v2 mathjax in voila.

### DIFF
--- a/js/src/mathjax.js
+++ b/js/src/mathjax.js
@@ -48,4 +48,16 @@ export function renderMathJax() {
     .typeset()
     .updateDocument()
     .reset();
+  window.MathJax = {
+    Hub: {
+      Queue: ([_ignore, _ignore2, node]) => {
+        html.findMath({ elements: [node] })
+            .compile()
+            .getMetrics()
+            .typeset()
+            .updateDocument()
+            .reset()
+      }
+    }
+  }
 };


jasongrout:
The JLab issue for upgrading to MathJax 3 is https://github.com/jupyterlab/jupyterlab/issues/7218

The non-core extension to upgrade to MathJax 3 is https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/mathjax3-extension

maartenbreddels:
Great, that example is useful, it wasn't easy figuring out the API.

maartenbreddels:
Might be good to take a look at some of the commits in https://github.com/widgetti/solara/pull/155 where I continued this work.

PR #535, opened 2020-01-30
fix: throttle the rate AND number of request_state msg'es from the frontend

### LINKED ISSUE #534: Many widgets fail to render
cc @aschlaep 
Eg.
```python
import ipywidgets as widgets
widgets.Widget.close_all()
def recursive(n):
    if n == 0:
        return widgets.Button(description='Hi')
    else:
        cls = (widgets.VBox if n % 2 else widgets.HBox)
        return cls([recursive(n-1), recursive(n-1)])
display(recursive(9))
display(len(widgets.Widget.widgets))
```
Will often fail to render in voila. I think it's likely to be a timing/order issue in the manager (message received before an event handler is attached), or packets being dropped (iorate limit?)

maartenbreddels:
Fixes #534

Related: https://github.com/jupyter-widgets/ipywidgets/pull/2765

@jasongrout this is a variant of your mapBatch that will try to put 'room' number of messages in flight, so once 1 message comes in (and the rate is not too high), the next message will be directly sent. This should decrease latency.

Let me know what you think, we probably want this functionality to go into jupyter-widgets.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -27,6 +27,7 @@
     "css-loader": "^3.0.0",
     "file-loader": "^4.0.0",
     "npm-run-all": "^4.1.5",
+    "promise-semaphore": "^0.2.8",
     "style-loader": "^0.23.1",
     "url-loader": "^1.0.0",
     "webpack": "^4.29.3",
--- a/js/package.json
+++ b/js/package.json
@@ -27,7 +27,7 @@
     "css-loader": "^3.0.0",
     "file-loader": "^4.0.0",
     "npm-run-all": "^4.1.5",
-    "promise-semaphore": "^0.2.8",
+    "p-limit": "^2.2.2",
     "style-loader": "^0.23.1",
     "url-loader": "^1.0.0",
     "webpack": "^4.29.3",
--- a/js/src/manager.js
+++ b/js/src/manager.js
@@ -16,6 +16,7 @@ import * as controls from '@jupyter-widgets/controls';
 import * as PhosphorWidget from '@phosphor/widgets';
 
 import { requireLoader } from './loader';
+import { batchRateMap } from './utils';
 
 if (typeof window !== "undefined" && typeof window.define !== "undefined") {
     window.define("@jupyter-widgets/base", base);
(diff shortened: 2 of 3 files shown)


js/package.json
```diff
@@ -27,6 +27,7 @@
     "css-loader": "^3.0.0",
     "file-loader": "^4.0.0",
     "npm-run-all": "^4.1.5",
+    "promise-semaphore": "^0.2.8",
```
jasongrout: This package is still using bluebird promises and hasn't been updated in 3 years.

maartenbreddels: Yes, I'm not happy with that package either, looking for something else.  https://github.com/sindresorhus/p-queue may actually do all of this..

maartenbreddels:
https://www.npmjs.com/package/p-limit is a bit easier to understand than p-queue.

maartenbreddels:
I thought about it a bit more, my argument was that my method would be faster with latency. But with 2000 widgets, at 100 per batch, you get an extra 20x the latency, which could be 2 seconds? On a total time of maybe 20 seconds?

So I think in terms of performance, we don't gain much actually.

maartenbreddels:
Ok, apart from the performance, we may still hit the iopub rate limit without this p-limit.

jasongrout:
What about this implementation of concurrent batching?

```typescript
/**
 * A batched map implementation with concurrency
 *
 * Take a list, a batch size, a mapping function, and a max concurrency
 */
export async function mapBatch(
  list: any[],
  batchSize: number,
  concurrency: number,
  fn: Function,
  thisArg: any
): Promise<any[]> {
  const results = new Array(list.length);

  const chunks: PromiseClosure[] = [];

  // Process a single chunk and resolve to the next chunk if available
  async function processChunk(chunk: any[], start: number): Promise<void> {
    const chunkResult = await Promise.all(
      chunk.map((v, i) => fn.call(thisArg, v, start + i, list))
    );

    // Splice the chunk results into the results array. We use results.length
    // because the last chunk may not be a full batchSize.
    results.splice(start, chunkResult.length, ...chunkResult);

    // Start the next work item by returning its promise
    if (chunks.length > 0) {
      return chunks.shift()!();
    }
  }

  // Make closures for each batch of work
  for (let i = 0; i < list.length; i += batchSize) {
    chunks.push(() => processChunk(list.slice(i, i + batchSize), i));
  }

  // Start the concurrent chunks. Each chunk will automatically start
  // the next available chunk when it finishes.
  await Promise.all(chunks.splice(0, concurrency).map(f => f()));
  return results;
}
```

jasongrout:
> What about this implementation of concurrent batching?

See the cleaned-up implementation at https://github.com/jupyter-widgets/ipywidgets/pull/2765

jasongrout:
Actually, I don't think we need throttling. Having a concurrency of, say, 100 will naturally introduce the appropriate amount of throttling, i.e., the network roundtrip. Since all we care about is not hitting the zmq socket with more than 100 requests or so simultaneously, concurrency naturally throttles to about that.

jasongrout:
> Actually, I don't think we need throttling.

Ah, the throttling is for the separate iopub rate limiting. Yes, good point.

SylvainCorlay:
Merging this!

PR #536, opened 2020-02-03
(backport 0.1.18) fix: throttle the rate AND number of request_state msg'es from the frontend

### LINKED ISSUE #534: Many widgets fail to render

maartenbreddels:
Fixes #534, backport of #535

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -6750,9 +6750,9 @@
       "dev": true
     },
     "p-limit": {
-      "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.2.0.tgz",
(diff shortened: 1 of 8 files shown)


maartenbreddels:
The failure in` test_execute_output` I think is unrelated, I've seen it before:
```
E               AssertionError: assert {'output_type...text': u'hi4'} == {u'output_type...xt': u'hi4\n'}
E                 Omitting 2 identical items, use -vv to show
E                 Differing items:
E                 {'text': 'hi4'} != {'text': 'hi4\n'}
E 
```

PR #541, opened 2020-02-11
Async cell execution

davidbrochart:
This uses the asynchronous cell execution mode of `nbclient`, which requires an async client from `jupyter_client`. Also, `nbconvert` has to enable async in Jinja.


davidbrochart:
This can't work with Python 3.5 because of Jinja async support (see https://github.com/pallets/jinja/issues/1058).

maartenbreddels:
Running `$ time py.test tests/app/` before and after this PR 
```
py.test tests/app/  21.50s user 5.80s system 51% cpu 53.168 total
py.test tests/app/  16.38s user 4.25s system 87% cpu 23.649 total
```
So less time, and more cpu usage 🚀
(I did rebase on master, since it has an extra test)

maartenbreddels:
I cannot seem to break it yet, if you rebase on master, you'll also get the `tests/app/many_iopub_messages_test.py` test, which is an important one to pass, since it tests one of the issues of #534, but it passes locally, great work!

davidbrochart:
Trigger CI.

jf---:
can't wait for this to land -- both significant performance gains at the "cost" of considerably cleaner code --- duly impressive

davidbrochart:
This is ready to be merged.

davidbrochart:
Trigger CI.

davidbrochart:
Trigger CI.

davidbrochart:
The Jupyter server app has to use [an AsyncMappingKernelManager](https://github.com/voila-dashboards/voila/pull/541/files#diff-abf50f62fd8b49c6e772c65785cbc58fR26). It that is acceptable, this PR is ready to be merged.

share/jupyter/voila/templates/default/nbconvert_templates/lab.tpl
```diff
@@ -1,4 +1,4 @@
-{%- extends 'display_priority.tpl' -%}
```
maartenbreddels, line 1: With https://github.com/jupyter/nbconvert/pull/1173 we could again use the tpl, if we care about backwards compatibility

setup.py
```diff
@@ -376,17 +376,19 @@ def get_data_files():
...
         'pygments>=2.4.1,<3'  # Explicitly requiring pygments which is a second-order dependency.
                               # An older versions is generally installed already and is otherwise not updated by pip.
     ],
     'extras_require': {
         'test': [
             'mock',
-            'pytest<4',
-            'pytest-tornado',
+            'pytest',
+            'pytest-tornasync',
```
maartenbreddels, line 391: cool, but i wonder if we should follow jupyter-server by doing with pytest-asyncio.

voila/handler.py
```diff
@@ -118,12 +113,29 @@ async def get(self, path=None):
     def redirect_to_file(self, path):
         self.redirect(url_path_join(self.base_url, 'voila', 'files', path))
 
-    @tornado.gen.coroutine
-    def _jinja_kernel_start(self):
+    async def _jinja_kernel_start(self, nb):
         assert not self.kernel_started, "kernel was already started"
-        # See command above aboout not being able to start the kernel from a different thread
+        kernel_id = await self.kernel_manager.start_kernel(kernel_name=self.notebook.metadata.kernelspec.name, path=self.cwd)
```
maartenbreddels, line 118:
```suggestion
        kernel_id = await self.kernel_manager.start_kernel(kernel_name=nb.metadata.kernelspec.name, path=self.cwd)
```

maartenbreddels: or insert `assert nb is self.notebook` to make this explicit.

voila/handler.py
```diff
@@ -118,12 +113,29 @@ async def get(self, path=None):
...
+        # When Voila is launched as an app, its kernel manager's type is AsyncMappingKernelManager, and thus
+        # its kernel client's type is AsyncKernelClient.
+        # But this has to be explicitly configured when launched as a server extension, with e.g.:
+        # --ServerApp.kernel_manager_class=jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager
+        # If it's not done, the kernel manager might not be async, which is not a big deal, but we want the kernel
+        # client to be async, so we explicitly configure it for this particular case:
+        km.client_class = 'jupyter_client.asynchronous.AsyncKernelClient'
+        self.executor = VoilaExecutor(nb, km=km, config=self.traitlet_config)
+        self.executor.kc = km.client()
+        self.executor.kc.start_channels()
```
maartenbreddels, line 129: I am/was a bit worried about this. I remember when I just started adding an async client that the jupyter_server part used the same client. Maybe that was indeed because I configured it to be async.
Are you saying that both jupyter_server and voila have their own client to the kernel, jupyter_server does whatever it does, and we do it async?
Did you check if https://github.com/voila-dashboards/voila/blob/master/notebooks/basics.ipynb works with both the voila app and server extension?

.github/workflows/build.yml
```diff
@@ -15,7 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
+    - name: install dependencies
+      run: |
+        sudo apt update
+        sudo apt-get install -y --no-install-recommends libgbm-dev
```
maartenbreddels, line 22: Could you add a comment why we need this?

share/jupyter/voila/templates/default/nbconvert_templates/lab.tpl
davidbrochart: I reverted to `.tpl`.

setup.py
davidbrochart: jupyter-server uses pytest-tornado too: https://github.com/jupyter/jupyter_server/blob/4f2d91d1754a30c9e46b48638c5a67907729fcb6/setup.py#L101

voila/handler.py
davidbrochart: Actually `nb` is not needed in `_jinja_kernel_start`, I removed it.

voila/handler.py
davidbrochart: Yes it works both with the voila app and the server extension. I'm not sure I understand what you mean, Voila has its own kernel anyway, right?

.github/workflows/build.yml
davidbrochart: I'm adding the GH issue where the fix comes from.

voila/handler.py
maartenbreddels: Well, the kernel is also known to jupyter server, which is responsible for the WebSocket connection. So jupyter server had a connection to the kernel as well (actually it may be in parallel to us)

.github/workflows/main.yml
```diff
@@ -34,25 +34,25 @@ jobs:
...
         conda update -y -n base conda setuptools
-    
+
     - name: Init conda
       run: |
         conda init bash
         conda info -a
 
     - name: Create the conda environment
-      run: conda create -q -y -n voila-tests -c conda-forge python=$PYTHON_VERSION pip jupyterlab_pygments==0.1.0 nbconvert=5.5 pytest-cov nodejs flake8 ipywidgets matplotlib xeus-cling
```
SylvainCorlay, line 44: Don't we still depend on nbconvert for the html rendering?

setup.py
```diff
@@ -376,17 +376,19 @@ def get_data_files():
     },
     'install_requires': [
         'async_generator',
-        'jupyter_server>=0.1.0,<0.2.0',
-        'nbconvert>=5.5.0,<6',
+        'jupyter_server>=0.3.0',
+        'jupyter_client>=6.1.3',
+        'nbclient>=0.2.0',
+        'nbconvert==6.0.0a1',
```
SylvainCorlay, line 382: Do we really need 6.0.0 or is this merely for testing?

voila/execute.py
```diff
@@ -149,11 +148,11 @@ def preprocess(self, nb, resources, km=None):
...
         return result
 
-    def preprocess_cell(self, cell, resources, cell_index, store_history=True):
+    async def execute_cell(self, cell, resources, cell_index, store_history=True):
         try:
             # TODO: pass store_history as a 5th argument when we can require nbconver >=5.6.1
-            # result = super(VoilaExecutePreprocessor, self).preprocess_cell(cell, resources, cell_index, store_history)
-            result = super(VoilaExecutePreprocessor, self).preprocess_cell(cell, resources, cell_index)
+            # result = super(VoilaExecutor, self).execute_cell(cell, resources, cell_index, store_history)
+            result = await super(VoilaExecutor, self).async_execute_cell(cell, cell_index)
```
SylvainCorlay, line 155: should we pass store_history to `async_execute_cell`?

.github/workflows/main.yml
davidbrochart: It's in the `setup.py` already.

setup.py
davidbrochart: We need nbconvert `6.0.0.a1` for async jinja template execution.

SylvainCorlay: Thanks

.github/workflows/main.yml
SylvainCorlay: Yes, I realized tha after.

voila/exporter.py
```diff
@@ -106,24 +99,12 @@ async def generate_from_notebook_node(self, nb, resources=None, extra_context={}
...
-                await yield_((output, resources))
-
-        async for output, resources in async_jinja_generator():
-            await yield_((output, resources))
+        async for output in self.template.generate_async(nb=nb_copy, resources=resources, **extra_context):
+            yield (output, resources)
 
     @property
     def environment(self):
+        self.enable_async = True
```
SylvainCorlay, line 107: Maybe add a comment to say what this is?

voila/handler.py
SylvainCorlay: could we use nbclient's async context manager `async with self.async_setup_kernel()` instead of manually doing

```
        self.executor.kc = km.client()
        self.executor.kc.start_channels()
```

SylvainCorlay:
Weird, there is a test failure over a new line character on Python 3.8 - and also I cannot restart the build.

davidbrochart:
Trigger CI.

SylvainCorlay:
Why Can't I restart CI from the travis UI?

davidbrochart:
No idea.

SylvainCorlay:
:tada:

PR #542, opened 2020-02-14
Add performance test

davidbrochart:
This test launches 10 clients, all executing the same notebook. The notebook consists of 100 cells, each cell sleeps for 0.1 second. We check:
- the overall time execution of all notebooks, taking into account the kernel launch time (0.5 s) and the notebook execution time (100 * 0.1 = 10 s), with a margin of 100% on the notebook execution time (`TimeoutError` if exceeded).
- the mean cell execution time (0.1 s) with a 50% margin (`exitcode` 1 if exceeded).


tests/app/performance_test.py
```diff
@@ -0,0 +1,101 @@
+import subprocess
+import time
+import os
+import shutil
+import sys
+
+
+def get_cell(seconds):
+    cell = ', { "cell_type": "code", "execution_count": null, "metadata": {}, "outputs": [], ' \
```
jtpio: Wondering whether these cells could be created with `nbformat`?

davidbrochart: It seems a bit overkill to me.

jtpio: Wouldn't this help improve readability?

davidbrochart: It is true that it is not very readable. But do you mean using `nbformat` or Jinja?

jtpio: With `nbformat`, something similar to Voila's support for script files:

https://github.com/voila-dashboards/voila/blob/8557d9d270100e55a40911e1afe61046736eba18/voila/handler.py#L195-L204

And then `nbformat.write` to save the notebook file.

davidbrochart: I didn't know about that, I will look at it. Thanks!

davidbrochart:
There might be a temporary conda issue on OSX, but other than that this is ready to be reviewed @jtpio.

SylvainCorlay:
This looks pretty good. I am happy to get this in if you think this is ready @davidbrochart.

I think we can probably set up ASV after this.

davidbrochart:
Should it be only in ASV? The thing is that we don't know which machine we get in the CI, so if the machine is very slow the test can fail, or we should somehow take into account the machine's CPU frequency in the test.

PR #544, opened 2020-02-25
Show timeout error in cell output

### LINKED ISSUE #538: Execution timeout doesn't make it to the UI
If cell execution times out, you get the `Timeout waiting for execute reply (30s)` message in the console, but the UI keeps spinning on the "Executing m of n" message.

davidbrochart:
When a cell execution times out, the page hangs with the spinner. There is a timeout exception in the terminal but the user has no clue as to what it happening.
The notebook execution should be aborted and a timeout error should be printed in the cell output.
Fixes #538

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -21,11 +21,6 @@ from nbconvert.preprocessors import ClearOutputPreprocessor
 from .execute import executenb, VoilaExecutePreprocessor
 from .exporter import VoilaExporter
 
-try:
-    TimeoutError  # Py 3
-except NameError:
-    TimeoutError = RuntimeError  # Py 2
-
 
 class VoilaHandler(JupyterHandler):
 
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -21,6 +21,11 @@ from nbconvert.preprocessors import ClearOutputPreprocessor
 from .execute import executenb, VoilaExecutePreprocessor
 from .exporter import VoilaExporter
 
+try:
+    TimeoutError  # Py 3
+except NameError:
+    TimeoutError = RuntimeError  # Py 2
+
 
 class VoilaHandler(JupyterHandler):
 
@@ -140,9 +145,16 @@ class VoilaHandler(JupyterHandler):
         nb, resources = ClearOutputPreprocessor().preprocess(nb, {'metadata': {'path': self.cwd}})
         ep = VoilaExecutePreprocessor(config=self.traitlet_config)
 
+        stop_execution = False
         with ep.setup_preprocessor(nb, resources, km=km):
             for cell_idx, cell in enumerate(nb.cells):
-                res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                if stop_execution:
+                    break
+                try:
+                    res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                except TimeoutError:
+                    res = (cell, resources)
+                    stop_execution = True
 
                 yield res[0]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,14 +26,12 @@ jobs:
     - name: Add conda to $PATH
       run: echo ::add-path::$CONDA/condabin
 
-    - name: Update conda on Mac
+    - name: Fix permissions on Mac
       if: matrix.os == 'macos-latest'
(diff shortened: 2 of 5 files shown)


voila/handler.py
```diff
@@ -21,6 +21,11 @@
 from .execute import executenb, VoilaExecutePreprocessor
 from .exporter import VoilaExporter
 
+try:
+    TimeoutError  # Py 3
+except NameError:
+    TimeoutError = RuntimeError  # Py 2
```
jtpio: Do we still need this?

jtpio:
Thanks @davidbrochart, this looks good!

voila/handler.py
davidbrochart: Probably not, I will do some cleaning of python2 code, thanks.

jtpio: :+1:

davidbrochart:
Not sure why the Voila Preview JupyterLab extension doesn't install.

jtpio:
It's because of https://github.com/voila-dashboards/voila/pull/554 (if this one was rebased on master).

davidbrochart:
It should be ready to merge now.

maartenbreddels:
What I notice, running just the test, is that it takes ~1 second to run it (the green dot appears after 1 second), but it takes then 4-5 seconds to finish running pytest. Is this a cleanup phase that hangs? Can you confirm this? I don't see this happening to execute_test.py for instance.

voila/handler.py
```diff
@@ -140,9 +140,16 @@ def _jinja_cell_generator(self, nb, kernel_id):
...
         with ep.setup_preprocessor(nb, resources, km=km):
             for cell_idx, cell in enumerate(nb.cells):
-                res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                if stop_execution:
+                    break
+                try:
+                    res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
+                except TimeoutError:
+                    res = (cell, resources)
+                    stop_execution = True
```
maartenbreddels: What about
```suggestion
            for cell_idx, cell in enumerate(nb.cells):
                try:
                    res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
                except TimeoutError:
                    res = (cell, resources)
                    stop_execution = True
                    return
                finally:
                    yield res[0]
```

davidbrochart: Would that mean we don't `yield res[0]` for the cell where we stop the execution?

davidbrochart:
> What I notice, running just the test, is that it takes ~1 second to run it (the green dot appears after 1 second), but it takes then 4-5 seconds to finish running pytest. Is this a cleanup phase that hangs? Can you confirm this? I don't see this happening to execute_test.py for instance.

It is because the test sleeps for 10s. Actually there is another 5s timeout somewhere, maybe for shutting down the kernel. We could add `--VoilaExecutePreprocessor.interrupt_on_timeout=True` to the test, but then it should catch the `KeyboardInterrupt`. What do you think?

voila/handler.py
maartenbreddels: finally *always* run, otherwise, the test would have picked that up right?

maartenbreddels:
Yes, you are right, it is for shutting down the kernel, to confirm I ran: `$ time py.test tests/app/timeout_test.py --fulltrace` and then ctrl-c after the green dot appears (so the test passed). The stacktrace shows it was in `jupyter_client/manager.py:finish_shutdown`.
What about:
```patch
diff --git a/tests/app/timeout_test.py b/tests/app/timeout_test.py
index 9da1876..df3c37f 100644
--- a/tests/app/timeout_test.py
+++ b/tests/app/timeout_test.py
@@ -10,7 +10,7 @@ def voila_notebook(notebook_directory):
 
 @pytest.fixture
 def voila_args_extra():
-    return ['--VoilaExecutePreprocessor.timeout=1']
+    return ['--VoilaExecutePreprocessor.timeout=1', '--KernelManager.shutdown_wait_time=0.1']
 
 ```

davidbrochart:
Sounds good, thanks for the suggestion.

davidbrochart:
There was a directory permission issue on Mac. It should be good to go now.

maartenbreddels:
Awesome work, thanks!

PR #546, opened 2020-02-28
Throttling forwardport

SylvainCorlay:
This is a forward port of the remaining commits of #536. 

@maartenbreddels will self merge in a few. Please let me know if you have any comment.


maartenbreddels:
I think this is good to go.

PR #571, opened 2020-04-08
Fix xleaflet notebook example

davidbrochart:
Draw control doesn't seem to work in xleaflet.


davidbrochart:
Trigger CI.

SylvainCorlay:
I'd rather fix the xleaflet draw control than remove it from the example?

Issue #526, opened 2020-01-27
Run tests on Windows


jtpio:
Tests should eventually also run on Windows, now that GitHub Actions have been enabled and used (https://github.com/voila-dashboards/voila/pull/445).

PR #573, opened 2020-04-09
Simplify actions

martinRenou:


SylvainCorlay:
Now that we can drop 3.5, rebasing should get the tests to pass.

@martinRenou

jtpio:
Linking to the conversation on gitter: https://gitter.im/QuantStack/Lobby?at=5ee247be36589658b12b9026

We might want to try with `setup-miniconda` action, and even enable `mamba`.

It might even help fixing #526 too.

jtpio:
wow thanks @martinRenou for the update. Let's see how it goes!

martinRenou:
I'll enable Windows too, see what happens

jtpio:
Windows used to be an issue because of xeus-cling (but there could be a condition to handle that)

martinRenou:
Looks like `python` refers to the `/usr/bin` one, not the conda installed one.

jtpio:
The other jobs (ubuntu and macos) are passing, nice :+1: 

We could also fix windows separately.

martinRenou:
Yeah although there is still one failing on Macos. I'll look at it later.

martinRenou:
I rerun the test and it's now passing. There might be a race condition in the tests.

Let's disable Windows tests for now then.

martinRenou:
@jtpio All green. Ok to merge?

jtpio:
That looks good and much simpler, thanks a lot!

Happy to go for it if we're fine using mamba on CI.

PR #574, opened 2020-04-09
test: add render test with pyppeteer - chromium based

maartenbreddels:
Using pyppeteer (a python port of puppeteer) we let chrome render the page, make a screenshot, and compare it.


maartenbreddels:
It silently fails with this in main.yml:
```

    - name: Upload artifacts (from test failures)
        uses: actions/upload-artifact@v1
        if: succeeded() || failed()
        with:
          name: artifacts
          path: artifacts
```

maartenbreddels:
sometimes github just silently ignores incorrect yaml.

maartenbreddels:
As mentioned https://github.com/puppeteer/puppeteer/issues/661 it seems to be important to disable gpu rendering to avoid text rendering differences. Note that the pixel values differed by an absolute value of 1, so it's important to understand the difference, but we should probably expect more platform rendering differences. 
We might want to have a tolerance of say 0.1 (10%) for passing tests?

maartenbreddels:
Finding significant differences between the OS's:
Linux:
print_truth
OSX:
print_testrun
(Note the H and l)

But these should pass when we allow a relative pixel difference of 10%.

PR #579, opened 2020-04-16
fix: prevent scaling down the viewport on mobile devices

mariobuikhuizen:
This way the right breakpoint for the screen size on mobile devices
will be used.

With the use of a breakpoint at 960px for showing the two elements side by side

Without this PR:
Schermafbeelding 2020-04-16 om 17 05 50

With this PR:
Schermafbeelding 2020-04-16 om 17 06 05

### DIFF
--- a/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
+++ b/share/jupyter/voila/templates/default/nbconvert_templates/base.tpl
@@ -6,6 +6,7 @@
 <head>
 {%- block html_head -%}
 <meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
 {% set nb_title = nb.metadata.get('title', '') or resources['metadata']['name'] %}
 <title>Voila: {{nb_title}}</title>
 


jtpio:
Thanks @mariobuikhuizen, this looks like a good default to have :+1:

maartenbreddels:
I had to read a bit on this to understand. It seems that by default, browsers on small screens lie about the screensize, e.g. say they are 960px wide, render it at that size, but show it zoomed in/out. This was originally to support non-small-screen-optimized (non-responsive) websites. This PR will make them no-lie, letting them report the very small screensizes.

I am afraid a lot of voila notebooks would also not be optimized for small screens, and the current default would make more sense. However, I can see the use of this for more advanced usage.

I wonder if we could put this behind a flag, similar to --theme? Or should we have a different template for it?

jtpio:
Then maybe we could do side-by-side comparisons with "complex" voila dashboards?

It sounded like it would have been useful to have this same default across Jupyter frontends:

- classic notebook: https://github.com/jupyter/notebook/blob/master/notebook/templates/page.html#L12
- JupyterLab: https://github.com/jupyterlab/jupyterlab/blob/b0a1ca62628a0b450ecec483c2f7aca6457d3b32/jupyterlab/staging/templates/template.html#L6

maartenbreddels:
Ah, well, if jupyter lab and notebook have it, we should follow. But I guess this should go upstream to nbconvert. I'll make a PR for that.

jtpio:
Fixed by https://github.com/jupyter/nbconvert/pull/1322.

Thanks @mariobuikhuizen and @maartenbreddels!

Issue #587, opened 2020-04-30
Performance when running a "heavy" cell


thvasilo:
I'm trying to use Voila in combination with torchtext for some NLP visualizations.

At one point I use torchtext to load a dataset, that takes a few seconds, maybe 10 seconds to load the data, then 10-20sec to create a vocabulary.

What I'm seeing when I launch the notebook through the `voila` command is a `TimeoutError: Cell execution timed out`.

I have two questions then: What is causing the cell to take much longer to run within Voila vs. when I'm executing it in Jupyter, and if there's a way to increase the timeout so long-running cells have time to complete.

davidbrochart:
You can disable the timeout with `--VoilaExecutor.timeout=None` (see `voila --help-all`).
There is no reason for a cell to take more time to execute in Voila than in a notebook.
```
--VoilaExecutor.timeout=<Int>
    Default: None
    The time to wait (in seconds) for output from executions. If a cell
    execution takes longer, a TimeoutError is raised.
    `None` or `-1` will disable the timeout. If `timeout_func` is set, it
    overrides `timeout`.
```

thvasilo:
Hello @davidbrochart 

Thanks, couldn't find that option in the version I'm using (0.1.21) but found `--VoilaExecutePreprocessor.timeout`. However, setting that to None or -1 as the helptext suggested to disable it led to errors:

```
  File "/Users/thvasilo/opt/anaconda3/lib/python3.7/site-packages/jinja2/runtime.py", line 430, in __init__
    self._after = self._safe_next()
  File "/Users/thvasilo/opt/anaconda3/lib/python3.7/site-packages/jinja2/runtime.py", line 450, in _safe_next
    return next(self._iterator)
  File "/Users/thvasilo/opt/anaconda3/lib/python3.7/site-packages/voila/handler.py", line 145, in _jinja_cell_generator
    res = ep.preprocess_cell(cell, resources, cell_idx, store_history=False)
  File "/Users/thvasilo/opt/anaconda3/lib/python3.7/site-packages/voila/execute.py", line 156, in preprocess_cell
    result = super(VoilaExecutePreprocessor, self).preprocess_cell(cell, resources, cell_index)
  File "/Users/thvasilo/opt/anaconda3/lib/python3.7/site-packages/nbconvert/preprocessors/execute.py", line 438, in preprocess_cell
    reply, outputs = self.run_cell(cell, cell_index, store_history)
[...]

PR #589, opened 2020-05-01
Fix timeout=None

davidbrochart:
See https://github.com/voila-dashboards/voila/issues/587#issuecomment-622457068


maartenbreddels:
#541 will remove this again right?

davidbrochart:
I can make it part of #541.

maartenbreddels:
Whatever you think is more natural, i'm not sure where it should go actually after that PR is merged, since that code/method is removed.

davidbrochart:
Ah you're right, it should be in nbclient now but it's already fixed there. Closing!

PR #592, opened 2020-05-06
Use ExtensionApp

davidbrochart:


Zsailer:
@davidbrochart Just letting you know that I subscribed to this PR. I'm happy to help at anytime (review, code, etc.). Ping me if you have any questions.

davidbrochart:
Thanks @Zsailer, I think I do need your help at this point. There are three failures for which you might have an idea. Also, with a general review.

voila/app.py
```diff
@@ -7,78 +7,56 @@
...
-        'no-browser': ({'Voila': {'open_browser': False}}, _('Don\'t open the notebook in a browser after startup.'))
+        'debug': (
+            {'ServerApp': {'log_level': logging.DEBUG}},
+            _("Set the log level to logging.DEBUG")
+        ),
+        'no-browser': (
+            {'Voila': {'open_browser': False}},
+            _("Don't open the notebook in a browser after startup.")
+        ),
+        "standalone": (
```
maartenbreddels, line 56: It seems this disappeared, but I am not sure how we can detect if we are in 'standalone' mode, do you know?

davidbrochart: It think this is replaced by `load_other_extensions`, which I set to `True` because of this issue: https://github.com/jupyter/jupyter_server/issues/211

voila/app.py
```diff
@@ -7,78 +7,56 @@
...
 def _(x):
     return x
 
 
-class Voila(Application):
-    name = 'voila'
+class Voila(ExtensionApp):
+    extension_name = 'voila'
+    extension_url = '/voila'
+    load_other_extensions = True
```
maartenbreddels, line 42: Why is this set to True? This will load jupyterlab/notebook in the future if we run voila in standalone mode, or am I misunderstanding this flag?

davidbrochart: Because of this bug: https://github.com/jupyter/jupyter_server/issues/211
It is fixed now, but not released yet.

maartenbreddels:
Not sure how to fix the `enable_nbextensions` issue. Who should be responsible for this? Jupyter server?

Zsailer:
> Not sure how to fix the enable_nbextensions issue. Who should be responsible for this? Jupyter server?

I think this should be handled by [nbclassic](https://github.com/Zsailer/nbclassic). nbclassic will likely need to be a dependency in Voila. It will serve notebook frontend extensions and shim notebook traits appropriately.

smith-kyle:
Just a heads up, you can review notebook changes like this using [GitNotebooks](https://gitnotebooks.com).

e.g. this pull request [https://gitnotebooks.com/voila-dashboards/voila/pull/592](https://gitnotebooks.com/voila-dashboards/voila/pull/592)

It's free for public repos, so might be helpful for y'all.

PR #593, opened 2020-05-06
feat: convert preview to a document widget extension

mariobuikhuizen:


SylvainCorlay:
I was able to test this and it works beautifully. Thanks!

jtpio:
For reference, here is a screenshot of what it looks like:

voila-document-widget

PR #594, opened 2020-05-07
Docs: using nginx we do not want buffering

maartenbreddels:
Otherwise, we don't see the progress indicators. I think the example can be improved to allow buffering for static content, but I think this example is an improvement.


jtpio:
Waiting for Travis to complete.

PR #610, opened 2020-05-20
Don't return resources

davidbrochart:
This led to the following error when a cell was erroring out:
```
jinja2.exceptions.UndefinedError: 'tuple object' has no attribute 'get'
```

### DIFF
--- a/tests/app/syntax_error_test.py
+++ b/tests/app/syntax_error_test.py
@@ -1,0 +1,17 @@
+import pytest
+
+
+@pytest.fixture
+def syntax_error_notebook(base_url):
+    return base_url + "voila/render/syntax_error.ipynb"
+
+
+@pytest.fixture
+def voila_args(notebook_directory, voila_args_extra):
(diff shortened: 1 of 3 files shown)


maartenbreddels:
I'd like to have a test for this, do you feel like doing that?

davidbrochart:
It's tricky because a test with e.g. a syntax error in a cell doesn't fail, and doesn't show the error in `stdout` or `stderr` as if Voila were run on the CLI.

maartenbreddels:
It should give this error msg in red right? `Run with --debug` .... I think that feature isn't really tested at all.

davidbrochart:
We want to test the presence of `'tuple object' has no attribute 'get', but it's not there with pytest.

maartenbreddels:
I think we don't have to test that detailed. By default, voila should render a clean error message, now it does not do that. So I think if we just add a new test notebook, with say an import error, or syntax error, we can test if the rendered HTML contains that nice error message. What do you think about that?

davidbrochart:
Sounds good, I added a test for that and checked that it was failing before this PR.

Issue #528, opened 2020-01-27
Kernel not shut down on page reload in Firefox


jtpio:
Reloading the page in Firefox doesn't seem to be shutting the kernel down, while it does in Chrome. It is the same issue as the one mentioned in https://github.com/voila-dashboards/voila/issues/55#issue-403537084, but seems to be specific to Firefox at the moment.

Tested with:

- Firefox Developer Edition 73.0b10
- Firefox 72.0.1
- Chromium 79

vidartf:
Does the network page of the developertools reveal any special caching?

jtpio:
Investigating this further. It looks like the cache is not involved, but instead it might be related to Adblock.

With Adblock enabled, the shutdown request doesn't show up in the network logs:

![image](https://user-images.githubusercontent.com/591645/73352192-e735ad80-4290-11ea-9721-9f8651081e01.png)

And with Adblock disabled, the shutdown request (`DELETE`) is logged:

![image](https://user-images.githubusercontent.com/591645/73352245-016f8b80-4291-11ea-8a89-d38efbef3915.png)

With Chrome, the `DELETE` request is fired whether Adblock is enabled or not:

![image](https://user-images.githubusercontent.com/591645/73352606-c6218c80-4291-11ea-9863-9082f2ebce87.png)

jtpio:
Moreover, the request doesn't show up in the Adblock tab in the dev tools. This might be because the page is reloaded and the list of blocked requests deleted.

Interestingly, whitelisting `localhost` doesn't let the `DELETE` request go through either.

jtpio:
It looks like it can be reproduced by enabling other browser extensions, such as [HTTPS Everywhere](https://addons.mozilla.org/en-US/firefox/addon/https-everywhere/) (and disabling ad-blockers).

vidartf:
I'm looking into if it makes sense to start a ServiceWorker here for managing shutdown submissions. A short postMessage in `onbeforeunload` would hopefully be more reliable.

vidartf:
Or possibly `sendBeacon`[https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon] ?

jtpio:
`sendBeacon` looks interesting and could be a substitute for what is currently in https://github.com/voila-dashboards/voila/pull/612 (although it seems to be for POST requests only).

maartenbreddels:
Even better! Learning every day :) Thanks @vidartf !

[...]

Issue #55, opened 2019-01-27
Kernel shutdown on page refresh


jtpio:
Looks like kernels are not shut down when refreshing the page (using the standalone app):

![voila-kernels](https://user-images.githubusercontent.com/591645/51799790-1badfe00-2226-11e9-87eb-8b4ab62d7a0b.gif)

Seems to be related to how the `beforeunload` event is triggered:

https://github.com/QuantStack/voila/blob/c437a7fc564aee20e76526fa3c9acc5a5fef326f/voila/static/main.js#L12-L14

jf---:
Ran into the same thing today; 

- especially when hitting reload for ~10 repeatedly, this indeed results in 10x new kernels spinning up.

The reason why the kernels arent shutting down due to:
https://github.com/QuantStack/voila/blob/c437a7fc564aee20e76526fa3c9acc5a5fef326f/voila/app.py#L154-L162

While its too easy to spin up / [forkbomb](https://en.wikipedia.org/wiki/Fork_bomb) the server, at least processes do get reap'd when the following options are thrown in the mix:

```python
        self.kernel_manager = MappingKernelManager(
            cull_busy=False,
            cull_connected=False,
            cull_idle_timeout=6,
            kernel_info_timeout=12,
           ...
```

jtpio:
@jf--- Looked at it earlier today with the intent of switching from using voila as a server extension to the standalone app. As a server extension the kernel culling can be set via the notebook config.

> ```python
>         self.kernel_manager = MappingKernelManager(
>             cull_busy=False,
>             cull_connected=False,
>             cull_idle_timeout=6,
>             kernel_info_timeout=12,
>            ...
> ```
> 
> ( PR follows shortly )

I opened #80, but feel free to comment or add anything missing against this PR.

jf---:
hi @jtpio , so the PR misses out on a few things I think:

First, the `cull_connected` should be True ( yikes, excuse me... ) and is missing in the PR.
This is critical, since otherwise effectively no culling will take place.

```python
           cull_busy=False,
            cull_connected=True,
```

Furthermore, I do think that healthy defaults need to be set for `cull_idle_time` and `kernel_info_timeout`

```python
            cull_idle_timeout=6,
            kernel_info_timeout=12,
```

As the PR is as of now, the issue effectively isn't resolved, since the defaults yields the same program state as pre-PR...

[...]

PR #612, opened 2020-05-20
Send kernel shutdown request manually on beforeunload

jtpio:
cc @maartenbreddels 

This manually sends a `DELETE /api/kernels/<kernel-id>` with async = false

https://www.w3schools.com/js/js_ajax_http_send.asp

Related: https://github.com/voila-dashboards/voila/issues/528 https://github.com/voila-dashboards/voila/issues/55

kernel-shutdown-page-refresh

### DIFF
--- a/share/jupyter/voila/templates/base/static/main.js
+++ b/share/jupyter/voila/templates/base/static/main.js
@@ -50,9 +50,12 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
 
         async function init() {
             // it seems if we attach this to early, it will not be called
+            const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
+            const xsrfToken = (matches && matches[1]) || '';
             window.addEventListener('beforeunload', function (e) {
                 const xhttp = new XMLHttpRequest();
                 xhttp.open("DELETE", `/api/kernels/${kernel.id}`, false);
+                xhttp.setRequestHeader('X-XSRFToken', xsrfToken);
                 xhttp.send();
                 kernel.dispose();
             });
--- a/share/jupyter/voila/templates/base/static/main.js
+++ b/share/jupyter/voila/templates/base/static/main.js
@@ -51,7 +51,9 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
         async function init() {
             // it seems if we attach this to early, it will not be called
             window.addEventListener('beforeunload', function (e) {
-                kernel.shutdown();
+                const xhttp = new XMLHttpRequest();
+                xhttp.open("DELETE", `/api/kernels/${kernel.id}`, false);
+                xhttp.send();
                 kernel.dispose();
             });
             await widgetManager.build_widgets();


share/jupyter/voila/templates/default/static/main.js
```diff
@@ -40,7 +40,9 @@ require(['static/voila'], function(voila) {
         async function init() {
             // it seems if we attach this to early, it will not be called
             window.addEventListener('beforeunload', function (e) {
-                kernel.shutdown();
+                const xhttp = new XMLHttpRequest();
+                xhttp.open("DELETE", `/api/kernels/${kernel.id}`, false);
```
vidartf, line 44: Note that if end up deciding that this is the best way, we would need to ensure we get/set all the appropriate headers and cookies.

jtpio: Which is why using `sendBeacon` instead sounds interesting :+1:  (https://github.com/voila-dashboards/voila/issues/528 for more context)

vidartf: There are more things than just cookies though, i.e. a bunch of headers:
https://github.com/jupyterlab/jupyterlab/blob/64425baacfba677b75efaf05cfb825a82264230d/packages/services/src/serverconnection.ts#L236-L282

How many of those are needed for the DELETE request is a bit unknown though.

I think it might be possible to do a workaround by using a ServerConnection with our own settings that override `fetch` such that it instead uses a sync XHTTPRequest? It would require "translating" the request first though.

jtpio: > There are more things than just cookies though, i.e. a bunch of headers:

Indeed. And these are not picked but by `sendBeacon`.

jtpio:
Experimented a bit with `sendBeacon` locally. The following diff exposes a separate POST endpoint to shut the kernels down (following the idea in https://github.com/voila-dashboards/voila/issues/528):

```diff
diff --git a/share/jupyter/voila/templates/default/static/main.js b/share/jupyter/voila/templates/default/static/main.js
index ae5baef..2e1ce4e 100644
--- a/share/jupyter/voila/templates/default/static/main.js
+++ b/share/jupyter/voila/templates/default/static/main.js
@@ -40,9 +40,7 @@ require(['static/voila'], function(voila) {
         async function init() {
             // it seems if we attach this to early, it will not be called
             window.addEventListener('beforeunload', function (e) {
-                const xhttp = new XMLHttpRequest();
-                xhttp.open("DELETE", `/api/kernels/${kernel.id}`, false);
-                xhttp.send();
+                window.navigator.sendBeacon('/api/voila/shutdown/' + kernel.id);
                 kernel.dispose();
             });
             await widgetManager.build_widgets();
diff --git a/voila/app.py b/voila/app.py
index 17d9f33..067165d 100644
--- a/voila/app.py
+++ b/voila/app.py
@@ -55,6 +55,7 @@ from ipython_genutils.py3compat import getcwd

 from .paths import ROOT, STATIC_ROOT, collect_template_paths
 from .handler import VoilaHandler
+from .stophandler import VoilaStopHandler
 from .treehandler import VoilaTreeHandler
 from ._version import __version__
 from .static_file_handler import MultiStaticFileHandler, WhiteListFileHandler
@@ -505,6 +506,9 @@ class Voila(Application):
                  }),
             ])

+        handlers.append(
+            (url_path_join(self.server_url, r'/api/voila/shutdown/(.*)'), VoilaStopHandler)
+        )
         self.app.add_handlers('.*$', handlers)
         self.listen()

diff --git a/voila/stophandler.py b/voila/stophandler.py
new file mode 100644
index 0000000..b8a677d
--- /dev/null
+++ b/voila/stophandler.py
@@ -0,0 +1,10 @@
+import tornado
+
+from jupyter_server.base.handlers import APIHandler
+
+class VoilaStopHandler(APIHandler):
+    @tornado.web.authenticated
+    async def post(self, kernel_id):
+        await self.kernel_manager.shutdown_kernel(kernel_id)
+        self.set_status(204)
+        self.finish()
```

Although this seems to be fine in voila standalone locally, we are indeed missing the authorization headers (when uses as a server extension for example).

jtpio:
https://github.com/voila-dashboards/voila/pull/612/commits/2ea00a677519a4ccd265035eec83f5db3df8ccba and https://github.com/voila-dashboards/voila/pull/612/commits/deda754c894d8378f812ae0b723d6b6d885ce2a3 seem to improve things slightly, mostly in the case of the preview extension.

Before:

refresh-kernel-no-shutdown


After:

refresh-kernel-shutdown


However it doesn't prevent a user to hit refresh very quickly (#55). In that case not all the requests make it and kernels are not shut down.

It also doesn't seem to be effective when the preview panel is closed.

jtpio:
@vidartf let me know what you think if you get to try it locally.

It should already be a slight improvement. Although a more robust fix would probably be to detect when a connection is closed on the server instead, and shut down the corresponding kernel (or at least make that behavior configurable).

maartenbreddels:
Is this something we'd like to test with say puppeteer?

Also, I think long term we should upstream this into jupyterlab right?

maartenbreddels:
> Although a more robust fix would probably be to detect when a connection is closed on the server instead, and shut down the corresponding kernel (or at least make that behavior configurable).

Culling set to 0.01 minutes?

jtpio:
> Is this something we'd like to test with say puppeteer?

It would be very interesting to check how this is handled with puppeteer. Whether the request is correctly fired or not.

> Also, I think long term we should upstream this into jupyterlab right?

You mean the option to send requests synchronously? 

> Culling set to 0.01 minutes?

That could work. Otherwise wouldn't it be possible to know on the server when a client has disconnected? Because this should correspond to a websocket connection closed.

vidartf:
> Although a more robust fix would probably be to detect when a connection is closed on the server instead, and shut down the corresponding kernel (or at least make that behavior configurable).

While I believe there might be a technical solution possible here, I don't know anything about how hard it would be to code it and get a PR accepted.

> Culling set to 0.01 minutes?

This would cause a network glitch (the train driving through a tunnel) to kill your kernel, so probably not ideal, but yes, having an aggressive culling time seems to be the current best option.

Regarding the most robust solution, I saw someone recommend using a fetch request with the [keepalive](https://fetch.spec.whatwg.org/#request-keepalive-flag) flag set to true. I'm reading more on that now.

vidartf:
Regarding the keepalive flag, I think this would be the most ideal thing for us (it was designed for our use case, and it is often what `sendBeacon` uses under the hood). The only hiccup is that safari doesn't seem to support it (according to MDN, and here https://bugs.webkit.org/show_bug.cgi?id=168865).

maartenbreddels:
I think the current solution is an improvement (not a total fix though). I'd be happy to merge this, until we find a better solution, possibly in the jupyterlab org/repo?
Does that sound like a good idea for everyone?

jtpio:
> Does that sound like a good idea for everyone?

Sounds good :+1:

jtpio:
To test this branch on Binder: https://mybinder.org/v2/gh/jtpio/voila/shutdown-page-closed?urlpath=voila/tree

PR #629, opened 2020-06-15
Support specifying template from notebook metadata

cailiang9:
metadata format: 
{
  "voila": {
    "template": "voila-vuetify"
  }
}

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -20,7 +20,7 @@ from nbconvert.preprocessors import ClearOutputPreprocessor
 
 from .execute import VoilaExecutor
 from .exporter import VoilaExporter
-
+from .paths import collect_template_paths
 
 class VoilaHandler(JupyterHandler):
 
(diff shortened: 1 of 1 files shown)


timkpaine:
this probably needs documentation, and I would expect the command line `--template` to take precedence over the notebook metadata

cailiang9:
voila already has --template options, in standalone mode.
template options in notebook metadata will allow debugging in jupyter, other than standalone mode.

maartenbreddels:
I like it, and this is something we wanted to have for quite a while, and should almost be possible now that #208 is merged. There still is an issue though. Depending on the template, the static files that are served are different. I'll see if I can work on that.

> I would expect the command line `--template` to take precedence over the notebook metadata

I would expect the metadata to be more specific since it 'belongs' to the notebook. You can imagine having 2 notebooks with 2 different templates. While I see the cmd line argument (or the trait value) as a 'default' (least priority).

timkpaine:
@maartenbreddels right but if you want a unified look and feel you don't want individual notebooks breaking that. Just means yet another customization we'll have to do (stripping metadata) to make voila more friendly for enterprise.

maartenbreddels:
I think indeed it should be possible to disable that, with exactly that use
case in mind. Would you be happy with that?

(from mobile phone)

maartenbreddels:
Replaced by https://github.com/voila-dashboards/voila/pull/637

PR #632, opened 2020-06-17
Prompt before open_voila.

cailiang9:
Prompt before open_voila to avoid accidentally trigger voila button.
Because some temporary cells would contain commands such as "!rm file.csv", thus executing all cells would cause data loss.


cailiang9:
looks like the check failure is irrelevant to the modification of voila/static/extension.js .

jtpio:
Thanks @cailiang9.

It would be nice to still stay one click away from the voila dashboard for the users who don't want the modal.

Maybe the modal should show a "Do not ask again" checkbox? Or be made configurable?

Issue #105, opened 2019-04-10
Specify (some) voila options through URL parameters


SylvainCorlay:
We could specify the template name through a URL parameter.

choldgraf:
(copypasted from a duplicated issue #299)

I've been getting a lot of questions from people asking how they can share Voila dashboards that can both remove the code **and** show the code depending on the notebook. 

To follow up on this issue, it would be useful if one could pass options when Voila is being used as part of the Jupyter server (e.g. within Binder). Something like:

`myhub.org/voila/pathtonotebook?parameter=value`

and then this could control some of the behavior from within a running jupyter session. This could then be used to control whether code is hidden or not, e.g.:

`myhub.org/voila/pathtonotebook?stripOutputs=true`

or

`myhub.org/voila/pathtonotebook?stripOutputs=false`

jtpio:
That would definitely be useful indeed.

Perhaps there should also be a way to provide a spec of the options that can be controlled via the URL parameters (`strip_sources`, `theme`, `template`...) so that custom extensions can present the available values to the users.

These options could be requested from an endpoint (`/options` for instance) that could return something like this (inspired by the JupyterLab settings schema):

```json
{
  "options": {
    "strip_sources": {
      "title": "Strip Sources",
      "description": "Show or hide input cells",
      "default": false,
      "type": "boolean"
    },
    "theme": {
      "title": "Theme",
      "description": "Theme",
      "default": "light",
      "enum": ["light", "dark"],
      "type": "string"
    },
    "template": {
      "title": "Template",
      "description": "Template name to be used by voila",
      "default": "default",
      "enum": ["default", "gridstack", "vuetify"],
      "type": "string"
    }
  }
}
```

timkpaine:
https://github.com/voila-dashboards/voila/pull/218#issuecomment-553653123

maartenbreddels:
Partially implemented by #414 and the other idea (parametrized templates) will be covered by #637

mkcor:
@jtpio @maartenbreddels as of today, is there a way to pass the equivalent of CLI option `--strip_sources=False` either in the URL as a query parameter or in the notebook metadata? Thanks!

mkcor:
> No, but we could implement it similar to [3354ceb](https://github.com/voila-dashboards/voila/commit/3354ceb7c7b9bfc9ba12e89f87801c2424d10cdb) but except with the default being 'NOTEBOOK' or 'NO' (for security)

[...]

Issue #635, opened 2020-06-23
Per notebook template


maartenbreddels:
Now that #208 is merged, we should work towards having a per-notebook template. There should be two ways to specify this:
 * In the notebook metadata, see #629 
 * Via HTTP query string, see #201 #258 

I believe both options should be possible by default, but it also should be possible to disable this on an application level (e.g. a company/sysadmin does not want this, and it might be a security risk).

# Notebook metadata
As described in #629, it makes sense to put this under the `metadata.voila.template` JSON key in the notebook file.

# Query string
To avoid name collisions, and to be consistent with naming, I propose `?voila-template=foo`, and document that all `voila-...` query names are reserved, and the rest is for use of users (when we merge #414).

In terms of priority, it makes most sense to override them in this order:
 * VoilaConfiguration.template - seen as the default
 * Notebook metadata - overrrides the default, see this as a user set template, that matches the design/intent of the user (could be disabled by the voila server).
 * Http query string - overrides everything, useful for testing, demos etc (could also be disabled by the voila server.

# Static resources issue

Now, we have a per server template, and a single `/voila/static/(.*)` tornado hander, that assumes there is only 1 template. We could change this to `/voila/static/(?P<template>.+?)/(.*)`such that static file resolving works on a per template basis. This does mean that each template needs to hardcode its template name in the URLs in the template:
```
<link href='{{resources.base_url}}voila/static/vuetify-base/index.css' rel="stylesheet">
```

This will break the idea of an inheriting template being able to override the static files. E.g., if the vuetify-default (which inherits from vuetify-base), wants to override this css file, it always needs to change this HTML tag.

To fix this, we could expose the `template` variable via Jinja, so we get:
```
<link href='{{resources.base_url}}voila/static/{{resources.template}}/index.css' rel="stylesheet">
```

# Breakage
This does mean, we would break all templates again. Note that we already broke all templates (for the next release), so this is the right moment to do this.

Issue #258, opened 2019-06-18
Collect template paths during request


martinRenou:
cc. @maartenbreddels @SylvainCorlay @jtpio 

From what I understand of the code base, the way `template_paths` are collected feels bad to me. The paths are collected once and computed given the template name, when starting the `app` or the `server_extension`. Then the request handler always uses the same list of template paths, which makes it quite impossible to correctly choose the right template (e.g. when trying to choose the template from the `url` like @jtpio is doing in this PR #201).

I think the proper way of doing it would be to do a `collect_template_paths` for each request.

jtpio:
https://github.com/QuantStack/voila/pull/201 is collecting the template paths based on the folder structure. Then all templates can be found without specifying a template name anymore.

But it is still done only once at startup that's right.

maartenbreddels:
That is exactly why I want to change the template system, good you see it the same way. Once https://github.com/QuantStack/voila/pull/208 is in, this will be much cleaner.

martinRenou:
> #201 is collecting the template paths based on the folder structure. Then all the templates can be found without specifying a template name anymore.

Oh I see, then we collect everything. The thing is having to sort the list of template paths later feels bad to me... There should be a cleaner way of doing it.

> That is exactly why I want to change the template system, good you see it the same way. Once #208 is in, this will be much cleaner.

It is still not clear in my head what you are doing in your PR, I'll take a better look at it.

Maybe we should all discuss together, there will be conflicts with those two PRs

maartenbreddels:
Yes, as I said, it would be good to discuss this over a high bandwidth channel (video). It is good you've hit the same issues, that makes it easier to understand. Let's try on Thursday!

jtpio:
Yes that would be great.

In the end #201 and #208 complement each other:
- #201 is about going through all the installed templates and making them available to the end user
- #208 is about refactoring how these templates are stored on disk and potentially changing the search path

martinRenou:
> #201 is about going through all the installed templates and making them available to the end user

I would rather keep collecting the template paths for a specific template name, but doing it in the handler (per request). What do you think? Could it be slow?

maartenbreddels:
I thought about that as well, it might be slow indeed, I think we want to cache that.

martinRenou:
Fair enough, so another idea: instead of having a list of template paths, maybe a dictionary 
```
{ "template_name": [template_paths, ...], ... }
``` 
?

jtpio:
[...]

PR #637, opened 2020-06-24
Feature: template and theme override

maartenbreddels:
Implements #635  Replaces #629, #201 and #258

This allow overriding the template and theme, e.g.:
```
http://localhost:8866/voila/render/basics.ipynb?voila-template=material&voila-theme=dark
```

Or via notebook metadata:
```
....
 "metadata": {
  "voila": {
    "template": "material",
    "theme": "dark"
  },
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
....
```

Configuration to disable this:
```
$ voila notebooks/ --VoilaConfiguration.allow_template_override=NO  # never allow this
$ voila notebooks/ --VoilaConfiguration.allow_template_override=NOTEBOOK  # only via notebook metadata
$ voila notebooks/ --VoilaConfiguration.allow_theme_override=NOTEBOOK  # similar for theme
```

### DIFF
--- a/voila/configuration.py
+++ b/voila/configuration.py
@@ -16,6 +16,9 @@ class VoilaConfiguration(traitlets.config.Configurable):
     allow_template_override = Enum(['YES', 'NOTEBOOK', 'NO'], 'YES', help='''
     Allow overriding the template (YES), or not (NO), or only from the notebook metadata.
     ''').tag(config=True)
+    allow_theme_override = Enum(['YES', 'NOTEBOOK', 'NO'], 'YES', help='''
+    Allow overriding the theme (YES), or not (NO), or only from the notebook metadata.
+    ''').tag(config=True)
     template = Unicode(
         'lab',
         config=True,
--- a/voila/configuration.py
+++ b/voila/configuration.py
@@ -8,11 +8,14 @@
 #############################################################################
 
 import traitlets.config
-from traitlets import Unicode, Bool, Dict, List, Int
+from traitlets import Unicode, Bool, Dict, List, Int, Enum
 
 
 class VoilaConfiguration(traitlets.config.Configurable):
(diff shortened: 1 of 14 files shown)


maartenbreddels:
Note that this PR also changes the static URL to include a hash, like `voila.js?v=ae356' such that we don't get bitten by aggressive caching. This is a feature of tornado, that also the notebook uses, and we simply didn't make use of.

voila/configuration.py
```diff
@@ -8,11 +8,17 @@
...
 import traitlets.config
-from traitlets import Unicode, Bool, Dict, List
+from traitlets import Unicode, Bool, Dict, List, Enum
 
 
 class VoilaConfiguration(traitlets.config.Configurable):
     """Common configuration options between the server extension and the application."""
+    allow_template_override = Enum(['YES', 'NOTEBOOK', 'NO'], 'YES', help='''
+    Allow overriding the template (YES), or not (NO), or only from the notebook metadata.
+    ''').tag(config=True)
```
jtpio, line 18: Wondering whether something like `['url', 'notebook', 'all']` would sound more explicit?

jtpio:
Nice, thanks @maartenbreddels :+1: 

We will probably want to add a quick word about this feature in the docs too.

voila/configuration.py
maartenbreddels: My reasoning was that you probably don't want the combination of 'permit from URL', but not from notebook metadata. Or can you come up with te use case?

jtpio: It was mostly for users who want to set the option without reading the docs or the help string, to know what 'YES' and 'NO' mean in this case.

maartenbreddels:
Closes #105

PR #639, opened 2020-06-24
Remove Voilà CSS preprocessor

SylvainCorlay:
This is not needed now that this is handled in nbconvert.

Also removing requirements from `setup.py`.


maartenbreddels:
nice cleanup! LGTM

PR #643, opened 2020-07-01
fix: use template_paths instead of old template_path

maartenbreddels:
In https://github.com/jupyter/nbconvert/pull/1310 we got rid of template_path, and set the template_paths. This also means we don't have to override the get_template_paths since template_paths is a trait, where the old get_template_paths is the default (method).


maartenbreddels:
Strange failure, also see it in master

maartenbreddels:
Fixed after rerunning

PR #655, opened 2020-07-20
Add docs for the theme option

martinRenou:
@jtpio


jtpio:
Really nice, thanks!

The change can be checked on the read the docs directly: https://voila--655.org.readthedocs.build/en/655/customize.html#changing-the-theme

PR #667, opened 2020-08-03
feat: when a widget fails, show a fallback error widget

### LINKED ISSUE #657: Missing nbextensions: we should be more resilient to one widget library missing in the frontend
And enable displaying other widgets that were found.

maartenbreddels:
Fixes #657 
Consider this code:
```python
import ipywidgets as widgets
from traitlets import Unicode

class BadWidget(widgets.DOMWidget):
    _model_module = Unicode('badwidget').tag(sync=True)
    _view_module = Unicode('badwidget').tag(sync=True)
    _view_name = Unicode('DoesNotExist').tag(sync=True)
bad_widget = BadWidget()
widgets.HBox([bad_widget, widgets.Button(description='Hi')])
```

This would now render to:

image

If the semver is off:
```python
import ipywidgets as widgets
from traitlets import Unicode

class BadVersion(widgets.Button):
    _model_module_version = Unicode('11.0').tag(sync=True)
bad_version = BadVersion()
widgets.HBox([bad_version, widgets.Button(description='Hi')])
```

You'd see this:
image

And if the widget model name is off:

```python
import ipywidgets as widgets
import ipyvolume as ipv
from traitlets import Unicode

class BadName(ipv.Figure):
    _model_name = Unicode('Fig').tag(sync=True)
bad_name = BadName()
widgets.HBox([bad_name, widgets.Button(description='Hi')])
```

It renders:

image

So in all of these cases, something gets rendered. The only case that is not handled, is when the view_name is off (difficult to catch)

### DIFF
--- a/packages/voila/src/errorwidget.ts
+++ b/packages/voila/src/errorwidget.ts
@@ -1,0 +1,30 @@
+import * as widgets from '@jupyter-widgets/base';
+
+// Create a Widget Model that captures an error object
+export function createErrorWidget(error: string): any {
+  class ErrorWidget extends widgets.DOMWidgetModel {
+    constructor(attributes?: any, options?: any) {
+      attributes = {
+        ...attributes,
+        _view_name: 'ErrorWidgetView',
+        _view_module: 'voila-errorwidget',
+        _model_module_version: '1.0.0',
+        _view_module_version: '1.0.0',
+        failed_module: attributes._view_module,
+        failed_model_name: attributes._model_name,
+        error: error
+      };
+      super(attributes, options);
+    }
+  }
+  return ErrorWidget as any;
+}
+
+export class ErrorWidgetView extends widgets.DOMWidgetView {
+  render() {
+    const module = this.model.get('failed_module');
+    const name = this.model.get('failed_model_name');
+    const error = String(this.model.get('error').stack);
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
cc @jasongrout

maartenbreddels:
Me might want to have a troubleshoot link in the message. Maybe we should then also have a troubleshoot page at ipywidgets?

Another thing we should consider, is add two methods to the widget manager:
 * loadViewClass (default implementation can call loadClass)
 * loadModelClass (idem)

Since now we do not know when a Model or View needs to be loaded, we cannot create a fallback Model or View.

maartenbreddels:
Code is simplified a bit (force pushed)

maartenbreddels:
@SylvainCorlay I think we can merge this, the failure on Travis seems a fluke, and we cannot restart it 🤷 

We might want to consider upstreaming this to core jupyter-widgets (working on it), but I think that will take a while.

SylvainCorlay:
I ported this to typescript since we moved to typescript, but it does not seem we can create anonymous classes on the fly like this.

SylvainCorlay:
@maartenbreddels I just rebased your PR and translated the code to AnyScript so that it works with the new typescript codebase.

trungleduc:
This feature is added in `ipywidgets 8`.

PR #680, opened 2020-08-19
refactor: use macro for voila setup and override jupyter-widget macro

maartenbreddels:
less code duplication and makes it easier to reuse or override the macro


SylvainCorlay:
Let's have another alpha and update `voila-[gridstack/reveal]` with that?

PR #686, opened 2020-08-31
chore(ci): many_iopub_message test still timed out

maartenbreddels:
Let's run CI a few times on this to make sure this is a good enough timeout.


maartenbreddels:
Ok, this seems solid after 2 restarts.

PR #690, opened 2020-09-02
fixup-traitlets-5-breakage

SylvainCorlay:
@maartenbreddels 

Let's do a cleaner fix in master.


maartenbreddels:
Yeah, fine with that!

SylvainCorlay:
I prefer this over the cleaner fix for 0.1 because it is functionally completely equivalent.

maartenbreddels:
Yeah, i'm good with this.

Issue #689, opened 2020-09-01
voilaexporter has no attribute contents_manager for default template


richtong:
We are tearing our hair out here. We have a working configuration of Voila running Vuetify. The underlying Dashboard displays fine with Voila. The very strange thing is that we have all dependencies pinned so it seems unlikely to be a version problem, but then if you sometimes you get this error and sometimes you don't.

We think there must be something changing templates outside of our github repo. You seem to copy things into ~/share in case our pipenv, so maybe these are somehow changing? IN any case this does not work

`voila file.ipynb` using default but `voila --template=vuetify_default file.ipynb` does work. It seems intermittent. In that even a clean install will sometimes show the problem and sometimes not.

```
  File "/Users/rich/.local/share/virtualenvs/nb-FpFnjW0W/lib/python3.8/site-packages/nbconvert/exporters/../templates/skeleton/null.tpl", line 81, in block "any_cell"
    {%- block markdowncell scoped-%} {%- endblock markdowncell -%}
  File "/Users/rich/.local/share/virtualenvs/nb-FpFnjW0W/share/jupyter/voila/templates/default/nbconvert_templates/lab.tpl", line 92, in block "markdowncell"
    {{ cell.source  | markdown2html | strip_files_prefix }}
  File "/Users/rich/.local/share/virtualenvs/nb-FpFnjW0W/lib/python3.8/site-packages/voila/exporter.py", line 55, in markdown2html
    contents_manager=self.contents_manager,
'VoilaExporter' object has no attribute 'contents_manager'
```

We have seen this error before and pinned jupyter-server to 0.1.1 and this seems to help  but that is already set

We are running bqplot 0.12.16, voila-vuetify==0.4.01 and voila 0.1.22

gbrault:
I have a similar issue, just loading ipywidgets in voila:

File "/home/gilbert/venv/lib/python3.8/site-packages/voila/exporter.py", line 55, in markdown2html

**context**

[piplist.txt](https://github.com/voila-dashboards/voila/files/5160763/piplist.txt)

    contents_manager=self.contents_manager,
AttributeError: 'VoilaExporter' object has no attribute 'contents_manager'

SylvainCorlay:
cc @maartenbreddels

maartenbreddels:
looking into this.
@richtong which version of nbconvert are you using?

SylvainCorlay:
This appears to be due to the recent release of traitlets 5.0.0, which is backward incompatible.

[...]

PR #691, opened 2020-09-02
fix: content_manager was not a trait

maartenbreddels:
Equivalent of #690, now using traitlets, fix for #689


maartenbreddels:
I think the problem is that we use `--ServerApp.token=''`, which is now the Python string "''" (e.g. not an empty string!) in trailets 5 :(
this must also effect all notebook users I guess?

SylvainCorlay:
I removed your last commit working around the `None` bug which is solved in traitlets 5.0.3.

PR #700, opened 2020-09-11
Fix installation steps in the doc tests section

martinRenou:


docs/source/contribute.rst
```diff
@@ -143,13 +143,13 @@ Install the test dependencies
 
 .. code-block:: bash
 
-   python -m pip install -e ".[test]"
+   python -m pip install ".[test]"
```
maartenbreddels, line 146: We can leave the -e if we enable the extension right?

martinRenou: Maybe? I was not able to run any test on my side. I did not find a combination that works.

maartenbreddels: Did you run `(cd tests/test_template; pip install .)` ?

martinRenou: yep

Issue #708, opened 2020-09-17
500 with voila 0.2.1 as a server extension


jtpio:
Running voila 0.2.1 as a server extension with:

```bash
conda create -n voila-test -c conda-forge voila ipywidgets pandas python notebook -y
conda activate voila-test
jupyter notebook --no-browser
```

And rendering a notebook with http://localhost:8888/voila/render/basics.ipynb gives the following:

```
      File "/home/jtp/miniconda/envs/voila-test/lib/python3.8/site-packages/traitlets/traitlets.py", line 690, in error
        raise TraitError(e)
    traitlets.traitlets.TraitError: The 'contents_manager' trait of a VoilaExporter instance expected a ContentsManager, not the LargeFileManager at '0x7fab992d3100'.
```

This seems to be reproducible with the current Binder link too:

https://mybinder.org/v2/gh/voila-dashboards/voila/43fcf505622c66005e542ec19e3b39b505233cea?urlpath=%2Fvoila

![image](https://user-images.githubusercontent.com/591645/93476616-4d437380-f8fa-11ea-9419-c01933e9d2e1.png)

**Versions**

```
$ conda list | grep -E "(voila|notebook|traitlets|nbconvert|nbclient)"
nbclient                  0.5.0                      py_0    conda-forge
nbconvert                 6.0.3            py38h32f6830_0    conda-forge
notebook                  6.1.4            py38h32f6830_0    conda-forge
traitlets                 5.0.4                      py_0    conda-forge
voila                     0.2.1                      py_0    conda-forge
```

SylvainCorlay:
should we pin voilà to 4.3.3 while we fix these issues.

maartenbreddels:
This seems unrelated to traitlets, could be because of notebook and jupyter_server both having those classes... just weird we have not seen this before.

maartenbreddels:
Of course, before this was not a trait!

maartenbreddels:
I see no other solution then to make this of 'Any' type.

jtpio:
Was wondering why it was not caught on CI but `traitlets` seems to be pinned here:

https://github.com/voila-dashboards/voila/blob/43fcf505622c66005e542ec19e3b39b505233cea/.github/workflows/main.yml#L34

SylvainCorlay:
So it *is* a traitlets issue?

jtpio:
Actually tests run with both traitlets 4 and 5?

[...]

PR #709, opened 2020-09-17
fix: allow contents_manager to be a notebook content manager

maartenbreddels:
Fixes https://github.com/voila-dashboards/voila/issues/708 which was caused by  https://github.com/voila-dashboards/voila/pull/691

This can be reverted once notebook uses jupyter server.


jtpio:
Thanks @maartenbreddels :+1: 

Sounds like the simplest for now.

PR #710, opened 2020-09-17
fix: support DeferredConfigString (hack)

maartenbreddels:
This is the best backward compatible fix I can think of, but I did not confirm yet it works.


maartenbreddels:
TBH, why are we using `self.voila_configuration.config.VoilaConfiguration.resources` instead of just `self.voila_configuration.resources`?

SylvainCorlay:
@jtpio @maartenbreddels should this be enough to work with the classic notebook server? Ready to go?

jtpio:
Just tried this fix together with https://github.com/voila-dashboards/voila/pull/709, and it seems to work with the classic notebook server :+1:

maartenbreddels:
The travis failure is strange, but cannot be related to this. I'm ok-ish to merge this.

SylvainCorlay:
Merged. When all is green, I will tag 0.2.2.

PR #719, opened 2020-09-24
fix: when using the back button Voila doesn't load

mariobuikhuizen:
When navigating away and later coming back via the browser back (or forward) button, the Voilà page doesn't load, showing the following error in the console: `Error: No running kernel with id: 24b5b2dd-f00b-4149-80b5-036ae1f471de`

This PR fixes that by preventing the browser to use a cached version of the page with a stale kernel id.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -147,6 +147,9 @@ class VoilaHandler(JupyterHandler):
 
         # Compose reply
         self.set_header('Content-Type', 'text/html')
+        self.set_header('Cache-Control', 'no-cache, no-store, must-revalidate')
+        self.set_header('Pragma', 'no-cache')
+        self.set_header('Expires', '0')
         # render notebook in snippets, and flush them out to the browser can render progresssively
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Great, do you have resources where you found this info? Why the 3 headers, why not more or less?

mariobuikhuizen:
Good point, I used the following info:

https://stackoverflow.com/questions/49547/how-do-we-control-web-page-caching-across-all-browsers#answer-2068407

maartenbreddels:
Excellent work, I think many people have hit this issue!

SylvainCorlay:
LGTM!

nuraymuzaffarova:
We will definitely benefit from this. Thanks!

maartenbreddels:
Thanks Nuray!

jtpio:
cc @agoose77 (as this might fix https://gitter.im/QuantStack/Lobby?at=5f6738bddf4af236f91c594f)

Issue #712, opened 2020-09-19
Pre-execute notebooks


davidbrochart:
In order to hide the start-up time, it would be nice to have the possibility to create a kernel and have it execute the notebook in advance, so that when we enter the URL in the browser, all there is left to do is rendering the HTML and connecting to the kernel.
We could have an option such as `--pre-execute=nb` (default to 0), that would create a pool of already executed notebooks ready to be served, in case there is sudden batch of connections.

maartenbreddels:
Yes, I think that's a really good thing to have, especially when voila hosts a single notebook. The only requirement is that the notebook is a 'pure function', i.e. not dependant on parameters, time, etc.

SylvainCorlay:
Also linking
 - https://github.com/jupyter/jupyter_client/pull/441
 - https://github.com/ipython/ipykernel/pull/410

timkpaine:
> Yes, I think that's a really good thing to have, especially when voila hosts a single notebook. The only requirement is that the notebook is a 'pure function', i.e. not dependant on parameters, time, etc.

For those that are not, or when you want to do things like per-user auth inside the notebook once a user has connected, it would be useful to have a few options:

- no preexecution
- fully preexecute notebook
- partial prexecution (start notebooks, maybe accept a list of modules to import to "warm" the kernel, e.g. things like `numpy`, `pandas`, `tensorflow`, anything else that you might want to setup but without any of the side effects of executing the notebook) 

As always, fully preexecuting the notebook should never be the default or only behavior since this makes per-user authenticated notebooks difficult or impossible (we use the details of the user's request to inject auth into the notebook).

davidbrochart:
> partial prexecution (start notebooks, maybe accept a list of modules to import to "warm" the kernel, e.g. things like `numpy`, `pandas`, `tensorflow`, anything else that you might want to setup but without any of the side effects of executing the notebook)

partial pre-execution sounds to me like "execute first N cells", and we would let the author of the notebook arrange it so that the first N cells are user-independent.

timkpaine:
> partial pre-execution sounds to me like "execute first N cells", and we would let the author of the notebook arrange it so that the first N cells are user-independent.

this presupposes that the notebook author and the voila deployer are the same person (or closely in sync), which is not always the case. However with my prelaunch hook, I can always inject cells, so I can implement "warm these modules" with my own custom logic by injecting a cell and saying "warm 1 cell". So tl;dr, I can use it if we say partial is "n cells".

[...]

PR #724, opened 2020-09-25
reenable prelaunch-hook

timkpaine:
A function that is called prior to the launch of a new kernel instance
when a user visits the voila webpage. Used for custom user authorization
or any other necessary pre-launch functions.
                                   
Should be of the form

```python
def hook(req: tornado.web.RequestHandler,
                notebook: nbformat.NotebookNode,
                cwd: str)
```

Although most customizations can leverage templates, if you need access
to the request object (e.g. to inspect cookies for authentication),
or to modify the notebook itself (e.g. to inject some custom structure,
althought much of this can be done by interacting with the kernel
in javascript) the prelaunch hook lets you do that.


vvatsa:
Would echo the previous sentiment, I too would  appreciate if the maintainers can comment on Tims' patch.

angusfong:
Echoing the above two comments. Please advise on how I can pass a single string variable to my notebook. I am using `voila.app.Voila.start()` to create the kernel, per https://github.com/voila-dashboards/voila/pull/218#issue-454676599.

timkpaine:
I am getting this back up and working, is there any specific examples/docs I should add? I no longer have access to the how I used this to enable custom token-based authentication at JP Morgan (maybe @vidartf can construct some examples off this) but im planning on adding a simple test to extract a query parameters and run papermill so I can close https://github.com/voila-dashboards/voila/pull/484

alkasm:
This PR would solve my needs for authentication in a very clean way. If I want to support auth via authorization token or cookies, I could simply read them via this hook and then mutate a notebook cell to include the authorization information required to make additional client requests.

.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
     - name: Install Dependencies
       run: |
-        python -m pip install -U jupyterlab~=3.0 jupyter_packaging~=0.10
+        python -m pip install -U jupyterlab~=3.0 jupyter_packaging~=0.10 papermill
```
alkasm, line 25: Looks like you may need to add this here: https://github.com/voila-dashboards/voila/blob/main/setup.cfg#L49-L56 instead

alkasm: I believe this is where the test deps get installed for CI: https://github.com/voila-dashboards/voila/blob/de712eae5e64ed880177cc80e21f9c77e4b1a02e/.github/workflows/main.yml#L42

Link to CI error line for easy access: https://github.com/voila-dashboards/voila/runs/6181097089?check_suite_focus=true#step:6:93

alkasm: Ah and on the windows tests it uses this line instead: https://github.com/voila-dashboards/voila/blob/de712eae5e64ed880177cc80e21f9c77e4b1a02e/.github/workflows/main.yml#L85

alkasm:
Well I put the minor fix in a PR here hoping to see the workflows run: https://github.com/voila-dashboards/voila/pull/1148
but you need a maintainer to allow the workflows to run if you're a first-time contributor (sensible policy). If you have time today, you can apply that commit here, or whatever makes the most sense.

timkpaine:
I will do it here tonight, apologies

timkpaine:
@alkasm I merged your branch into mine, as a weird side effect your PR and my PR are now both running the tests but I think they're the same

alkasm:
Awesome, and no worries on the timeline. I deleted my branch. Thanks for being so responsive on a relatively stale PR!

timkpaine:
Wanted to note that this is NOT a breaking change, so does not necessarily need to wait for 4.0.

alkasm:
Seems like the tests are flaky? The previous commit `d10918b` failed on macos 3.7 and 3.9 and win 3.8 and 3.10, and the no-code commit `9a1619f` failed on macos 3.8, win 3.7 and 3.10. So by the sandwich theorem (or some other nonsense) that just leaves 3.10, but it has similar timeout errors to the other flakes.

timkpaine:
Yeah it looks like flaky tests, I tried 4-5 times (rebased and force pushed empty commits) and got different tests failing every time

timkpaine:
@trungleduc @jtpio @maartenbreddels @martinRenou @davidbrochart ready for review / discussion

timkpaine:
Just bumping, I can't sponsor this feature right now with my current employer but I can reveal that a large former employer, J.P. Morgan, uses/used this feature extensively which is a decent endorsement:

https://tim.paine.nyc/talks/jupytercon.html#/5/1/1

trungleduc:
Hi @timkpaine, thank you very much for your contribution! I'm quite busy during this period but I will take a look at this PR in a few weeks. 
My plan is to target this PR to `0.4.0` in order to facilitate the tracking of new features then we can back-port it to `0.3.x`

alkasm:
I've noticed that this only works when you serve a single notebook, e.g. by overriding the `notebook_path` argument on the `voila.app.Voila` class or equivalently by passing a positional arg to the `voila` CLI. If you try to serve a directory of notebooks (e.g. with `root_dir` on the `Voila` class or by launching the `voila` CLI with no positional args), none of the requests get intercepted: neither the request which serves the directory nor the `<base_url>/voila/render/<notebook>.ipynb` links provided in the directory view.

docs/source/customize.rst
```diff
@@ -209,6 +209,59 @@ There is a Voilà template cookiecutter available to give you a running start.
 This cookiecutter contains some docker configuration for live reloading of your template changes to make development easier.
 Please refer to the `cookiecutter repo <https://github.com/voila-dashboards/voila-template-cookiecutter>`_ for more information on how to use the Voilà template cookiecutter.
 
+Accessing the tornado request (`prelaunch-hook`)
+---------------------------------------------------
+
+Unfortunately it is not currently possible to use custom templates to access the tornado request object, which might be necessary in certain custom setups if you need
```
trungleduc, line 215:
```suggestion
Unfortunately, it is not currently possible to use custom templates to access the tornado request object, which might be necessary in certain custom setups if you need
```

docs/source/customize.rst
```diff
@@ -209,6 +209,59 @@ There is a Voilà template cookiecutter available to give you a running start.
...
+to check for authentication cookies or access details about the request headers, etc.
+Instead, you can leverage the `prelaunch-hook`, which lets you inject a function to inspect the notebook and request prior to executing them.
```
trungleduc, line 217:
```suggestion
Instead, you can leverage the `prelaunch-hook`, which lets you inject a function to inspect the notebook and the request prior to executing them.
```

docs/source/customize.rst
```diff
@@ -209,6 +209,59 @@ There is a Voilà template cookiecutter available to give you a running start.
...
+
+The format of this hook should be:
+
+.. code-block:: python
+
+   def hook(req: tornado.web.RequestHandler,
+            notebook: nbformat.NotebookNode,
+            cwd: str) -> Optional[nbformat.NotebookNode]:
+
+Here is an example of a custom `prelaunch-hook` to execute a notebook with `papermill`:
```
trungleduc, line 227: Can you add the description of the arguments and output of the hook function?

docs/source/customize.rst
```diff
@@ -209,6 +209,59 @@ There is a Voilà template cookiecutter available to give you a running start.
...
+            # loading logic
+            for cell in notebook.cells:
+                if 'tags' not in cell.metadata:
+                    cell.metadata.tags = []
+                if "papermill" not in notebook.metadata:
+                    notebook.metadata.papermill = {}
+
+            # Parameterize with papermill
+            return parameterize_notebook(notebook, parameters)
+
```
trungleduc, line 263: Can you describe how we can add the hook function to the `Voila` app?

trungleduc:
Thanks @timkpaine for your contribution, it looks really nice.
I have one question about the hook and preheated kernel mode. In this mode, the notebook is rendered only once, so the hook function is only called once and usually when no user is connected. So any hook that reads the request headers is not usable in preheated kernel mode. How do we inform users about the incompatibility of these two features?

trungleduc:
> I've noticed that this only works when you serve a single notebook, e.g. by overriding the `notebook_path` argument on the `voila.app.Voila` class or equivalently by passing a positional arg to the `voila` CLI. If you try to serve a directory of notebooks (e.g. with `root_dir` on the `Voila` class or by launching the `voila` CLI with no positional args), none of the requests get intercepted: neither the request which serves the directory nor the `/voila/render/<notebook>.ipynb` links provided in the directory view.
> 
> edit: I think probably adding it to here is sufficient:
> 
> https://github.com/voila-dashboards/voila/blob/e658bbee2cd07d2eb367825e435cc18956f9d06c/voila/app.py#L568-L573
> 
> I'll test that out shortly
> 
> edit 2: Yeah, that works

I confirm, the `prelaunch_hook` does not work with the case of serving a directory.

timkpaine:
> Thanks @timkpaine for your contribution, it looks really nice.
> I have one question about the hook and preheated kernel mode. In this mode, the notebook is rendered only once, so the hook function is only called once and usually when no user is connected. So any hook that reads the request headers is not usable in preheated kernel mode. How do we inform users about the incompatibility of these two features?

That is correct, I can add a note to the documentation. I have long rallied against the preheat mode for this reason: https://github.com/voila-dashboards/voila/issues/712#issuecomment-695789886

trungleduc:
I think the documentation for this conflict is not enough, we might need some checks on startup and warn users if they use both features.

trungleduc (approved):

trungleduc:
Thanks @timkpaine!

timkpaine:
thanks @trungleduc ! cc @vidartf @ptomecek

jtpio:
Just updated the milestone to `0.3.x`. 

According to the diff and the [comment above](https://github.com/voila-dashboards/voila/pull/724#issuecomment-1110470839). there doesn't seem to be breaking changes, so this can be shipped in the next `0.3.x` release.

PR #725, opened 2020-09-25
switch to allowlist and denylist

timkpaine:


SylvainCorlay:
Thanks @timkpaine.

Since this is a change of configuration key, it should probably be in a 0.3.x release, unless we find a way to cleanly deprecate the previous wording.

timkpaine:
@jtpio @trungleduc mind approving me, first contributions since GitHub turned on workflow approvals

trungleduc:
Hi @timkpaine, I think we can keep both `file_blacklist`, `file_whitelist` and add a validator for them which will issue a `DeprecationWarning`  and suggest the new keys.

Issue #729, opened 2020-09-27
ipympl.ipynb doesn't work on Safari/Firefox


abunimeh:
I am running on macos 10.15.

steps to reproduce

```
git clone ...
conda env update -f environment.yml
cd notebooks/
voila ipympl.ipynb
```
The notebooks renders correctly on Chrome Version 85.0.4183.121
It doesn't work on Safari v14/Firefox v81.0.

Screenshots attached
<img width="1682" alt="image" src="https://user-images.githubusercontent.com/8106823/94356644-ab343180-0045-11eb-94b6-5498b2648379.png">
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/8106823/94356661-c8690000-0045-11eb-98d0-c1c15e571a3f.png">
<img width="1258" alt="image" src="https://user-images.githubusercontent.com/8106823/94356682-ee8ea000-0045-11eb-9515-a8f18efe79a2.png">

martinRenou:
@abunimeh Do you think it's the same issue an in here: https://github.com/voila-dashboards/voila/issues/693? Can you look at the JavaScript console and confirm you see the following error message:
`Uncaught (in promise) DOMException: The object could not be cloned.`

abunimeh:
yes it seems to be the same error reported in that ticket.

Safari spits out:
```
[Error] Unhandled Promise Rejection: DataCloneError: The object can not be cloned.
postMessage (localhost:30)
(anonymous function) (localhost:30)
(anonymous function) (voila.js:137:111880)
promiseReactionJob
```
Firefox:
```
Uncaught (in promise) DOMException: The object could not be cloned. localhost:8866:30
```

martinRenou:
Thanks, this should be fixed on master. It will be included in the next release. Thanks for reporting.

Issue #693, opened 2020-09-04
ipympl figures not shown in Firefox (works fine with Chrome)


paugier:
It's sad... Firefox 80.0

I reproduce with voila/notebooks/ipympl.ipynb

Voila, ipympl and all packages installed in an empty environment today with conda.

The console in Firefox says `Uncaught (in promise) DOMException: The object could not be cloned.`

I tried with `voila ipympl.ipynb --debug`. The log ends by
```
WARNING:tornado.general:404 GET /api/kernels/d5b19400-beaf-434c-9053-a0ad52b539fe/channels?session_id=2a60a4bb-36f4-49c0-9f7a-a8e6f34f5300 (127.0.0.1): Kernel does not exist: d5b19400-beaf-434c-9053-a0ad52b539fe
[Voila] Using contents: services/contents
WARNING:tornado.access:404 GET /api/kernels/d5b19400-beaf-434c-9053-a0ad52b539fe/channels?session_id=2a60a4bb-36f4-49c0-9f7a-a8e6f34f5300 (127.0.0.1) 36.14ms
[Voila] Initializing websocket connection /api/kernels/d5b19400-beaf-434c-9053-a0ad52b539fe/channels
[Voila] WARNING | Replacing stale connection: d5b19400-beaf-434c-9053-a0ad52b539fe:2a60a4bb-36f4-49c0-9f7a-a8e6f34f5300
[Voila] Initializing websocket connection /api/kernels/d5b19400-beaf-434c-9053-a0ad52b539fe/channels
[Voila] WARNING | Replacing stale connection: d5b19400-beaf-434c-9053-a0ad52b539fe:2a60a4bb-36f4-49c0-9f7a-a8e6f34f5300
```

paugier:
It is really a very annoying issue that Voilà + ipympl do not work with recent versions of Firefox. 

Note that it works fine with Firefox + Jupyter Notebook + ipympl.

paugier:
Ah interesting, it works fine with voila 0.1.21, as in https://github.com/voila-dashboards/voila/blob/master/environment.yml.

But not with  0.1.22,  0.1.23 and master.

martinRenou:
I see no problem with latest ipympl, voila 0.1.23, and Firefox 80.0.1 (64-bit). Which Firefox version do you have?

Could you paste a bit more detail about the `Uncaught (in promise) DOMException: The object could not be cloned`? It seems like the interesting part, in which file was this triggered?

[...]

PR #731, opened 2020-09-28
fix: do not fail when a comm msg has no buffers

maartenbreddels:
This happens with ipyml, and because of #730 was not visible in the
JS console for Firefox (and Safari).
Related #729 #693


maartenbreddels:
Fixes  #729 #693

martinRenou:
Thanks a lot for fixing this @maartenbreddels !

PR #732, opened 2020-09-28
Update the preview extension to JupyterLab 3.0

jtpio:
Update the preview extension so it is compatible with JupyterLab 3.0.

Note: this only updates the compatibility range, and does not use the new extension system and packaging (yet).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,12 +25,12 @@ jobs:
       with:
        node-version: '12.x'
     - name: Install Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: '3.7'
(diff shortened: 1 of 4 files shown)


jtpio:
This should now be compatible with the latest rc:

lab-3-rc2

SylvainCorlay:
Feel free to go ahead and merge when you think this is ready.

jtpio:
Looks like the tests are now failing, similar to https://github.com/voila-dashboards/voila-gridstack/issues/70

jtpio:
Keeping a pin on `jupyter_server` for now for the tests. To be fixed in #785.

Follow-up PR will be to use the new distribution system for the preview extension: https://github.com/voila-dashboards/voila/issues/761

PR #735, opened 2020-10-05
Move require.min.js script element to be above notebook execution

jwminton:
When base.tpl was reworked into index.html.j2 files, the require.min.js script element was relocated from the top of the page to just above voila's requirejs(), after the notebook cell loads. This removed it from notebook cell scope and breaks plotly (and anything else depending on require.min.js already being called). 

This puts the script element back to where it was by taking it from the voila_setup macro and piggybacking on the logging macro. 

A more principled fix might mirror the logging macro instead of appropriating it.

### DIFF
--- a/share/jupyter/voila/templates/base/log.macro.html.j2
+++ b/share/jupyter/voila/templates/base/log.macro.html.j2
@@ -1,13 +1,5 @@
 {% macro js() %}
-
-<!-- voila require js -->
-  <script
-    src="{{base_url}}voila/static/require.min.js"
-    integrity="sha256-Ae2Vz/4ePdIu6ZyI/5ZGsYnb+m0JlOmKPjt6XZ9JJkA="
-    crossorigin="anonymous">
-  </script>
-
-<!-- voila log js -->
+  <!-- voila log js -->
   <script>
     const _debug = console.debug;
     const _info = console.info;
--- a/share/jupyter/voila/templates/base/log.macro.html.j2
+++ b/share/jupyter/voila/templates/base/log.macro.html.j2
@@ -1,5 +1,13 @@
 {% macro js() %}
-  <!-- voila log js -->
+
+<!-- voila require js -->
+  <script
+    src="{{base_url}}voila/static/require.min.js"
+    integrity="sha256-Ae2Vz/4ePdIu6ZyI/5ZGsYnb+m0JlOmKPjt6XZ9JJkA="
+    crossorigin="anonymous">
+  </script>
+
+<!-- voila log js -->
   <script>
     const _debug = console.debug;
     const _info = console.info;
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Hi,

thanks for this fix. I think this should have been the way we've done things, except putting the script tag in the index.html.j2. But this requires changes to
 * lab and classic template 
* voila-reveal https://github.com/voila-dashboards/voila-reveal/blob/d5b0887630cf2c4d9cbd3f070537f40266ff0893/share/jupyter/voila/templates/reveal/index.html.j2

(voila-gridstack and voila-vuetify are ok).

Maybe we should do this in a coordinated way.

jwminton:
Yeah I was torn. The choice is to create a new macro file and then edit
both index.html.j2 files to include it (and possibly explain to anyone else
who's made their own index.html.j2 files?) OR just lump it into the
existing logging macro and move on with life. Personally, I would avoid
editing the index.html.j2 files until you need to add more into them and
just stick it in the logging macro with a good comment.

jwminton:
I think get it now. You're suggesting no new macro file but putting the required script element directly into the index.html.j2 files? The script element would still need to be pulled out of the voila_setup macro. If that is a preferred approach, I could update this PR with that instead. Given that there's some judgement and coordination involved, I'm completely open to whatever you would prefer to do next, just let me know what you think the next steps should be for the best result.

maartenbreddels:
Indeed, similar to the nbconvert templates. I'm wondering if this should go
into 0.3...

(from mobile phone)

jwminton:
I updated the PR to put the script elements directly into the index.html.j2 files and should have reverted the logging macro file to its previous state. I tried to replace the requires block about where it was originally (above the logging entries). Hopefully that works with overall template pattern. I tried to be minimal but it may need some tweaking for the final generated source to look perfect, if it matters. 

As you mentioned, the external index.html.j2 file in voila-reveal will need to have the same script element added.

In regards to doing this in a coordinated way, what do you suggest?

jwminton:
The new test [template_sanity_test.py](https://github.com/voila-dashboards/voila/blob/master/tests/app/template_sanity_test.py) is failing because it's testing for a leading slash in the src content and during the test execution the base_url is apparently empty causing the src content to look like this for both lab and classic output:
```html
  <script
    src="voila/static/require.min.js"
    integrity="sha256-Ae2Vz/4ePdIu6ZyI/5ZGsYnb+m0JlOmKPjt6XZ9JJkA="
    crossorigin="anonymous">
  </script>
```
So the assert fails:
```python
    assert '/voila/static/require.min.js' in html_text
```

It might be better to fix it in a separate PR, I'm not sure how you want to handle it, just let me know how you'd like to proceed.

maartenbreddels:
The failure was real, it had to be resources.base_url instead of base_url. Also took me a while to see that :)

SylvainCorlay:
LGTM.

maartenbreddels:
Thanks Jasen!

PR #741, opened 2020-10-19
fix: classic template did not pass base_url to macro

### LINKED ISSUE #734: Classic template bug
I get the following error when running a notebook in voila with the "classic" template when in directory mode (does not occur for single-notebook mode)
`WARNING:tornado.general:404 GET /voila/render/voila/static/require.min.js (172.20.0.1): No such file or directory: voila/static/require.min.js`

I can get rid of it if I pass the arguments `resources.base_url` and `resources.nbextensions` to `voila_setup`, following the pattern in the 'lab' template.
Line to modify:
https://github.com/voila-dashboards/voila/blob/b003a7fc62023e5b4c8dab7dd64b94a920610c15/share/jupyter/voila/templates/classic/index.html.j2#L46

I didn't look too much into the purposes of the different templates, but this seems like a bug

maartenbreddels:
Fixes #734

### DIFF
--- a/share/jupyter/voila/templates/classic/index.html.j2
+++ b/share/jupyter/voila/templates/classic/index.html.j2
@@ -43,6 +43,6 @@
 {% endblock body_header %}
 
 {% block footer_js %}
-{{ voila_setup() }}
+{{ voila_setup(resources.base_url, resources.nbextensions) }}
 {% endblock footer_js %}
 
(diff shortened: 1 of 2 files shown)


maartenbreddels:
many_iopub_messages on osx fails again, and rtd wants to install with python3.9, which fails

PR #753, opened 2020-10-29
fix: log cell execution errors

### LINKED ISSUE #752: Cell execution errors not displayed in log anymore
This has started since v0.2.0b0. I traced it back to https://github.com/voila-dashboards/voila/pull/681

mariobuikhuizen:
This was removed in 544ab3a. Fixes #752.


maartenbreddels:
Indeed, looks like this is the right place to put it! Thanks

PR #630, opened 2020-06-15
Add show_tracebacks trait

vidartf:
Fixes #616 .

Currently not tested as current tests fail on Windows 😢

PR #758, opened 2020-11-02
feat: allow showing of stacktraces in server extension mode

### LINKED ISSUE #751: When running as a server extension, no ability to display python errors with --debug
We have Voila running as a server extension.  When there is a python error, Voila displays "There was an error when executing cell [X].  Please run Voilà with --debug to see the error message."  We can customize this message with `cell_error_instruction`, but we can not show the python stack.  It seems that there is no way to run with the --debug flag when running as a sever extension, rather than running it directly.

maartenbreddels:
Closes #751
This is a followup of #630

@vidartf @: this is breaking in the sense of configuration `Voila.show_tracebacks` does not work anymore (if should now be `VoilaConfiguration.show_tracebacks`, *but* running voila with `voila --show_tracebacks=True` does still work (it is an alias). Should we consider this a breaking change, and up the minor version number (pre 1.0).
cc @aschlaep

Note that we could have skipped the `VoilaConfiguration.show_tracebacks` and only use `VoilaExecutor.show_tracebacks`, but there will be other code paths in the future that will also need to react to this (more on this later). So I think a better place for configuration is `VoilaConfiguration`.

### DIFF
--- a/tests/app/show_traceback_test.py
+++ b/tests/app/show_traceback_test.py
@@ -1,0 +1,23 @@
+import pytest
+
+
+@pytest.fixture(params=[True, False])
+def show_tracebacks(request):
+    return request.param
+
+
+@pytest.fixture
+def voila_args(notebook_directory, voila_args_extra, show_tracebacks):
+    return ['--VoilaTest.root_dir=%r' % notebook_directory, f'--VoilaConfiguration.show_tracebacks={show_tracebacks}'] + voila_args_extra
+
+
+async def test_syntax_error(http_server_client, syntax_error_notebook_url, show_tracebacks):
+    response = await http_server_client.fetch(syntax_error_notebook_url)
+    assert response.code == 200
+    output = response.body.decode('utf-8')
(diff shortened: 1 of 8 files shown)


vidartf:
> @vidartf @: this is breaking in the sense of configuration `Voila.show_tracebacks` does not work anymore (if should now be `VoilaConfiguration.show_tracebacks`, _but_ running voila with `voila --show_tracebacks=True` does still work (it is an alias). 

TBH I've never really understood the split in the config (what are the rules for what belongs where, and why is it like that?). If you are asking me whether I'm actively using the setting, then no, not yet. If you are asking me whether I consider this to strictly be a breaking change, then yes. :)

maartenbreddels:
The 'rule' is, if it is something that applies to the app and the extension, it should go in VoilaConfiguration. If it only applies to the voila app (like `--autoreload`) it goes in VoilaApp. Since the server extension does not create the app, it cannot 'see' the configuration options of VoilaApp.

Issue #764, opened 2020-11-16
Performance: voila takes double the time to create widgets compared to Notebook


maartenbreddels:
Creating ~600 widgets using:
```python
import ipywidgets as widgets
def recursive(n):
    if n == 0:
        return widgets.Button(description='Hi')
    else:
        cls = (widgets.VBox if n % 2 else widgets.HBox)
        return cls([recursive(n-1), recursive(n-1)])
recursive(7)
```

Takes about 1.5-2.0 seconds to display in the notebook, while it takes over 4.0 seconds using Voila.

I think at least one problem lies in the way we update the widgets / sync the state. We ask each widget/comm separately for an update, which causes a 'busy', 'update' and 'idle' message, 3x more than needed. This causes in total 1800 websocket messages. I am also not sure why that seems so slow.

maartenbreddels:
https://github.com/jupyter/jupyter_client/pull/590 takes off 50% of the time.
~2.8 to ~1.8 seconds

maartenbreddels:
The 4 seconds mentioned above was with Chrome with the dev console open.

maartenbreddels:
Minor contributor (few %): https://github.com/ipython/traitlets/pull/638

maartenbreddels:
https://github.com/tornadoweb/tornado/pull/2955 is another 80msec per 1000 messages.

maartenbreddels:
adding `--Session.key= ` avoid signing of messages, which also takes off a few %.

maartenbreddels:
https://github.com/jupyter-widgets/ipywidgets/pull/3020
and
https://github.com/ipython/traitlets/pull/639

scrape of 30%

maartenbreddels:
A 5x improvement might be possible based on https://github.com/voila-dashboards/voila/pull/766 and https://github.com/jupyter-widgets/ipywidgets/pull/3021

jtpio:
@maartenbreddels is this issue still valid with #933 now merged and released in `0.2.11`?

pierrotsmnrd:
Hi everyone, 
I've tested this with Voilà `0.2.16` and I can confirm the issue is still valid. Rendering the snippet above with voilà is taking at least twice more time.

martinRenou:
Things should be improved with the coming ipywidgets release and https://github.com/voila-dashboards/voila/pull/766

pierrotsmnrd:
Hi everyone :) 
Any update on this issue and the next ipywidgets release ? 
[...]

PR #766, opened 2020-11-19
Feature: fetch all widgets in one single comm message using the control channel

maartenbreddels:
Related to https://github.com/voila-dashboards/voila/issues/764

The way we fetch the widgets from the a kernel, is that we request all comm id's, and send an 'update' message to each comm to request the widget state. That means that for each widget, we send 1 message (update) and we receive 3 messages (busy, comm, idle). This leads to bad performance, since there is a per-message overhead in all layers (kernel, jupyter server, browser). On top of that, we don't know if the Comms are actually widgets.

I think we should have some higher level of communication, to talk 'widget' instead of 'comm', through a single comm object. This will allow us to request the full widget state in 1 request, and get the full state as 1 reply (or multiple if we want to).

Performance results:
Fetching of 1000 Layout widgets from the kernel
 * Before: 2500msec
 * After: 500msec

That's a 5x increase in performance, and *I think* much less load on the voila/jupyter_server server.

Note that this PR only downloads the widget data, it does not build the widgets yet.

### DIFF
--- a/packages/voila/src/manager.ts
+++ b/packages/voila/src/manager.ts
@@ -194,7 +194,123 @@ export class WidgetManager extends JupyterLabManager {
     });
   }
 
+  /**
+   * This is the implementation of building widgets models making use of the
+   * jupyter.widget.control comm channel
+   */
   async _build_models(): Promise<{ [key: string]: base.WidgetModel }> {
+    const models: { [key: string]: base.WidgetModel } = {};
+    const commId = base.uuid();
+    const initComm = await this._create_comm(
+      'jupyter.widget.control',
+      commId,
+      { widgets: null },
+      { version: '1.0.0' }
+    );
+
+    // Fetch widget states
+    let data: any;
+    let buffers: any;
+    try {
+      await new Promise((resolve, reject) => {
+        initComm.on_msg(msg => {
+          data = msg['content']['data'];
+
+          if (data.method !== 'update_states') {
+            console.warn(`
+              Unknown ${data.method} message on the Control channel
+            `);
+            return;
+          }
+
+          buffers = (msg.buffers || []).map((b: any) => {
+            if (b instanceof DataView) {
+              return b;
+            } else {
+              return new DataView(b instanceof ArrayBuffer ? b : b.buffer);
+            }
+          });
+
+          resolve(null);
+        });
+
+        initComm.on_close(reject);
+
+        // Send a states request msg
+        initComm.send({ method: 'request_states' }, {});
+      });
+    } catch (error) {
+      console.warn(
+        'Failed to open "jupyter.widget.control" comm channel, fallback to slow fetching of widgets.',
+        error
+      );
+      // Fallback to the old implementation for old ipywidgets versions (<=7.6)
+      return this._build_models_slow();
+    }
+
+    initComm.close();
+
+    const states: any = data.states;
+
+    // Extract buffer paths
+    // Why do we have to do this? Is there another way?
+    const bufferPaths: any = {};
+    for (const bufferPath of data.buffer_paths) {
+      if (!bufferPaths[bufferPath[0]]) {
+        bufferPaths[bufferPath[0]] = [];
+      }
+      bufferPaths[bufferPath[0]].push(bufferPath.slice(1));
+    }
+
+    const widgetPromises: Promise<base.WidgetModel>[] = [];
+
+    // Start creating all widgets
+    for (const [widget_id, state] of Object.entries(states) as any) {
+      try {
+        const comm = await this._create_comm('jupyter.widget', widget_id);
+
+        // Put binary buffers
+        if (widget_id in bufferPaths) {
+          const nBuffers = bufferPaths[widget_id].length;
+          base.put_buffers(
+            state,
+            bufferPaths[widget_id],
+            buffers.splice(0, nBuffers)
+          );
+        }
+
+        const modelPromise = this.new_model(
+          {
+            model_name: state.model_name,
+            model_module: state.model_module,
+            model_module_version: state.model_module_version,
+            model_id: widget_id,
+            comm: comm
+          },
+          state.state
+        );
+        widgetPromises.push(modelPromise);
+      } catch (error) {
+        // Failed to create a widget model, we continue creating other models so that
+        // other widgets can render
+        console.error(error);
+      }
+    }
+
+    // Wait for widgets to be created
+    const widgets = await Promise.all(widgetPromises);
+    for (const model of widgets) {
+      models[model.model_id] = model;
+    }
+
+    return models;
+  }
+
+  /**
+   * This is the old implementation of building widgets models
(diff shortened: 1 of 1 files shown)


martinRenou:
Adding the widget model creation to the PR.

Performance results:
Fetching of 400 Button widgets from the kernel

- Before: 4200msec
- After: 1000msec

martinRenou:
The model creation works now (no support for binary buffers yet), but passing all messages through one comm doesn't seem to work, I need to explore more on this.

maartenbreddels:
I'm afraid this is more difficult, since we may start listening after say 20% of the widgets are created, and if we want to create a widget while we missed the creation of a previously created widget that we depend on, we'll enter a 'deadlock' : We'll await the creation of our depending widget, and during that time we will not process any other message, since we 'await through the whole callstack', and the incoming websocket messages just accumulate. This means that we also cannot ask for the state of a previously created widget.

I hope my description makes sense, let me know if want need a video call for some higher bandwidth channel.

martinRenou:
I am not sure I understand your description 😅 getting into a call could help indeed.

What I am exploring now is to use your `jupyter.widget.control` target **only** for the first message, and then fake a `comm-open` message for all widgets so that all comms get created on the front-end, and if I understand correctly this should create the widget models.
This way the widgets should behave normally I believe, and we would still use the "one comm=one widget" logic.

martinRenou:
@maartenbreddels it seems to work with the following approach:

- One "single shot" comm to fetch all models at once
- Loop over all the received states and: 
  - Create model
  - Create comm for this model


Performances are still good:
Fetching of 400 Button widgets from the kernel

Before: 4400msec
After: 800msec

maartenbreddels:
I am also afraid this will go over the 10mb per websocket message. I guess this is still an issue

martinRenou:
Oh I didn't know this was a limitation. Don't websockets send messages in batches?

martinRenou:
@maartenbreddels I went a bit far testing this 😅 I created a grid of 1225 bqplot plots, each plot being a 1000 points line. The entire websocket message was 23_446_769 bytes, and it worked!

maartenbreddels:
Haha, cool! 
I don't know what the status is now, but this used to be a problem https://github.com/jupyter/notebook/issues/3468

martinRenou:
Oooh it's a tornado limitation, I see... 

Well, the good thing is we can bypass this with `c.NotebookApp.tornado_settings = {"websocket_max_message_size": x}`, we could maybe document this in the Voila documentation.

packages/voila/src/manager.ts
```diff
@@ -194,7 +194,118 @@ export class WidgetManager extends JupyterLabManager {
...
+        initComm.on_close(reject);
+
+        // Send a states request msg
+        initComm.send({method: 'request_states'}, {});
+      });
+    }
+    catch {
+      console.warn('Failed to open "jupyter.widget.control" comm channel, fallback to slow fetching of widgets.');
+      // Fallback to the old implementation for old ipywidgets versions (<=7.6)
+      return this._build_models_slow();
```
martinRenou, line 244: If the backend for ipywidgets does not implement the control channel, we fallback to the old implementation

martinRenou:
This PR could technically be merged already. Because it falls back to the previous implementation if the control channel is not responding.

martinRenou:
Marking this PR as ready to review.

This can be merged and released before ipywidgets supports the control channel in an official release. This way it will work as soon as ipywidgets 8 or 7.7 is out

packages/voila/src/manager.ts
```diff
@@ -194,7 +194,125 @@ export class WidgetManager extends JupyterLabManager {
...
+   * jupyter.widget.control comm channel
+   */
   async _build_models(): Promise<{ [key: string]: base.WidgetModel }> {
+    const models: { [key: string]: base.WidgetModel } = {};
+    const commId = base.uuid();
+    const initComm = await this._create_comm(
+      'jupyter.widget.control',
+      commId,
+      { widgets: null },
+      { version: '1.0.0' }
```
trungleduc, line 208: Can we get the comm name and version from `ipywidgets` instead of hard coding it?

packages/voila/src/manager.ts
```diff
@@ -194,7 +194,125 @@ export class WidgetManager extends JupyterLabManager {
...
+          resolve(null);
+        });
+
+        initComm.on_close(reject);
+
+        // Send a states request msg
+        initComm.send({ method: 'request_states' }, {});
+      });
+    } catch {
+      console.warn(
```
trungleduc, line 243: It might be useful to display the error message along with the warning.

packages/voila/src/manager.ts
martinRenou: This version is the version of the control channel protocol that Voila supports. If ipywidgets's control channel protocol changes to 2.0.0, we very likely need to update Voila manually and we don't want Voila to advertise that it supports 2.0.0 if it doesn't.

packages/voila/src/manager.ts
```diff
@@ -194,7 +194,125 @@ export class WidgetManager extends JupyterLabManager {
...
+        bufferPaths[bufferPath[0]] = [];
+      }
+      bufferPaths[bufferPath[0]].push(bufferPath.slice(1));
+    }
+
+    const widgetPromises: Promise<base.WidgetModel>[] = [];
+
+    // Start creating all widgets
+    for (const widget_id in states) {
+      const state = states[widget_id];
```
trungleduc, line 268:
```suggestion
    for (const [widget_id, state] of Object.entries(states)) {
```

packages/voila/src/manager.ts
```diff
@@ -194,7 +194,125 @@ export class WidgetManager extends JupyterLabManager {
...
+
+        // Put binary buffers
+        if (widget_id in bufferPaths) {
+          const nBuffers = bufferPaths[widget_id].length;
+          base.put_buffers(
+            state,
+            bufferPaths[widget_id],
+            buffers.slice(0, nBuffers)
+          );
+          buffers = buffers.slice(nBuffers);
```
trungleduc, line 281:
```suggestion
            buffers.splice(0, nBuffers)
          );
```
I'm not sure about the performance hit so feel free to discard this suggestion

packages/voila/src/manager.ts
trungleduc: Thanks for the explanation!

martinRenou:
Thanks for your review!

trungleduc (approved):

martinRenou:
The ui-tests are failing but I guess it's only due to the ipympl update: https://github.com/voila-dashboards/voila/pull/1048

martinRenou:
Thanks!!

maartenbreddels:
Btw, the reason we don't see this tornado limit is that it's a limit for incoming traffic, such that clients cannot pump the server full of terrabytes of data which it is forced to keep into memory. Downloading has no limit, since there the server is more in control.

PR #768, opened 2020-11-19
WIP feat: progressively load the widgets while the kernel is executing

maartenbreddels:
This will connect to the kernel as soon as possible, and tries to create
widget during execution. This avoids having to get all the widgets after
execution, and is an alterative to:
https://github.com/voila-dashboards/voila/pull/766

Issues:
 * since the whole jupyter message handling is 'synchronous' it awaits
   at a top level and goes all the way down, get_model needs to be resolved
   without any new message coming from the kernel.
 * Despite making handle_comm_open not returning a promise, we still get
   a stalling of the whole message q.

Despite these issues, the loading time is significantly reduced, since
the page renders almost instantly when it succeeds (e.g. when no widgets
are missed).

### DIFF
--- a/js/src/manager.js
+++ b/js/src/manager.js
@@ -62,6 +62,7 @@ export class WidgetManager extends JupyterLabManager {
             mimeTypes: [WIDGET_MIMETYPE],
             createRenderer: options => new WidgetRenderer(options, this)
         }, 1);
+        this.missed_models = {}
         this._registerWidgets();
         this.loader = requireLoader;
     }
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Note to self, we can do this a different way:
 * create the comms at the client, but don't create the widgets, store the msg that was send when the widget was created.
 * when an update msg gets send, the easy way would be to drop the last msg.
 * when we build the widgets, ask for the comms, and see what we missed, and only ask updates for those widgets.

PR #771, opened 2020-11-26
chore: show a user friendly error msg when an error occurs

maartenbreddels:
I'm not sure if we'd like to explain a bit more besides showing the error, especially when the stacktrace is shown it is a but 'crude' to users.

In any case, this is an example for @aschlaep on how one can add some custom error msg to a template.


maartenbreddels:
Using the html details and summary tags, it looks like (default collapsed):
image

PR #774, opened 2020-12-02
Fix stray tag in the lab template

jtpio:
Noticed with the `voila-gridstack` template which is using `lab` as the base template: https://github.com/voila-dashboards/voila-gridstack/blob/23d747087aaff5007769074b7ac780c59595a71b/share/jupyter/nbconvert/templates/gridstack/conf.json#L1

image

Moving the `voila_setup` above to the `body_footer` block should place the script before the `</body>` closing tag:

image


jtpio:
cc @maartenbreddels since if you are currently looking at an issue with the templates.

PR #775, opened 2020-12-02
Remove .travis.yml

### LINKED ISSUE #556: Remove Travis
Travis is slow. This build for example stalled 1 hr 44 min before failing: https://travis-ci.org/github/voila-dashboards/voila/builds/661178167

Since we have been running on GitHub Actions for a while now, we should be able to drop Travis completely.

jtpio:
Fixes #556


jtpio:
cc @SylvainCorlay @maartenbreddels @martinRenou 

I remember there were some concerns to fully switch to GitHub Actions when we first started to use them for the repo. However it has now been a couple of months and although random failures might still happen, they seem to be happening less often than on Travis.

Dropping Travis would also help maintain only one CI workflow.

jtpio:
For the record, we discussed this over video and there was no objection to dropping Travis.

So let's do it.

PR #789, opened 2020-12-22
Test on Python 3.9

jtpio:
Python 3.9 has been out for a while so we should test against it too on CI.


jtpio:
Ah looks like `traitlets==4.3.3` is not noarch on conda:

image

Only starting from `5.0.0`.

jtpio:
Looks good now after removing `traitlets=4.3.3` from the `mamba install` command, since `traitlets=4` is installed in the traitlets 4 CI workflow step.

jtpio:
All green, merging.

PR #794, opened 2021-01-05
Unpin jupyter_server in the tests

jtpio:
Unpin the `jupyter_server` dependency in the `test` `extras_require`, and define a `base_url` fixture instead.

Similar to https://github.com/voila-dashboards/voila-gridstack/pull/74


jtpio:
We should be able to reuse the Jupyter Server pytest fixtures after switching to `ExtensionApp`.

PR #797, opened 2021-01-07
Describe cell execution timeout in docs

### LINKED ISSUE #742: Execution timeout change
I just noticed the default cell execution timeout changed from 30s in 0.1 to None in 0.2.  I realize this is inherited from upstream (https://github.com/jupyter/nbconvert/issues/791) and that I can set a timeout myself, but I thought it might be worth mentioning here.  From a UX perspective, at some point the user will get tired of watching the "Executing" spinner and wonder if something is wrong.  So, does it make sense for Voila to set its own default?  If no, that's fine (as discussed in the nbconvert thread, the "right" number might be different for everyone), but maybe it's worth calling attention to the configuration setting in the docs.

jeffyjefflabs:
Voila inherits this setting from nbconvert, and the default is no timeout.  This may be undesirable UX in a dashboard, since the user has no indication of whether something has gone wrong or if they just need to keep waiting longer for the notebook to execute and render.  Fixes #742.

### DIFF
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -328,3 +328,14 @@ To hide both the code cell and the output cell (if any) for every cell that has 
 
 You can use any tag you want but be sure to use the same tag name in the Voilà command.
 And please note that this functionality will only hide the cells in Voilà but will not prevent them from being executed.
+
+Cell execution timeouts
+=======================
+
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks!

Looks good on the preview: https://voila--797.org.readthedocs.build/en/797/customize.html#cell-execution-timeouts

image

jtpio (approved):

PR #799, opened 2021-01-08
Fix typo in #797

jeffyjefflabs:
Oops, sorry @jtpio, just noticed I had a triple `---` instead of double `--` in #797... here's a quick fix.  Sorry about that, not sure how I missed it.


jtpio (approved):
Thanks for catching it!

PR #801, opened 2021-01-10
[Doc] Fix Voilà endpoint in Binder config

### LINKED ISSUE #776: [Documentation] Configuration field for running Voila with Binder is misleading
## What
In [the documentation on how to configure Binder to run Voila](https://voila.readthedocs.io/en/stable/deploy.html#deployment-on-binder), step number 4 says 
> In Path to a notebook file, select URL and use the Voilà endpoint: /voila/render/path/to/notebook.ipynb \

I think the endpoint should be `voila/render/path/to/notebook.ipynb` (no leading`/`)

## Why
Both the current format and the one I'm mentioning work fine when specifying a tree (`/voila/tree/path/to/folder`). But the current format fails when specifying a notebook file (`/voila/render/path/to/notebook.ipynb`), where the one I'm mentioning does not.\
I'm not certain, but I think this is an issue on the Binder side.

It is also the format used in the [README](https://github.com/voila-dashboards/voila#) Binder button of this very repo. You can notice that on the URL the Binder button points to (https://mybinder.org/v2/gh/voila-dashboards/voila/stable?urlpath=voila%2Ftree%2Fnotebooks) as `voila%2Ftree%2Fnotebooks` has no leading `%2F`.

## How to reproduce
I have tested that with a repo of mine and made the 2 options available with a Binder button on each: https://github.com/thomas-bc/voila-examples
The former cause the redirection to lead to `<binder_url>/tree/voila/tree` which fails with a 404 Not Found. The latter leads to `<binder_url>/voila/tree`, which works as expected. 

If this is a something you want to go forward with, I can make a PR.

thomas-bc:
Fixes #776

As presented in #776, the Voilà endpoint (when using Voilà in Binder) can fail with a leading slash.
I simply propose to make the no-leading-slash version the proposed one in the official documentation.

_It isn't much, but I spent a couple minutes figuring it out.. so I hope this will save some time to other people in the future._

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -66,7 +66,7 @@ being used for reproducible research, making it an excellent fit for deploying V
 
 3. Go to `mybinder.org <https://mybinder.org>`__ and enter the URL of the repository.
 
-4. In ``Path to a notebook file``, select ``URL`` and use the Voilà endpoint: ``/voila/render/path/to/notebook.ipynb``
+4. In ``Path to a notebook file``, select ``URL`` and use the Voilà endpoint: ``voila/render/path/to/notebook.ipynb``
 
 5. Click ``Launch``.
 


jtpio (approved):
Thanks!

PR #818, opened 2021-01-27
Add TypeScript and tsconfig.json

### LINKED ISSUE #809: Update to TypeScript for the Voila frontend
This will be the follow-up item after https://github.com/voila-dashboards/voila/pull/808 is merged.

The idea is to do a pass on the Voila frontend: https://github.com/voila-dashboards/voila/tree/master/packages/voila, and convert it to TypeScript.

Additionally, it might be possible to also update some of the packages to the `3.0` version at the same time. If not, then it could be a separate PR.

jtpio:
Fixes #809 

Opening early so we can multiple passes on this if needed to gradually add typings.

### TODO

- [x] Add `typescript` and `tsconfig.json`
- [x] Rename files to `.ts`
- [x] Add typings
- [x] Fix ESLint warnings
- [x] Remove `allowJs` in `tsconfig.json`


jtpio:
cc @JohanMabille since you were also interested in looking into this.

This is the first step to move to TypeScript, and there are still quite a few `any` to fix (here or in follow-up PRs).

jtpio:
Another follow-up task would be to try updating to more recent packages.

JohanMabille:
Thanks for pinging me @jtpio I have been swamped with other tasks and could not start this one :s

SylvainCorlay:
@jtpio should we get this in?

jtpio:
Sure.

As mentioned above this might need one more pass, but can be iterated on in follow-up PRs.

SylvainCorlay:
Thanks!

I am looking at something that may conflict with this so, merging now!

PR #822, opened 2021-02-04
Log information about classic extension loading

SylvainCorlay:
Splitting the requirejs call and adding some informative logs.


jtpio:
For reference, this adds logs to the browser dev tools console that look like this:

image

jtpio (approved):
Thanks!

Issue #410, opened 2019-10-09
Access Query Params


derek-pyne:
We've started launching some epic Voila dashboards in our team. One big thing that we would like to do is click a link from one dashboard to open another dashboard with some parameters set. The usually case for this is for drilling down into a more detailed dashboard.

We could do this by adding query params to the link. 

This is doable inside Notebooks by accessing the url in a javascript cell like this:
```
%%javascript
IPython.notebook.kernel.execute("URL = '" + window.location + "'");
```

However this doesn't work in Jupyter Lab or Voila (https://github.com/jupyterlab/jupyterlab/issues/5888). Assuming Voila is using Jupyter lab behind the scenes?

Any ideas on how to open a link to a Voila dashboard with some data in params?

jtpio:
Thanks @derek-pyne for opening this issue!

Voila doesn't allow arbitrary code execution so this code snippet will not work.

This is slightly related to https://github.com/voila-dashboards/voila/issues/105 and https://github.com/voila-dashboards/voila/pull/201 too but more towards using them in notebooks.

Linking to the gitter conversation here for reference: https://gitter.im/QuantStack/Lobby?at=5d9f33c4fcf7602cc5657eb2

timkpaine:
https://github.com/voila-dashboards/voila/pull/218#issuecomment-553653123

maartenbreddels:
Solved in https://github.com/voila-dashboards/voila/pull/414

PR #835, opened 2021-02-19
http_header_envs option to include http headers as an env var

danlester:
Similar to the QUERY_STRING environment variable that is set on a new kernel instance, this PR adds an HTTP_HEADERS env var which contains specified HTTP headers in a json-encoded dict.

There is a new configuration option, e.g. pass on the command line:

```
--VoilaConfiguration.http_header_envs="['X-CDSDASHBOARDS-JH-USER','user-agent']"
```

and then any new Voilà kernel will have an env var called HTTP_HEADERS containing a JSON-encoded dict which includes any of the specified headers that were available. For example:

```
{'X-CDSDASHBOARDS-JH-USER': '{"kind": "user", "name": "dan", "admin": true}', 'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit...'}
```

The user-agent string could be used to provide alternative implementations or a warning message for an unsupported browser; the X-CDSDASHBOARDS-JH-USER value can be details of an authenticated user from [ContainDS Dashboards](https://github.com/ideonate/cdsdashboards) (in which case, Voilà should only be accessible via the appropriate JupyterHub proxies, and not directly).

This system augments some really useful existing features or discussions such as #414 and #410.


nariyar:
Dear @danlester 
Is this merge expected to happen anytime soon? This would be a huge help. Will unblock progress of our work as we needed exactly this. Thanks in advance..

maartenbreddels:
I'll take a look soon, I like the idea. Does this follow any standard, like
CGI?

(from mobile phone)

danlester:
@nariyar Thank you for your comment. I'm not a Voilà maintainer so can't merge it myself.

Marten would have to do it (which is why he said he'd take a look).

@maartenbreddels Thanks for taking a look - I'm not aware of any standard that is relevant here to be honest. I guess if you want it to be "similar to HTTP" in some sense, you might want the headers passed on in "Key: Value\n" form (with a blank line at the end...) instead of JSON, but maybe that's a bit weird for env vars really.

I think the most important thing is ease of use on the Voilà (notebook) side - we shouldn't expose Voilà authors to standards just because we think they are "right" from a technical level.

Maybe the easiest thing for the notebook author would be to access headers as individual env vars, e.g. X_HEADER_USER_AGENT and X_HEADER_ANYTHING_ELSE, as long as we can be consistent over naming. That would save the Voilà notebook having to JSON-decode.

mariobuikhuizen:
I've looked into the available standards for this (on request of @maartenbreddels) and found that CGI has specifications for HTTP headers: https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18. As previous changes (#414) to supply server information to the kernel also conformed to the CGI standard, it would make sense to continue applying those standards here too.

Note that HTTP header names are case-insensitive: https://datatracker.ietf.org/doc/html/rfc2616#section-4.2, so the comparison in `handler.py` should also be done in a case-insensitive way.

Would you have the time and inclination to make these changes @danlester?

danlester:
@mariobuikhuizen Hi there! In theory I would like to help, and will in due course. But not sure when I'll get on to it, so if it is important to anyone else they should please feel free to make a note here that they are working on it :)

mariobuikhuizen:
Thank you @DanielAristidou, I will start working on it.

jtpio:
Fixed by https://github.com/voila-dashboards/voila/pull/922

PR #723, opened 2020-09-25
Migrate from phosphor to lumino, maintain compat

timkpaine:

PR #840, opened 2021-02-22
Update to newer packages for the voila frontend

jtpio:
The voila frontend was still using old packages that date back to JupyterLab 1.0.

This change updates to newer versions, and updates the code to handle breaking changes.

Closes https://github.com/voila-dashboards/voila/pull/723


jtpio:
Test passing and working fine locally.

Let's get this in so it's easier to iterate on the voila frontend.

PR #820, opened 2021-01-28
feat: kernel pooling to reduce latency

maartenbreddels:
cc @havok2063 @timkpaine 

Inpired by #726, related #30 

This creates a pool of kernel, of size `pool_size.get(kernel_name, pool_size_default)` (default=1) that are always kept in the background to reduce latency, *not* max throughput or server load.

When the kernel is started, a file with `{prefix}/voila_kernel_pool_init_{kernel_name}.{ext}` (e.g. `/Users/maartenbreddels/.jupyter/voila_kernel_pool_init_python3.py`) will be executed in the kernel, and if the kernel language is `python`, also the modules listed in `python_imports` will be imported, the most commonly anticipated usecase. (Note: this all happens before any notebooks are considered).

The effect of executing code beforehand (mainly importing) is to speed up the execution of the notebook. This *does not reduce the load* since the same amount of code will be executed, just distributed differently through time.

To make sure the cwd and env arguments of `start_kernel` are effective, we have some python update code, so this will currently only work with python. (*Note: env and cwd issues will also affect kernel forking, or #726*)

Example notebook:
```python
import jdaviz
```

Before (>3 seconds):
```
$ voila notebooks/preheat.ipynb --no-browser --PoolMappingKernelManager.pool_size_default=0
$ time curl http://localhost:8866/
curl http://localhost:8866/  0.01s user 0.01s system 0% cpu 3.727 total
```

After (<0.7 seconds):
```
$ voila notebooks/preheat.ipynb --no-browser  --PoolMappingKernelManager.python_imports=jdaviz --PoolMappingKernelManager.pool_size_default=5
$ time curl http://localhost:8866/
curl http://localhost:8866/  0.01s user 0.01s system 2% cpu 0.609 total
```

We can think about upstreaming this in jupyter_server.

TLDR: this speeds up the time to show a notebook with voila from >3 seconds to <0.7 seconds by having a warm kernel, with preimported modules.

PR #841, opened 2021-02-23
feat: make the multi kernel manager configurable

maartenbreddels:
This replaces #820 so that the multi kernel manager can evolve separately from voila.
Example configuration (voila.json):
```json
{
    "VoilaConfiguration": {
        "multi_kernel_manager_class": "hotpot_km.pool.PoolMappingKernelManager"
    },
    "PoolMappingKernelManager": {
        "pool_size_default": 3
    }
}
```

The kernel manager of #820 will go into https://github.com/voila-dashboards/hotpot_km (TODO)

cc @vidartf  @havok2063

### DIFF
--- a/voila/configuration.py
+++ b/voila/configuration.py
@@ -8,7 +8,7 @@
 #############################################################################
 
 import traitlets.config
-from traitlets import Unicode, Bool, Dict, List, Int, Enum
+from traitlets import Unicode, Bool, Dict, List, Int, Enum, Type
 
 
 class VoilaConfiguration(traitlets.config.Configurable):
@@ -85,3 +85,11 @@ class VoilaConfiguration(traitlets.config.Configurable):
     show_tracebacks = Bool(False, config=True, help=(
         'Whether to send tracebacks to clients on exceptions.'
     ))
+
+    multi_kernel_manager_class = Type(
+        default_value='jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager',
+        klass='jupyter_client.multikernelmanager.MultiKernelManager',
+        help="""The kernel manager class. This is useful to specify a different kernel manager,
+        for example a kernel manager with support for pooling.
+        """
+    ).tag(config=True)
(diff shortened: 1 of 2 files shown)


voila/configuration.py
```diff
@@ -85,3 +85,8 @@ class VoilaConfiguration(traitlets.config.Configurable):
     show_tracebacks = Bool(False, config=True, help=(
         'Whether to send tracebacks to clients on exceptions.'
     ))
+
+    multi_kernel_manager_class = Type(
+        default_value='jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager',
+        klass='jupyter_client.multikernelmanager.MultiKernelManager'
```
jtpio, line 91: Should there be a small `help` string similar to the other traits?

maartenbreddels: Good question, I considered it an expert feature, and the name self-describing, do you have a suggestion that adds information?

jtpio: Maybe something short like this is enough:

```python
help="""The kernel manager class. This is useful to specify a different kernel manager,
        for example a kernel manager with support for pooling.
        """
```

voila/configuration.py
```diff
@@ -85,3 +85,11 @@ class VoilaConfiguration(traitlets.config.Configurable):
...
         'Whether to send tracebacks to clients on exceptions.'
     ))
+
+    multi_kernel_manager_class = Type(
+        default_value='jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager',
+        klass='jupyter_client.multikernelmanager.MultiKernelManager',
+        help="""The kernel manager class. This is useful to specify a different kernel manager,
+        for example a kernel manager with support for pooling.
+        """
+    ).tag(config=True)
```
vidartf, line 95: `config=True` should be a kwarg, not a tag (the current behavior will work but is deprecated).

vidartf (approved):
Looks good other than one lint 👍

voila/configuration.py
maartenbreddels: Hmm yes, I see we do that a lot, lets fix that in a different PR ok?

jtpio (approved):
Thanks!

jtpio:
The `config=true` comment above is being addressed in https://github.com/voila-dashboards/voila/pull/847

PR #846, opened 2021-02-26
Switch to a lab-based app for the Voila frontend

jtpio:
Switch to a lab-based app for the Voila frontend.

[Binder](https://mybinder.org/v2/gh/jtpio/voila/lab-based?urlpath=voila/tree)

Closes #4.
Closes https://github.com/voila-dashboards/voila/pull/238

Might fix https://github.com/voila-dashboards/voila/issues/904

This replaces the current `requirejs` + `nbextensions` approach by a JupyterLab plugin based application. 

### Current Status

Basic rendering of widgets, still a couple of issues to fix:

image

### TODO

- [x] Drop `requirejs` -> keep `require.min.js` for `reveal` compatibility
- [x] Remove require based `main.js` file in the template
- [x] Expose the list of federated extensions (from `/share/jupyter/labextensions` for now)
- [x] Basic support for mimerenderer extensions
- [x] Add `beforeunload` logic in a plugin
- [x] Set `page_config`
- [x] Produce a single bundle? -> keeping a main `voila.js` and the the other chunks like in jupyterlab for now
- [x] Check voila only serves what is needed
- [x] Fix services not provided by the host Voila application
- [x] ~Remove dependency on `jupyterlab_server` if it is not strictly needed~ -> for now `jupyterlab_server` is only used for building the list of federated extensions with `get_page_config()`
- [x] Disable the default kernel and session polling with a plugin
- [x] Document the setup: https://voila--846.org.readthedocs.build/en/846/contribute.html#about-the-voila-frontend
- [x] Use the next `@jupyter-widgets/jupyterlab-manager@4.0.0-alpha.1` packages, to switch to a `KernelWidgetManager`: https://github.com/jupyter-widgets/ipywidgets/blob/1a121620aea08b2df4bcf92e851bd961317dc1db/jupyterlab_widgets/src/manager.ts#L409
- [x] ~[SKIP] Check how to keep the rate limiting logic~ -> will not do in this PR as this might not be needed based on the comment in https://github.com/voila-dashboards/voila/pull/846#issuecomment-998033258
- [ ] Depend on https://github.com/jupyter-server/jupyter_server_mathjax (https://github.com/voila-dashboards/voila/pull/846#pullrequestreview-649094759)

### Next steps

- Load JupyterLab extensions from the CDN? Requires upstream feature for dynamic extensions?

- Support JupyterLab themes, as an alternative or in addition to the theme and light themes from `nbconvert`: https://github.com/jupyter/nbconvert/blob/957c2e2476e1106fd4976af7d104c01cd5bb7562/setup.py#L52-L56


jtpio:
Support for third-party widgets built as a prebuilt extension for JupyterLab 3.0:

image

packages/voila/webpack.config.js
```diff
@@ -42,14 +60,28 @@ var distRoot = path.resolve(
...
+    plugins: [
+      new ModuleFederationPlugin({
+        library: {
+          type: 'var',
+          name: ['_JUPYTERLAB', 'CORE_LIBRARY_FEDERATION']
+        },
+        name: 'CORE_FEDERATION',
+        shared: {
+          ...data.dependencies
+        }
```
jtpio, line 83: This might be revisited at some point, based on the discussion and discussions in https://github.com/jupyterlab/jupyterlab/issues/9829

jtpio:
Testing with prebuilt mimerender extensions for JupyterLab 3.0, for example fasta:

image

bollwyvl:
It's looking really promising :tada: ! Some thoughts on things in the description:

> - [ ] Remove dependency on `jupyterlab_server` if it is not strictly needed

It seems like the localization, themes, and the (hopefully) upcoming licenses, would be worth the cost of the dependency. I also imagine workspaces being a fairly killer feature at some point, perhaps with a locked down DockPanel (e.g. hidden tabs if there aren't any, fixed aspect ratios). Settings would be... complicated.

> - [ ] Load JupyterLab extensions from the CDN? Requires upstream feature for dynamic extensions?

Whatever happens, a public CDN requirement (or having to run one's own, unless _shipped_ as part of this tool, or otherwise easily installed and served from the same host) is going to be a hard pass for a lot of deployment scenarios...

jtpio:
Thanks!

> It seems like the localization, themes, and the (hopefully) upcoming licenses, would be worth the cost of the dependency. I also imagine workspaces being a fairly killer feature at some point, perhaps with a locked down DockPanel (e.g. hidden tabs if there aren't any, fixed aspect ratios). Settings would be... complicated.

Probably it would make sense to depend on it, since it would make the Voila frontend a lot more like JupyterLab and easier to reuse extensions too. We just need to make sure it doesn't open the door to some misuses (by adding more endpoints).

For the settings the lab `ThemeManager` depends on it for now: 

https://github.com/jupyterlab/jupyterlab/blob/03d1edc02feb46bf7b7755d5694d0be5f7457a7d/packages/apputils/src/thememanager.ts#L488-L491

It might be possible to make it optional, so the `@jupyterlab/apputils-extension:themes` plugin could be reused as is. 
Apart from the themes, settings might not be needed and might not be useful for Voila dashboards anyway.

bollwyvl:
Re: endpoints: the addition of the handler (and adding the info to PageConfig, etc) is gated by whether the URL is set:
https://github.com/jupyterlab/jupyterlab_server/blob/master/jupyterlab_server/handlers.py#L177

This could probably be made even more explicit and regular... it would likely be an even better approach to break up that function into something with many smaller methods that can be overridden by the implementation. Made [an issue](https://github.com/jupyterlab/jupyterlab_server/issues/167)!

bollwyvl:
Also, I generally try to respect the draft status, and really appreciate you sharing... but if there's anything I can do to help, please let me know!

jtpio:
> Made [an issue](https://github.com/jupyterlab/jupyterlab_server/issues/167)!

Thanks for opening this!

> Also, I generally try to respect the draft status, and really appreciate you sharing... but if there's anything I can do to help, please let me know!

Yes the idea of opening the PR early is to get a clearer idea of potential issues, and also send upstream fixes early if needed. For now it would be great to try to go through the list of todo items in the top post, and reach parity with the existing frontend (with `enable_labextensions=True` but not necessarily loading from a CDN for now)

jtpio:
> This could probably be made even more explicit and regular... it would likely be an even better approach to break up that function into something with many smaller methods that can be overridden by the implementation

Another thing we might want to revive soon is the switch to an `ExtensionApp`: https://github.com/voila-dashboards/voila/pull/592

Which might help streamline the setup and reuse bits of other ExtensionApps. Although in that case the Voila ExtensionApp would be configured with `load_other_extensions = True`

packages/voila/src/manager.ts
```diff
@@ -7,93 +7,47 @@
...
  ****************************************************************************/
 
-import {
-  WidgetManager as JupyterLabManager,
-  WidgetRenderer
-} from '@jupyter-widgets/jupyterlab-manager';
+import { WidgetRenderer, output } from '@jupyter-widgets/jupyterlab-manager';
 
-import { output } from '@jupyter-widgets/jupyterlab-manager';
+import { KernelWidgetManager } from '@jupyter-widgets/jupyterlab-manager/lib/manager';
```
jtpio, line 12: Considering exporting this from the upstream package in https://github.com/jupyter-widgets/ipywidgets/pull/3166

voila/handler.py
```diff
@@ -30,7 +31,37 @@
...
+        page_config = {
+            "appVersion": __version__,
+            "baseUrl": self.base_url,
+            "terminalsAvailable": False,
+            "fullStaticUrl": url_path_join(self.base_url, "voila/static"),
+            "fullLabextensionsUrl": url_path_join(self.base_url, "voila/labextensions"),
+        }
+
+        mathjax_config = self.settings.get("mathjax_config", "TeX-AMS_HTML-full,Safe")
+        # TODO Remove CDN usage.
```
vidartf, line 45: You should be able to depend on this: https://github.com/jupyter-server/jupyter_server_mathjax

bollwyvl:
Raised over on https://github.com/jtpio/jupyterlite/discussions/202:

What if the `voila` frontend was "just" an built-on-demand `jupyterlite` app/site? It seems like many of the features we want over there (CDN independence, reproducible builds, service workers/offline support) would be highly desirable features here.

jtpio:
> What if the `voila` frontend was "just" an built-on-demand `jupyterlite` app/site?

Yes that sounds like a good way to go to easily be able to reuse components between the two projects. This should also be easier once we start publishing the `@jupyterlite/*` packages to npm.

packages/voila/src/plugins.ts
```diff
@@ -0,0 +1,124 @@
...
+  },
+  autoStart: true,
+  provides: JupyterFrontEnd.IPaths
+};
+
+/**
+ * A plugin to stop polling the kernels, sessions and kernel specs.
+ *
+ * TODO: a cleaner solution would involve a custom ServiceManager to the VoilaApp
+ * to prevent the default behavior of polling the /api endpoints.
```
jtpio, line 43: This plugin is temporary and should ideally be removed.


Two options for moving forward with this:

- submit upstream fixes to the JupyterLab `ServiceManager`, for example by making some of the services optionals (maybe a `ServiceManager` shouldn't require having a kernel manager)
- reuse the work or approach done in [JupyterLite](https://github.com/jupyterlite/jupyterlite), ideally also highly configurable

packages/voila/src/manager.ts
jtpio: https://github.com/jupyter-widgets/ipywidgets/pull/3166 has been merged and released so we should be able to update the import.

jtpio:
Squashing the previous commits to make subsequent rebases easier.

martinRenou:
> Check how to keep the rate limiting logic

This should not really be critical anymore once ipywidgets 8 (or 7.7) is out and https://github.com/voila-dashboards/voila/pull/766 is in. So we can probably just ignore this?

jtpio:
> This should not really be critical anymore once ipywidgets 8 (or 7.7) is out and #766 is in. So we can probably just ignore this?

That's good news. Looks like we could indeed ignore this then in this PR (will remove that from the todo list).

jtpio:
Linking to the upstream PR for reference: https://github.com/jupyterlab/jupyterlab/pull/11693. Which should help reuse some boilerplate code for loading federated extensions.

martinRenou:
Rebased.

We need to upstream https://github.com/voila-dashboards/voila/pull/766 into ipywidgets's `LabWidgetManager`, otherwise we'll loose this improvement.

And concerning your comment:

> Check how to keep the rate limiting logic

We should probably also upstream this.

martinRenou:
> We need to upstream #766 into ipywidgets's LabWidgetManager, otherwise we'll loose this improvement.

See https://github.com/jupyter-widgets/ipywidgets/pull/3335

jtpio:
> We should probably also upstream this.

I think this is the related PR: https://github.com/jupyter-widgets/ipywidgets/pull/2765

martinRenou:
Thanks for the pointer, I'll see if I can revive it

jtpio:
Nice thanks. In https://github.com/jupyter-widgets/ipywidgets/pull/2765#issuecomment-808366149 I attempted a rebase but it's probably outdated by now. So yeah reviving it would be great :+1:

jtpio:
Also as a heads-up, some upstream PRs such as the following should make it easier to reuse the boilerplate code to enable federated extensions: https://github.com/jupyterlab/jupyterlab/issues/11633

packages/voila/src/plugins.ts
jtpio: We should now be able to update to the `4.0.0-alpha.x` JupyterLab packages, which include the change from https://github.com/jupyterlab/jupyterlab/pull/11537 to allow for mock services.

packages/voila/src/app.ts
```diff
@@ -0,0 +1,151 @@
+import {
+  JupyterFrontEnd,
+  JupyterFrontEndPlugin
+} from '@jupyterlab/application';
+
+// TODO: export from upstream?
+import { createRendermimePlugins } from '@jupyterlab/application/lib/mimerenderers';
```
martinRenou, line 7:
```suggestion
import { createRendermimePlugins } from '@jupyterlab/application';
```

martinRenou: https://github.com/jupyterlab/jupyterlab/pull/10117 for reference

jtpio: Yes. This might also need a bump of the `@jupyterlab` packages too (not sure in which `3.1.x` this was released upstream).

We can also update to `4.0.0-alpha.x` directly to grab other fixes as mentioned in https://github.com/voila-dashboards/voila/pull/846#discussion_r777943940.

martinRenou: Would you mind if I push new commits to your branch for doing this?

jtpio: Sure, feel free!

packages/voila/index.js
```diff
@@ -0,0 +1,212 @@
...
+
+// Inspired by: https://github.com/jupyterlab/jupyterlab/blob/master/dev_mode/index.js
+
+import './style.css';
+
+import { PageConfig, URLExt } from '@jupyterlab/coreutils';
+
+import { VoilaApp } from '@voila-dashboards/voila';
+
+// Promise.allSettled polyfill, until our supported browsers implement it
```
jtpio, line 21: We removed this polyfill upstream in https://github.com/jupyterlab/jupyterlab/pull/11741.

So we can remove it here too if we consider we are targeting the same browsers.

packages/voila/src/app.ts
martinRenou: Thanks!

Actually I am having troubles using jupyterlab's `4.0.0-alpha.x`, as ipywidgets's `@jupyter-widgets/jupyterlab-manager` depends on `^3.0.0` and the tokens are conflicting.

Will it be impossible for me to use both jupyterlab's `4.0.0-alpha.x` and `@jupyter-widgets/jupyterlab-manager` because of this? Do I need to use a local version of `@jupyter-widgets/jupyterlab-manager` that depends on jupyterlab `4.0.0-alpha.x`?

jtpio: Which version of `@jupyterlab/services` is used? Normally it should be `7.0.0-alpha.x`

martinRenou: `@voila-dashboards/voila` depends on `@jupyterlab/services: 7.0.0-alpha.2`, but `@jupyter-widgets/base-manager@^1.0.0-beta.1` depends on `@jupyterlab/services: 6.0.0` (sorry my previous comment was missing this information)

jtpio: Ah ok then it is most likely related to this PR: https://github.com/jupyterlab/jupyterlab/pull/11537.

We could try to use `resolutions` to ensure there is a single version of `@jupyterlab/services` being used.

martinRenou:
Should we get rid of the `enable_labextensions` option and enable them by default?

If I recall correctly, the `enable_nbextensions` was originally introduced because it doesn't always make sense to load nbextensions (as some nbextensions were really meant for the classic Jupyter Notebook, adding new menus etc).

With this PR, it might make more sense to always load installed labextensions (themes, renderers etc).

jtpio:
Unfortunately the description of the PR that added the `enable_nbextensions` flag doesn't say much about the motivation behind that flag: https://github.com/voila-dashboards/voila/pull/185

> If I recall correctly, the enable_nbextensions was originally introduced because it doesn't always make sense to load nbextensions (as some nbextensions were really meant for the classic Jupyter Notebook, adding new menus etc).

This sounds plausible. Agree in that case it wouldn't be necessary for lab plugins. There were quite a few issues opened about the `enable_nbextensions` flag. So having it enabled by default would be a "good default".

martinRenou:
@SylvainCorlay maybe you have some background information about the purpose of `enable_nbextension` option?

martinRenou:
I made an (very hacky) attempt to make use of the JupyterLab theme manager and load the theme labextensions dynamically upon application start, you can see the diff here: https://github.com/jtpio/voila/compare/lab-based...martinRenou:theme_handling

As discussed in a meeting with @jtpio and @SylvainCorlay, it might be better to not take this approach yet. We could allow users to change the theme later on (with a dropdown button), but this is probably better to postpone such a feature for now.

Voila being very close to nbconvert (and actually reusing the Lab themes coming from nbconvert through the `light` or `dark` themes), we should probably change nbconvert so that it allows loading themes labextensions in the templates.

setup.cfg
```diff
@@ -34,6 +34,7 @@ python_requires = >=3.6
 install_requires =
     jupyter_server>=0.3.0,<2.0.0
     jupyter_client>=6.1.3,<8
+    jupyterlab_server>=2.3.0,<3
```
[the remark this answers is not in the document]
bollwyvl: I think there's a lot of value in getting reusable features into `juptyerlab_server`... optionally enable licenses, etc. Further, if it's all reuse-by-import (and not subclass) there's little risk of getting hard-to-reason-about stuff like settings and workspaces... but eventually, settings (at least readable ones with a no-op POST) would be super useful.

jtpio:
https://github.com/voila-dashboards/voila/pull/846/commits/3b7feee0b50157a8d32f7755d225e23cb0873195 dropped the `enable_labextensions` flag.

setup.cfg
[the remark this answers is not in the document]
jtpio: For reference https://github.com/voila-dashboards/voila/pull/1089 also adds `jupyterlab_server` as a dependency in its current state.

tests/app/nbextensions_test.py
```diff
@@ -22,6 +22,7 @@ def voila_config_file_paths_arg():
     return '--VoilaTest.config_file_paths=[%r]' % path
 
 
+@pytest.mark.skip(reason='TODO: update for JupyterLab extensions')
```
jtpio, line 25: Skipping this test for now since it is currently failing when trying to look up non existing model ids.

packages/voila/src/manager.ts
```diff
@@ -7,93 +7,47 @@
...
 
 import * as base from '@jupyter-widgets/base';
-import * as controls from '@jupyter-widgets/controls';
 
-import * as Application from '@jupyterlab/application';
-import * as AppUtils from '@jupyterlab/apputils';
-import * as CoreUtils from '@jupyterlab/coreutils';
-import * as DocRegistry from '@jupyterlab/docregistry';
-import * as OutputArea from '@jupyterlab/outputarea';
+import * as controls from '@jupyter-widgets/controls';
```
martinRenou, line 16: `controls` can probably be removed now?

packages/voila/package.json
```diff
@@ -8,47 +8,64 @@
   "main": "lib/index.js",
   "browserslist": ">0.8%, not ie 11, not op_mini all, not dead",
   "dependencies": {
-    "@jupyter-widgets/base": "^4.0.0",
-    "@jupyter-widgets/controls": "^3.0.0",
-    "@jupyter-widgets/jupyterlab-manager": "^3.0.0",
+    "@jupyter-widgets/base": "^5.0.0-alpha.3",
+    "@jupyter-widgets/controls": "^4.0.0-alpha.3",
```
martinRenou, line 12: The `@jupyter-widgets/controls` dependency can probably be removed?

martinRenou: Actually not possible in the current state

packages/voila/src/manager.ts
martinRenou: Actually not possible in the current state

jtpio:
Planning to do a rebase of this PR today.

jtpio:
So it's only the UI tests failing now:

image

With a couple more reference snapshots to update, but also real issues to fix:

image

jtpio:
Some of the failures yield a `TypeError: Cannot read properties of undefined (reading 'sessionContext')`.

For example with the `interactive` example:

image

jtpio:
> For example with the `interactive` example:

https://github.com/voila-dashboards/voila/pull/846/commits/ef544529d6672f848ea023a8f9236059b35d0ec6 seems to partially be fixing it (still lacking the output):

image

jtpio:
> Some of the failures yield a TypeError: Cannot read properties of undefined (reading 'sessionContext').

Investigating this a bit more, this seems to be because the `OutputModel` from `ipywidgets` for JupyterLab located [here](https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/python/jupyterlab_widgets/src/output.ts#L22) assumes it can access `context` from its `widget_manager` here: https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/python/jupyterlab_widgets/src/output.ts#L36

However in this PR we are switching to the `KernelWidgetManager` which doesn't have access to a document context, only to the kernel connection.

Manually changing the logic in the `@jupyter-widgets/jupyterlab-manager` under the local `node_modules` (just for testing for now) to handle that case seems to be fixing it.

image

I'll open a PR in the ipywidgets repo.

jtpio:
https://github.com/voila-dashboards/voila/pull/846/commits/38409998b5db79a6fb7bc78dba78325f2116b305 does fix the missing output. However the slider positions don't seem to be initialized correctly:

image

This is a behavior that is also happening in the other checks, like "":

jtpio:
> However the slider positions don't seem to be initialized correctly:

This is actually an issue in `ipywidgets` in JupyterLab and the Classic Notebook. Modifying the text and pressing enter does not change the position of the slider:

https://user-images.githubusercontent.com/591645/186756922-fa3b5b1d-3704-4659-9888-3cd38e0c4842.mp4

This was caught by the UI tests because they find the text input, modify it and press enter:

https://github.com/voila-dashboards/voila/blob/68ec2e949cd52f1eaae7871a0a4c3ec1ae91c3ef/ui-tests/tests/voila.test.ts#L112-L113

Probably introduced by the new implementation of the slider in ipywidgets 8. We can track the issue upstream. Depending on how fast this gets fixed and released we can just update the snapshot with the "wrong" position for now, and then update again later when the fix is out.

jtpio:
> Probably introduced by the new implementation of the slider in ipywidgets 8. We can track the issue upstream. Depending on how fast this gets fixed and released we can just update the snapshot with the "wrong" position for now, and then update again later when the fix is out.

Alright so the PR was merged and released. Just updated to the latest packages and most UI tests are now passing:

image

The `ipyvolume` seems to have changed, would need to double check if this is expected.

The only left is `reveal`, which is probably failing because `requirejs` has now been removed:

image

jtpio:
CI is now passing:

image

Although some tests have been marked as `skip` and should be investigated.

Posting some follow-ups here that we can track in a meta issue:

- [ ] Check why the `ipyvolume` reference snapshot had to be updated. Looks like it might be because of a debug mode?
- [ ] Update `ipyvolume` dev dependency to point to a released version of `ipyvolume` with support for ipywidgets 8
- [ ] Depend on https://github.com/jupyter-server/jupyter_server_mathjax ?
- [ ] Remove `require.min.js` (required for `reveal` compatibility)? This might require updating the reveal template to work without `requirejs`
- [ ] Remove the custom `OutputModel` if / when https://github.com/jupyter-widgets/ipywidgets/pull/3561 gets merged and released upstream
- [ ] Rich outputs (mime renders) should render on the first pass
- [ ] Update to `@jupyterlab` 4 packages to remove the plugin that stop the kernel and session polling. This might require using custom (no-op) kernel and session managers. 
- [ ] Add UI tests for the mimerenders notebook
- [ ] Investigate skipped tests

martinRenou:

jtpio:
Just rebased onto the latest `main` and squashed all the commits into one so it's easier to perform future rebases.

jtpio:
@trungleduc looks like the `version` field was adding back to `package.json` in your last rebase but I think we'll want to remove it in this PR to start publishing the `@voila-dashboards/voila` npm package:

https://github.com/voila-dashboards/voila/compare/352757771ff76069ae8b615b9f3c34032e59e627..b46c06bab23dd24463da81a91bedc881d97e3cd3#diff-3f623ad42a63daa4aed101035cca2f9f3843759d36b5eaeac80f35bb97fd51a9R3

packages/voila/src/output.ts
```diff
@@ -0,0 +1,98 @@
...
+export class OutputModel extends outputBase.OutputModel {
+  defaults(): Backbone.ObjectHash {
+    return { ...super.defaults(), msg_id: '', outputs: [] };
+  }
+
+  initialize(attributes: any, options: any): void {
+    super.initialize(attributes, options);
+    // The output area model is trusted since widgets are only rendered in trusted contexts.
+    this._outputs = new OutputAreaModel({ trusted: true });
+    this._msgHook = (msg): boolean => {
```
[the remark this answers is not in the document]
jtpio: Sure. Although this code comes from `ipywidgets` and should hopefully be removed soon when a fixes landed upstream.

See https://github.com/voila-dashboards/voila/pull/846#issuecomment-1222047852 for more details.

packages/voila/src/app.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * A namespace/prefix plugins may use to denote their provenance.
+   */
+  readonly namespace = this.name;
+
+  /**
+   * The version of the application.
+   */
+  readonly version = 'unknown';
```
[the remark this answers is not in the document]
jtpio: Probably the one from the `@voila-dashboards/voila` npm package. Which should already be kept in sync with the Python version.

packages/voila/src/plugins.ts
jtpio: Indeed. Not sure there is more to do for now if we stick to the JupyterLab 3 packages.

A follow-up would be to update to the JupyterLab 4 packages:

> We should now be able to update to the `4.0.0-alpha.x` JupyterLab packages, which include the change from [jupyterlab/jupyterlab#11537](https://github.com/jupyterlab/jupyterlab/pull/11537) to allow for mock services.

martinRenou:
For the Voilite work in JupyterLite, I need the `WidgetManager` do pass itself to the `createRenderer` factory as second argument:

```typescript
export class WidgetManager extends KernelWidgetManager {
  constructor(
    kernel: Kernel.IKernelConnection,
    rendermime: IRenderMimeRegistry
  ) {
    super(kernel, rendermime);
    rendermime.addFactory(
      {
        safe: false,
        mimeTypes: [WIDGET_MIMETYPE],
        createRenderer: options => new WidgetRenderer(options, this as any)
      },
      1
    );
```

This does not compile unless `as any` is used.

packages/voila/index.js
```diff
@@ -0,0 +1,189 @@
+/***************************************************************************
```
trungleduc, line 1: @jtpio can we convert this file into `typescript` and move it into the `src` directory?

jtpio: This file is inspired by the one in JupyterLab which corresponds to the entrypoint of the application. In the case of JupyterLab there is some templating with `handlebars` happening during the webpack build, which is probably why it can't be converted to TS there.

But it might be possible in the case of Voila yes.

trungleduc: Thanks! I will try to do it.

.binder/postBuild
```diff
@@ -6,4 +6,7 @@ jlpm && jlpm run build
 python -m pip install -e .
 jupyter labextension develop . --overwrite
 jupyter server extension enable voila.server_extension --sys-prefix
-jupyter serverextension enable voila.server_extension --sys-prefix
\ No newline at end of file
+jupyter serverextension enable voila.server_extension --sys-prefix
+
+# force installing ipywidgets 8
+python -m pip install --pre -U ipywidgets
```
martinRenou, line 12:
```suggestion
```

tests/execute_output_test.py
```diff
@@ -35,6 +37,7 @@ def normalize_outputs(outputs):
         normalize_output(output)
 
 
+@pytest.mark.xfail(reason='TODO: investigate this failing test')
```
martinRenou, line 40: Do we still need this?

```suggestion
```

martinRenou:
update galata references

tests/execute_output_test.py
martinRenou: It seems to still be needed

martinRenou:
I removed the ui-test with ipyvolume. We could probably reduce the number of ui-tests to the bare minimum. Checking that one custom widget extension works is already enough to check that Voila works properly with custom widgets, testing more can only lead to unreliable tests.

martinRenou (approved):
:shipit: :shipit: :shipit:

trungleduc (approved):

trungleduc:
Thanks to all for working on this awesome PR!

PR #847, opened 2021-03-02
Fix config tag deprecated

maartenbreddels:
Follow up of #841, where @vidartf noted that we use the deprecated `.tag(config=True)`


jtpio (approved):
Thanks!

Just need a quick rebase for https://github.com/voila-dashboards/voila/pull/841

PR #859, opened 2021-03-22
Wrap get_kernel with possible async handler

### LINKED ISSUE #858: Get kernel as async operation
A project I'm working on has overriden the KernelManager and the KernelSpecManager to make most of the methods async. Most calls to these managers have `ensure_async` wrappers, but this particular instance does not:
https://github.com/voila-dashboards/voila/blob/ae3d025dda21f72c2a663d04cf4ff10dc9215f08/voila/handler.py#L171

Is there any region that this couldn't be wrapped in the same way? I'll submit a PR and reference this issue.

declanvk:
Closes #858.

This change adds a wrapper around the `get_kernel` call, to account for a possibly async implementation.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -270,7 +270,7 @@ class VoilaHandler(JupyterHandler):
         kernelspec = notebook.metadata.kernelspec
         kernel_name = kernelspec.get('name', self.kernel_manager.default_kernel_name)
         # We use `maybe_future` to support RemoteKernelSpecManager
-        all_kernel_specs = await tornado.gen.maybe_future(self.kernel_spec_manager.get_all_specs())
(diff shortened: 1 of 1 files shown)


declanvk:
The second commit ([f8274ef](https://github.com/voila-dashboards/voila/pull/859/commits/f8274eff6be75be2a2b28c66d48f917d343d3f64)), is beyond the original scope of the PR and the linked issue, so I understand if this part is less likely to be merged. However, these were the full changes needed to make the async implementation of the KernelSpecManager work with voila.

SylvainCorlay:
Thank you. I think this is good to merge!

Ping @davidbrochart

PR #862, opened 2021-03-24
Run CI on all branches for push events

jtpio:
So it also runs when pushing to other branches like `0.1.x`.

Use `*` for all branches to avoid having to maintain an explicit list of branches.


jtpio:
Also in preparation for eventual `0.2.x` branch at some point.

PR #870, opened 2021-04-12
Update to font awesome 5

### LINKED ISSUE #869: Upgrade to fontawesome 5
Now that Voila 0.2.8 ships with a newer version of ipywidgets that includes https://github.com/jupyter-widgets/ipywidgets/pull/2793, I think we can upgrade this line from fontawesome 4 to fontawesome 5 with the fontawesome 4 shim:

https://github.com/voila-dashboards/voila/blob/7423f7c6197194b17a5108e4e4c1a1b9c08bbc67/share/jupyter/voila/templates/lab/index.html.j2#L20

jtpio:
Fixes #869

### DIFF
--- a/share/jupyter/voila/templates/lab/index.html.j2
+++ b/share/jupyter/voila/templates/lab/index.html.j2
@@ -17,7 +17,8 @@
 {%- endblock html_head_js -%}
 
 {%- block html_head_css -%}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/font-awesome@4.5.0/css/font-awesome.min.css" type="text/css" />
+<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.15.3/css/all.css" type="text/css" />
+<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.15.3/css/v4-shims.css" type="text/css" />
 {{ super() }}
 {{ spinner.css() }}
   <style>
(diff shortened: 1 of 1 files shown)


jasongrout:
This is similar to what we do in JupyterLab, except we use the minified versions: https://github.com/jupyterlab/jupyterlab/blob/1568a8811ef3f93ab8909890a2a5672ce5cb09cc/packages/application/style/index.css#L7-L8

jasongrout:
Are these files minified? If not, can we use minified versions?

jtpio:
Also the new version 5 doesn't seem to be served via `jsdelivr` (latest available version there is 4.7.0): https://www.jsdelivr.com/package/npm/font-awesome

So this change also switches to https://use.fontawesome.com, which is documented in https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4.

jtpio:
> Are these files minified? If not, can we use minified versions?

Although the links to include `.min` as part of the extension, they do seem to be minified:

image

jasongrout (approved):
Looks good to me.

jasongrout:
The jsdelivr urls are below, in case you do want to use the jsdelivr cdn after all:

https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5.15.3/css/all.min.css
https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5.15.3/css/v4-shims.min.css

jtpio:
Nice, thanks :+1: 

Looks like we could indeed stick to jsdelivr for consistency.

PR #880, opened 2021-04-30
success variable was undefined when no port can be found

danlester:
This fixes an uncaught error `UnboundLocalError: local variable 'success' referenced before assignment`


jtpio (approved):
Good catch, thanks!

Issue #620, opened 2020-05-26
Preview extension: switch to LabIcon for the refresh button


jtpio:
The switch to JupyterLab 2.0 was done in https://github.com/voila-dashboards/voila/pull/555.

One noticeable change in 2.x is the way icons are handled. Currently the dark theme is not handled properly (the refresh icon is too dim):

![image](https://user-images.githubusercontent.com/591645/82885444-7dc76200-9f45-11ea-8f3d-1546addfe6ed.png)

Here is an example PR that implements the switch to `LabIcon`: https://github.com/jupyterlab/debugger/pull/376

For the preview extension, this would mean importing the refresh icon with:

```typescript
import { refreshIcon } from '@jupyterlab/ui-components' ;
```

And using it here (with `icon` instead of `iconClass`): https://github.com/voila-dashboards/voila/blob/80b120c805fba3f3b33bc5b46f9eea52473b6f78/packages/jupyterlab-voila/src/preview.tsx#L65

jtpio:
Fixed by #884

PR #884, opened 2021-05-07
Preview extension: switch to LabIcon for the refresh button

ohrely:
Addresses https://github.com/voila-dashboards/voila/issues/620

Collaboration with @vickyprakash

### DIFF
--- a/packages/jupyterlab-preview/src/preview.tsx
+++ b/packages/jupyterlab-preview/src/preview.tsx
@@ -13,7 +13,7 @@ import {
 
 import { INotebookModel } from '@jupyterlab/notebook';
 
-import { refreshIcon } from '@juperlab/ui-components';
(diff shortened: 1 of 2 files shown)


packages/jupyterlab-preview/src/preview.tsx
```diff
@@ -13,6 +13,8 @@ import {
 
 import { INotebookModel } from '@jupyterlab/notebook';
 
+import { refreshIcon } from '@juperlab/ui-components';
```
ivanov, line 16: looks like a type here.
```suggestion
import { refreshIcon } from '@jupyterlab/ui-components';
```

ivanov (changes requested):

SylvainCorlay (approved):

jtpio (approved):
Thanks!

Issue #805, opened 2021-01-12
Automate releases


jtpio:
With https://github.com/voila-dashboards/voila/pull/798 now in place, we are one step away from automating releasing.

The missing step is to upload the wheel and sdist to PyPI, and the `@voila-dashboards/jupyterlab-preview` extension to npm (if it has changed).

Linking to some resources for reference and inspiration:

- https://github.com/vaexio/vaex/blob/master/.github/workflows/wheel.yml
- https://github.com/vaexio/vaex/blob/master/.github/workflows/wheel-universal.yml

cc @maartenbreddels who is also interested in this

jtpio:
We will probably want to use https://github.com/jupyter-server/jupyter_releaser directly once it's out of the experimental phase.

jtpio:
Closing as fixed.

`v0.2.13` was just released with the Jupyter Releaser: https://github.com/voila-dashboards/voila/releases/tag/v0.2.13

PR #912, opened 2021-06-23
Add CHANGELOG.md

### LINKED ISSUE #369: Add a changelog
We should add a CHANGELOG.md to list (or generate) the changes after each release (we already got some questions about things that were fixed in subsequent releases)

jtpio:
Fixes #369 

Bootstrap a `CHANGELOG.md` with a temporary script using `github-activity`.

This file will have to be continuously updated when new releases are cut (will be automated with #805). We can then also use the set of labels that `github-activity` understands to tag issues and PRs, so they are nicely summarized in the changelog when cutting a release.

## Changes

- [x] Create `CHANGELOG.md` 
- [x] Add missing releases
- [x] Make it compatible with the jupyter releaser

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,942 @@
+# Changelog
+
+<!-- <START NEW CHANGELOG ENTRY> -->
+
+## 0.2.10
+
+([full changelog](https://github.com/voila-dashboards/voila/compare/0.2.9...0.2.10))
+
+### Documentation improvements
+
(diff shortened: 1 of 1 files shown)


jtpio:
This initial generation is a bit tedious for now, but discussed in https://github.com/executablebooks/github-activity/issues/57

jtpio:
That should be enough for a first pass.

We can always update it later in case something is missing or went wrong during the generation.

PR #913, opened 2021-06-24
Add check-release step on CI

jtpio:
In preparation for using the [Jupyter Releaser](https://github.com/jupyter-server/jupyter_releaser).

Related to #805


jtpio:
The check release check on CI is now passing.

We should at least be able to use the [Draft Changelog](https://github.com/jupyter-server/jupyter_releaser#draft-changelog-workflow) workflow to generate new changelog entries.

PR #922, opened 2021-07-14
http_header_envs option to include http headers as an env var (following CGI standards)

mariobuikhuizen:
Continuation of #835 with CGI standards applied.

To try it out:
```
voila --VoilaConfiguration.http_header_envs=uSeR-agEnt my_notebook.ipynb
```

in my_notebook.ipynb
```python
import os
print(os.environ.get('HTTP_USER_AGENT', 'User agent env var not found'))
```


jtpio:
Kicking CI

maartenbreddels:
Thanks @jtpio !

PR #926, opened 2021-08-02
Obey ip argument and only bind to given address.

adament:
Previously since no address was given to the tornado listen call the
voila application listens on all interfaces.

## References

I have not opened an issue for this and could not find one when I quickly looked.

## Code changes

Simply change the call to tornado.web.Application.listen so that it passes along the Voila.ip parameter as the address argument.

## User-facing changes

If people previously relied on voila by default binding to all interfaces they will now experience a backwards compatibility break, the backwards compatibility can be preserved by changing the default value of ip to '0.0.0.0' (which is what it actually was). Personally I do however believe that the previous behaviour was a security issue (binding on all interfaces while claiming to bind only to localhost) and should be fixed, and users should explicitly opt in to binding to more than localhost. But I understand if preserving backwards compatibility is to preferred.

Same as User-facing changes.

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -547,7 +547,7 @@ class Voila(Application):
         success = False
         for port in self.random_ports(self.port, self.port_retries+1):
             try:
-                self.app.listen(port)
+                self.app.listen(port, self.ip)
                 self.port = port
                 self.log.info('Voilà is running at:\n%s' % self.display_url)
             except socket.error as e:


jtpio:
Thanks @adament!

> If people previously relied on voila by default binding to all interfaces they will now experience a backwards compatibility break

Adding to the 0.3.0 milestone for consideration.

Issue #743, opened 2020-10-21
Injecting Javascript when rendering notebooks with voila (disabled?)


joseberlines:
I am developing an app to be served rendering with voila from jupyter hub.
I am interested in copying text into the clipboard of the user after clicking a ipywidget button.

We are trying to solve this problem by "injecting" Javascript into the code as follows:

```
import ipywidgets as widgets
from IPython.display import display, HTML, Javascript
mybtn = widgets.Button(description='copy to clipboard',button_style='success')
output = widgets.Output()
display(mybtn, output)
def mybtn_event_handler(b):
    with output:
        display(Javascript('''
        var copyText = document.getElementById("myInput");
        copyText.select();
        copyText.setSelectionRange(0, 99999); /*For mobile devices*/
        document.execCommand("copy");
        alert("Copied the text: " + copyText.value);
        ''')) 
mybtn.on_click(mybtn_event_handler)
display(HTML('<input type="text" value="Hello World" id="myInput">'))
```

First we try to get some text into the console to see if we can manage it, and afterwards we would try to find a way to using JS to pass the content of the variable into the clipboard of the user.

The result is nevertheless the following:
![image](https://user-images.githubusercontent.com/33203439/118473057-dde08680-b709-11eb-90e9-34108368470f.png)

It seems that when Rendering with voila Javascript can not be used because it is disabled.
Is that right? Is there any way around this? 

NOTE for completeness. 
The code works in Jupyter Lab when simple run without rendering (Despite of this the error displayed when rendering with voila is "Javascript is disabled in Jupyter lab". (note: be aware that the code of the following screenshot is slightly different but the essence of it is the same. JS will not be injected from voila.)
![image](https://user-images.githubusercontent.com/33203439/96724318-faa11f80-13af-11eb-804a-98bf151660ad.png)

jtpio:
Maybe the cleanest in the long term couple be to have a custom clipboard widget, so that it could be combined with other widgets and used entirely from Python?

joseberlines:
We developed clipboard widget using  ipyvuetify template. But indeed a widget to pass whatever text content to the clipboard would be cool. So it would be contained in the widgets ecosystem and it will not be neccesary to resort to JS.

guimillet:
@jtpio Is there a reason for disabling Javascript code execution in an Output widget but enabling it in a cell execution (in Voilà)?

[...]

PR #931, opened 2021-08-20
Render Javascript output

trungleduc:
## References
https://github.com/voila-dashboards/voila/issues/743

## Code changes

Replace the javascript render factory from ` @jupyterlab/rendermime` with one from `@jupyterlab/javascript-extension`

## User-facing changes

javascript output now works in voila as in jupyterlab

voila

N/A

### DIFF
--- a/packages/voila/package.json
+++ b/packages/voila/package.json
@@ -15,6 +15,7 @@
     "@jupyterlab/apputils": "^3.0.0",
     "@jupyterlab/coreutils": "^5.0.0",
     "@jupyterlab/docregistry": "^3.0.0",
+    "@jupyterlab/javascript-extension": "^3.0.0",
     "@jupyterlab/notebook": "^3.0.0",
     "@jupyterlab/outputarea": "^3.0.0",
     "@jupyterlab/rendermime": "^3.0.0",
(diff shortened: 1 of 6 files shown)


jtpio:
> so it's time to update reference screenshot?

I think so yes.

jtpio:
Just rebased onto the latest to pick up the change from https://github.com/voila-dashboards/voila/pull/934

jtpio (approved):
Thanks!

gunstonej:
this is great - thanks @trungleduc  and @jtpio !

I just tested on my machine  -

it is working for me when Voila is rendered in `http://localhost:8866/` 👍 

... but when rendering using the Jupyter Lab voila renderer i get the following error: 

> `VM666:1 Blocked opening 'https://jupyter.org/' in a new window because the request was made in a sandboxed frame whose 'allow-popups' permission is not set.`

image

any ideas? 

cheers

jtpio:
It's probably because the preview does not allow the popups here:

https://github.com/voila-dashboards/voila/blob/9fa04ce10e2a5f76bb2d272a67cb79698ac138ba/packages/jupyterlab-preview/src/preview.tsx#L50

We could add that sandbox property to the list.

PR #935, opened 2021-09-01
Handle the case when one widget model fails to be created

martinRenou:
## Code changes

Handle the case when one widget model fails to be created, so that other widgets in the page can be rendered anyway.

This behavior is closer to other front-ends like JupyterLab.

## Reproduce

The following Notebook code can reproduce the issue. Here the grid model is invalid and throws an error when being created.

Due to this error, the slider could not be rendered.

```python
# Cell 1
from ipywidgets import Tab, IntSlider

tab = Tab()

import pandas as pd
from ipydatagrid import DataGrid, TextRenderer, VegaExpr

df = pd.DataFrame(
    {
        "column 1": [{"key": 11}, ["berry", "apple", "cherry"]],
        "column 2": [["berry", "berry", "cherry"], {"key": 10}],
    }
)

renderer = TextRenderer(
    background_color=VegaExpr(
        "cell.value[1] == 'berry' && cell.metadata.data['column 1']['key'] == 11 ? 'limegreen' # : 'pink'"
    )
)

tab.children = [DataGrid(
    df,
    layout={"height": "100px"},
    base_column_size=150,
    default_renderer=renderer,
), IntSlider()]
tab

# Cell 2

print('hey')

# Cell 3
IntSlider()
```

Result before this PR (the slider cannot render):
voilaold

Result after this PR (the slider can render):
voilanew

### DIFF
--- a/packages/voila/src/manager.ts
+++ b/packages/voila/src/manager.ts
@@ -217,18 +217,23 @@ export class WidgetManager extends JupyterLabManager {
     await Promise.all(
       widgets_info.map(async widget_info => {
         const state = (widget_info as any).msg.content.data.state;
-        const modelPromise = this.new_model(
-          {
-            model_name: state._model_name,
-            model_module: state._model_module,
-            model_module_version: state._model_module_version,
-            comm: (widget_info as any).comm
-          },
-          state
-        );
-        const model = await modelPromise;
-        models[model.model_id] = model;
-        return modelPromise;
+        try {
+          const modelPromise = this.new_model(
+            {
+              model_name: state._model_name,
+              model_module: state._model_module,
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks!

> This behavior is closer to other front-ends like JupyterLab.

This change seems to correspond to what the JupyterLab manager does: https://github.com/jupyter-widgets/ipywidgets/blob/29c29bf70655b49f0f50230dbeb2e01a15ce55cd/jupyterlab_widgets/src/manager.ts#L114-L118

jtpio (approved):
Looks good, thanks!

martinRenou:
Thanks!

PR #936, opened 2021-09-01
Add pooling heated kernels and pre-rendered notebook

trungleduc:
## New features

This PR aims to accelerate the startup time of Voila by adding two features:

###  Pooling pre-heated kernels 
-  A pool of kernels is started and kept in standby, the notebook  `Voila` started with is executed in every kernel of the pool.
- When a new client requests a kernel, the pre-heated kernel in pool is used and another kernel is started asynchronously to refill the pool.
-  This feature is partially addressed by [`hotpot_km`](https://github.com/voila-dashboards/hotpot_km).  Some code snippets of this PR are taken from this library with permission of the author.

### Pre-rendered notebook
- The HTML version of notebook is rendered in each pre-heated kernel and stored, when a client connects to `Voila`, under some conditions, the cached HTML is served instead of re-rendering the notebook.

## How it works
When Voila started with `preheat_kernel=True`, it will create multiple `asyncio.Task` for each notebook. Each task will start a kernel and execute the notebook in this kernel. Voila keeps a mapping of notebook name to the list of tasks started for this notebook. Each task will return a tuple which contains a list HTML string of rendered cells and an async generator which can be used to render the remaining cells.

As depicted in the image below, the content of elements in this tuple  depends on the moment `Voila` receives a request:

- If there is a kernel with notebook fully executed, `Voila` will get the HTML string of all cells then send to frontend.
- If all kernel are executing notebook, `Voila` will pick the first started  kernel and send the rendered cell to frontend while executing the remaining.

vlw

## Speed comparison 

- With a fully executed notebook  

 https://user-images.githubusercontent.com/4451292/131679292-4f530b9e-ea8c-4f06-ab9e-846900e0439d.mp4 

-With a partially executed notebook

https://user-images.githubusercontent.com/4451292/135833907-920f5712-7d9c-4507-8e5a-d351e732abda.mp4

##  Current status

The pre-heat kernel option works with any kernel manager, it is deactivated by default, re-activate it by setting `preheat_kernel = True`.  For example, with this command, for each notebook Voilà started with, a pool of 5 kernels is created and will be used for new connections.

``` bash

    voila --preheat_kernel=True --pool_size=5
```
To change the additional settings ( list of notebooks does not need pre-heated kernels, number of kernels in pool, refilling delay, environment variables for starting kernel...), users can use `voila.json` configuration file in the same folder containing the notebooks. Here is an example of settings with explanations for pre-heat kernel option. 

```python

   # voila.json
   {
      "VoilaConfiguration": {
         # Activate or deactivate preheat kernel option.
         "preheat_kernel": true 
      },
      "VoilaKernelManager": {
         # A list of notebook name or regex patterns to exclude notebooks from using preheat kernel.
         "preheat_blacklist": [
            "notebook-does-not-need-preheat.ipynb",
            "^.*foo.*$",
            ...
         ], 
         # Configuration for kernel pools
         "kernel_pools_config": { 
            # Setting for `voila.ipynb` notebook
            "voila.ipynb": {
               "pool_size": 3, # Size of pool
               "kernel_env_variables": { # The environment variables used to start kernel for `voila.ipynb`
                  "foo2": "bar2"
               }
            },
            # Setting for `test/sub-voila.ipynb` notebook
            "test/sub-voila.ipynb": {
               "pool_size": 1
            },
            ...
            # If a notebook does not have setting, it will use default setting
            "default": {
               "pool_size": 2,
               "kernel_env_variables": {
                  "foo": "bar"
               }
            },
         },
         # Delay time in second before filling the kernel pool.
         "fill_delay": 0
      }
   }
```

Notebook HTML will be pre-rendered with template and theme defined in VoilaConfiguration or in notebook metadata. The pre-heated kernel and cached HTML are used if these conditions are matched:

- There is an available pre-heated kernel in the kernel pool.
- If user overrides the template/theme with query string, it must match the template/theme used to pre-render the notebook.
- There is no other query strings than `voila-theme` and `voila-template`.

If the kernel pool is empty or the request does not match these conditions, Voila will fail back to start a normal kernel and render the notebook as usual.

## Code changes

- Refactoring `VoilaHandler`: the notebook rendering part of `VoilaHandler` is extracted as a new class `NotebookRenderer`, this class is used to render notebook in both voila handler and pre-heated kernel
- Add new function `voila_kernel_manager_factory` to decorate the kernel manager class defined in `VoilaConfiguration.multi_kernel_manager_class` with pooling features
- Add two new options `preheat_kernel` and `pool_size` to Voila application to set up kernel pool.

Aside `VoilaHandler` refactoring, the notable changes come from this PR are:
- Add two new options `--preheat_kernel` and `pool_size` to `Voila` CLI
- Decorate the user defined kernel manager of `Voila` with the preheat features if `preheat_kernel=True`.

Since `preheat_kernel` is set to `False` by defaut, we should not have any breaking changes with `0.2.x`


jtpio:
Thanks @trungleduc for starting this, it looks very promising!

vidartf:
Would you mind adding some details about how this compares to https://github.com/voila-dashboards/hotpot_km/ ?

trungleduc:
> Would you mind adding some details about how this compares to https://github.com/voila-dashboards/hotpot_km/ ? (i.e. what extra functionality does this add; are there any functionality missing, etc).

Actually I based on [`hotpot_km`](https://github.com/voila-dashboards/hotpot_km) ideal and some implementations to create this PR. Three main differences with respect to ` hotpot_km`:
- The pooling feature can be added to any kernel manager by decoration instead of a standalone kernel manager.
- The pre-executed code is the current notebook of Voila , the possibility to run user defined code is not available.
- The notebook is rendered and cached, when a user connected to voia, both pre-heated kernel and cached HTML are served.

jtpio:
Thanks @trungleduc for pushing on this.

Looks like there is a small conflict to fix.

trungleduc:
Weird,  I just rebase this PR on master to pick up latest changes, then the snapshot for `ipympl` failed 

ipympl-diff

jtpio:
> then the snapshot for `ipympl` failed

Could it be related to the new `0.8.0` release from yesterday?  https://github.com/matplotlib/ipympl/releases/tag/0.8.0

trungleduc:
Yes partially, updated to `0.8.0` on my pc, still have difference between output from CI and my local output. Maybe we need to disable `ipympl` snapshot check?

ipympl-diff

jtpio:
Or we update the screenshot with the one generated on CI?

trungleduc:
In this case the local test won't pass

jtpio:
Right, but I think the common practice is to use the ones generated on CI as reference since this is where the checks are running.

Or use Docker like in the JupyterLab repo to minimize differences, although that would complicate the setup.

jtpio:
By the way, this PR is looking good and promising.

Adding to the `0.3.0` milestone so we get a chance at merging it early next week maybe, if we switch `main` to target `0.3.0` pre-releases.

@vidartf @maartenbreddels would you like to give it a quick review since you have been doing similar work on [`hotpot_km`](https://github.com/voila-dashboards/hotpot_km)? Thanks!

jtpio:
Thanks @trungleduc for working on this.

Let me know if you would like to push more changes. Otherwise we can wait for the tests to pass, merge, and release `0.3.0a0` from `main` with this.

cc @SylvainCorlay as the `0.3.0a0` was discussed on other channels.

jtpio:
OK thanks for heads-up :+1: 

Feel free to ping here when you think it's ready.

trungleduc:
@jtpio It's all green now :smile:

SylvainCorlay:
🎉

jtpio:
Cutting `0.3.0a0` with this change now.

PR #949, opened 2021-09-03
Add Binder on PR workflow

jtpio:
## References

Similar setup as the `ipywidgets` repo: https://github.com/jupyter-widgets/ipywidgets

This will help test changes on Binder without having to check out the code locally.

## Code changes

Binder configuration.

## User-facing changes

Users and contributors can test their changes on Binder.

None

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -7,7 +7,6 @@ dependencies:
   - bqplot
   - scipy
   - ipympl
-  - voila=0.2.11
   - ipympl
   - xleaflet=0.14.0
   - xeus-cling=0.12.1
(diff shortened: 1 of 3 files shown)


jtpio:
The action will add Binder buttons like the following on new PRs:

[Binder](https://mybinder.org/v2/gh/jtpio/voila/binder-prs?urlpath=voila%2Ftree)

jtpio:
This also means we won't be using a released version of `voila` on Binder in this repo.

But it should be fine since the `stable` branch will still point to a release commit. The difference will be that `voila` will be built from source.

We can always quickly setup GitHub gists with `voila` releases if needed.

PR #952, opened 2021-09-07
Update to lerna 4

jtpio:
## References

Update to the latest `lerna`, which also cleans up the dependencies.

## Code changes

Top-level dependency update.

## User-facing changes

None

None


jtpio:
> which also cleans up the dependencies.

Nice diff:

image

jtpio:
Looking good on Binder:

image

PR #954, opened 2021-09-08
Add a script to bump the version

jtpio:
## References

So we can have a single command to bump the version of both the Python packages and the npm packages.

Instead of doing it manually. This way it will also be possible to use the releaser (#805).

## Code changes

Add a new script to bump the version.

### TODO

- [x] Add a script to bump the Python and JS versions
- [x] Bump Python version with `bump2version`
- [x] Align the `@voial-dashboards/voila` version of the JS frontend with the Python version
- [x] Update RELEASE.md with the version bump process (build, patch, major, minor)

## User-facing changes

None

None

### DIFF
--- a/packages/voila/package.json
+++ b/packages/voila/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@voila-dashboards/voila",
   "private": true,
-  "version": "1.0.0",
+  "version": "0.2.11",
   "description": "The Voilà Frontend",
   "author": "Voilà contributors",
   "license": "BSD-3-Clause",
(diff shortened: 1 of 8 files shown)


jtpio:
The "Check Release" check looking on CI:

image

jtpio:
We should also be able to use this script to bump the versions for the next manual release.

packages/voila/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@voila-dashboards/voila",
   "private": true,
-  "version": "1.0.0",
+  "version": "0.2.11",
```
jtpio, line 4: For now this package is still private, but will probably be released too when #846 lands

jtpio:
We might want to iterate further on the script, but for now it should be good to go.

PR #963, opened 2021-09-15
Make `INotebookTracker` optional

jtpio:
## References

This should help fix some dev tools console errors as seen in #846 

And make it possible to still use the `Voila-preview` factory in case the notebook tracker plugin is not available (for example in a custom lab distribution):

image

## Code changes

Make a dependency optional for the lab extension.

## User-facing changes

None

None


jtpio:
Looks good on Binder (no visible changes):

image

PR #972, opened 2021-09-19
 Use `sendBeacon` to send kernel shutdown request.

trungleduc:
## References

 Support for synchronous use of `XMLHTTPRequest()` during page unloads is removed from Chrome (https://www.chromestatus.com/feature/4664843055398912). 

## Code changes
Use `sendBeacon` to send kernel shutdown request on `beforeunload` event

## User-facing changes
N/A

N/A

### DIFF
--- a/share/jupyter/voila/templates/base/static/main.js
+++ b/share/jupyter/voila/templates/base/static/main.js
@@ -52,10 +52,12 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
             // it seems if we attach this to early, it will not be called
             const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
             const xsrfToken = (matches && matches[1]) || '';
+            const  configData = JSON.parse(document.getElementById('jupyter-config-data').textContent);
+            const baseUrl = configData.baseUrl;
             window.addEventListener('beforeunload', function (e) {
                 const data = new FormData();
                 data.append("_xsrf", xsrfToken);
-                window.navigator.sendBeacon(`/voila/api/shutdown/${kernel.id}`, data);
+                window.navigator.sendBeacon(`${baseUrl}voila/api/shutdown/${kernel.id}`, data);
                 kernel.dispose();
             });
             await widgetManager.build_widgets();
--- a/share/jupyter/voila/templates/base/static/main.js
+++ b/share/jupyter/voila/templates/base/static/main.js
@@ -52,7 +52,7 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
             // it seems if we attach this to early, it will not be called
             const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
             const xsrfToken = (matches && matches[1]) || '';
-            const  configData = JSON.parse(document.getElementById('jupyter-config-data').textContent);
+            const configData = JSON.parse(document.getElementById('jupyter-config-data').textContent);
             const baseUrl = configData.baseUrl;
             window.addEventListener('beforeunload', function (e) {
                 const data = new FormData();
--- a/share/jupyter/voila/templates/base/static/main.js
+++ b/share/jupyter/voila/templates/base/static/main.js
@@ -53,10 +53,9 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
             const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
             const xsrfToken = (matches && matches[1]) || '';
             window.addEventListener('beforeunload', function (e) {
-                const xhttp = new XMLHttpRequest();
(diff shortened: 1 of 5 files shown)


jtpio:
Thanks @trungleduc.

Re-reading the discussion in https://github.com/voila-dashboards/voila/pull/612, I think the original intent was to be able to set that flag to `false` so the request doesn't get aborted in the middle of the page unload.

jtpio:
>  is removed from Chrome (https://www.chromestatus.com/feature/4664843055398912)

Given this:

> "While this is reasonable, there are no good reasons to use synchronous XHR. Instead SendBeacon, Fetch keep-alive should be used (in addition to sending analytics periodically and on pagevisibility etc.)."

Maybe we should give another look at using `sendBeacon` which was experimented with in https://github.com/voila-dashboards/voila/pull/612#issuecomment-632751938

trungleduc:
Thanks @jtpio, sorry for didn't read the discussion in #612 before. I'll open an issue and try the `sendBeacon` with this PR

jtpio:
Thanks!

Testing on Binder with this link:

https://mybinder.org/v2/gh/trungleduc/voila/fix-kernel-shutdown?urlpath=voila%2Ftree%2Fnotebooks

It looks like the request is correctly sent, but returns a `405`:

https://user-images.githubusercontent.com/591645/134173273-16f9dde8-1636-4788-82da-dd55f78989e9.mp4

Not sure if it's a limitation on the Binder side. Voila runs as a server extension there and the diff looks good w.r.t that.

trungleduc:
It looks like the POST URL `https://hub.gke2.mybinder.org/voila/api/shutdown/...` is not correct  since Voila server is exposed at `https://hub.gke2.mybinder.org/user/trungleduc-voila-mxw256mu/`. It should send request to something like `https://hub.gke2.mybinder.org/user/trungleduc-voila-mxw256mu/voila/api/shutdown/...` to make it works

share/jupyter/voila/templates/base/static/main.js
```diff
@@ -53,10 +53,9 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
             const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
             const xsrfToken = (matches && matches[1]) || '';
             window.addEventListener('beforeunload', function (e) {
-                const xhttp = new XMLHttpRequest();
-                xhttp.open("DELETE", `/api/kernels/${kernel.id}`, false);
-                xhttp.setRequestHeader('X-XSRFToken', xsrfToken);
-                xhttp.send();
+                const data = new FormData();
+                data.append("_xsrf", xsrfToken);
+                window.navigator.sendBeacon(`/voila/api/shutdown/${kernel.id}`, data);
```
jtpio, line 58: So probably here we need to handle the base URL?

voila/app.py
```diff
@@ -470,6 +471,7 @@ def start(self):
                     'default_filename': 'index.html'
                 },
             ),
+            (url_path_join(self.server_url, r'/voila/api/shutdown/(.*)'), VoilaShutdownKernelHandler)
```
jtpio, line 474: Also maybe we could cover this new endpoint with a small unit test?

share/jupyter/voila/templates/base/static/main.js
```diff
@@ -52,11 +52,12 @@ require([window.voila_js_url || 'static/voila'], function(voila) {
             // it seems if we attach this to early, it will not be called
             const matches = document.cookie.match('\\b_xsrf=([^;]*)\\b');
             const xsrfToken = (matches && matches[1]) || '';
+            const  configData = JSON.parse(document.getElementById('jupyter-config-data').textContent);
```
jtpio, line 55: Minor formatting edit to remove an extra space:

```suggestion
            const configData = JSON.parse(document.getElementById('jupyter-config-data').textContent);
```

voila/shutdown_kernel_handler.py
```diff
@@ -0,0 +1,13 @@
...
+from jupyter_server.base.handlers import APIHandler
+
+
+class VoilaShutdownKernelHandler(APIHandler):
+    """ Handler to shut down kernel on page's `beforeunload` event.
+    """
+
+    @tornado.web.authenticated
+    async def post(self, kernel_id):
+        await self.kernel_manager.shutdown_kernel(kernel_id)
```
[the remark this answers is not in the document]
trungleduc: Yes, you're right. How can I have access to jupyter lab console like in this image? It seems very useful for debugging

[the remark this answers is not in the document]
jtpio: On Binder the log file is available under `~/.jupyter-server-log.txt`:

image

trungleduc:
it seems to work now in Binder 

voila

jtpio (approved):
Thanks @trungleduc!

PR #980, opened 2021-09-25
Update to nbconvert 6.2

SylvainCorlay:
Nbconvert 6.2 adds an argument to the `jupyter_widgets` macro.

Overriding this in voilà with the old number of arguments breaks calling it from the base nbconvert templates with the higher number of args.


SylvainCorlay:
The visual regression test failure appears to be a 1-px shift due to the ipympl release.

PR #983, opened 2021-09-29
Fix `post` function in log template

trungleduc:
Update `post` function signature  in log template.


trungleduc:
Failing checks are not related to PR

jtpio (approved):
Thanks!

PR #984, opened 2021-09-30
Implement FPS-based Voila server

### LINKED ISSUE #977: Implement a `fps` plugin for Voila
### Problem

In the [Jupyter Server Meeting yesterday](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-925941938), we discussed the use case of having a `fps` plugin for Voila. This would be a good example for having a tihrd-party plugin living in a different repo.

`fps` and `jupyverse` are still experimental projects, but since `jupyverse` relies on [`entry_points`](https://github.com/jupyter-server/jupyverse/blob/f36b9b1f1a4178202402508221925959a2d48b18/plugins/jupyterlab/setup.py#L14-L17) to load the plugins, we could imagine having the `fps-voila` plugin as part of the `voila` package already.

This could be shipped as an early preview and released as part of the `0.3.x` series.

It would also help refactor the voila logic so it can be used with both `jupyter_server` extensions and `jupverse` plugins.

### Proposed Solution

We add a `fps-voila` plugin some time in the `0.3.0` release (additive change).

### Additional context

- https://github.com/jupyter-server/team-compass/issues/11
- https://github.com/jupyter-server/fps
- https://github.com/jupyter-server/jupyverse

davidbrochart:
## References

Closes #977

## Code changes

- Add a Voila `fps` flag, allowing to use FPS instead of Tornado.
- Have `VoilaHandler` depend on a server-agnostic `_VoilaHandler` class, so that the FPS server can reuse most of the handler logic.
- Create a `fps-voila` FPS plugin, which implements a `FPSVoilaHandler` based on `_VoilaHandler`.

## User-facing changes

By default, Voila is backed by `jupyter-server` (no change). To use FPS, launch Voila with the `--fps` flag.

## Limitations

This is still a POC, which allows to serve a notebook with widgets, but:
- no progressive HTML rendering,
- no tree view,
- and probably more broken functionalities!


trungleduc:
@davidbrochart  https://github.com/voila-dashboards/voila/pull/936 is merged, [`NotebookRenderer`](https://github.com/voila-dashboards/voila/blob/5b10cf6e8ecdf854b5fd6db92e26b944110f1840/voila/notebook_renderer.py#L30) is now available.

davidbrochart:
Thanks, I will look at it.

davidbrochart:
I just rebased, now that #936 is merged.

trungleduc:
Thank @davidbrochart for your work on this, this is looking very nice!

trungleduc:
One question about testing, can we make all tests run on both `jupyter-server` and `fps` ?

voila/handler.py
```diff
@@ -24,29 +25,20 @@
...
 
-class VoilaHandler(JupyterHandler):
-    def initialize(self, **kwargs):
-        self.notebook_path = kwargs.pop('notebook_path', [])  # should it be []
-        self.template_paths = kwargs.pop('template_paths', [])
-        self.traitlet_config = kwargs.pop('config', None)
-        self.voila_configuration = kwargs['voila_configuration']
-        # we want to avoid starting multiple kernels due to template mistakes
-        self.kernel_started = False
+async def _get(self, path=None):
```
trungleduc, line 28: I think we might need docstring and typing for this function because using `self` as a function parameter is a little bit confusing.

davidbrochart:
> One question about testing, can we make all tests run on both `jupyter-server` and `fps` ?

I didn't look at tests, I think they are very tightly coupled with Tornado and jupyter-server.

davidbrochart:
@trungleduc any idea on where to start for running tests on FPS?

trungleduc:
> @trungleduc any idea on where to start for running tests on FPS?

Hi @davidbrochart, you can pass `--fps`  flag to `Voila` initializer to activate this mode.
https://github.com/voila-dashboards/voila/blob/41d4ba97db594eaf49e73bc2a729b3f82262bd74/tests/app/conftest.py#L53

trungleduc:
We can create a parameter fixture to switch between `--fps` `on` and `off` to test both implementations of the server.

davidbrochart:
Sure, but then this just runs a (blocking) server, and the tests won't run. Currently, tests are tightly coupled with Tornado because they use a client from pytest-tornasync, and they overwrite the `listen` method:
https://github.com/voila-dashboards/voila/blob/41d4ba97db594eaf49e73bc2a729b3f82262bd74/tests/app/conftest.py#L9-L11

trungleduc:
Ah sorry, I forgot that part. It seems to be more complicated than I thought.

davidbrochart:
Yes, I'm not sure we can factorize the tests so that they can be server-agnostic, that's why I didn't look at the tests yet.
Even for the functional Voilà part, the Tornado and FastAPI APIs are so different that making it server-agnostic results in something kind of hacky, where we stick to the Tornado API and try to adapt it to FastAPI, but it inherently doesn't match.

trungleduc:
I think so too, that's why I initially thought about refactoring the `Voila` app to create a common interface for implementing the server part so that we can add `fps` independently. It also helps us in the future if we decide to remove `Tornado` and use only `fps`

davidbrochart:
Ok, I'll try to make the Voila App server-agnostic, although I think that if/when we decide to remove Tornado, there won't be any point in keeping the current architecture, which doesn't fit well with FastAPI's API.

davidbrochart:
UI tests are now passing.

trungleduc:
I think you need to update the screenshots for `voila` project

trungleduc:
@davidbrochart should we close this PR? It looks like `FPS` won't be maintained anymore.

davidbrochart:
I didn't remember we started that. Yes I'll close it and we will need to open a new PR when https://github.com/jupyter-server/jupyverse/pull/277 is merged.

Issue #986, opened 2021-10-04
Voila build for python 3.8 on CI failed


trungleduc:
## Error log

```
  > tsc
  
  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
  lerna ERR! yarn run build:prod stderr:
  node:internal/crypto/hash:67
    this[kHandle] = new _Hash(algorithm, xofLen);
                    ^
  
  Error: error:0308010C:digital envelope routines::unsupported
      at new Hash (node:internal/crypto/hash:67:19)
      at Object.createHash (node:crypto:130:10)
      at module.exports (/tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/webpack/lib/util/createHash.js:135:53)
      at NormalModule._initBuildHash (/tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/webpack/lib/NormalModule.js:417:16)
      at handleParseError (/tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/webpack/lib/NormalModule.js:471:10)
      at /tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/webpack/lib/NormalModule.js:503:5
      at /tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/webpack/lib/NormalModule.js:358:12
      at /tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/loader-runner/lib/LoaderRunner.js:373:3
      at iterateNormalLoaders (/tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/loader-runner/lib/LoaderRunner.js:214:10)
      at Array.<anonymous> (/tmp/pip-req-build-uwl0cm2q/packages/voila/node_modules/loader-runner/lib/LoaderRunner.js:205:4)
      at Storage.finished (/tmp/pip-req-build-uwl0cm2q/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:55:16)
      at /tmp/pip-req-build-uwl0cm2q/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:91:9
      at /tmp/pip-req-build-uwl0cm2q/node_modules/graceful-fs/graceful-fs.js:123:16
      at FSReqCallback.readFileAfterClose [as oncomplete] (node:internal/fs/read_file_context:68:3) {
    opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
[...]

PR #987, opened 2021-10-04
Pin openssl to v1.1.1l

trungleduc:
Close https://github.com/voila-dashboards/voila/issues/986


jtpio:
Thanks @trungleduc!

If this is Python 3.8 specific, maybe we could do that for 3.8 only by checking the python version in the matrix?

trungleduc:
No this bug affects  both Python 3.8 and 3.9 but for some reason, Python 3.9 test on CI  installed version `1.1.1l`  instead of `3.0.0`

PR #995, opened 2021-10-05
Bot skips Binder badge comment

### LINKED ISSUE #994: Benchmarks workflow overwrites the Binder badge comment
## Description

At the moment the benchmark workflow overwrites the Binder badge comment:

image

## Reproduce

Example: https://github.com/voila-dashboards/voila/pull/993

## Expected behavior

This logic would need to handle other comments that are not benchmark comments:

https://github.com/voila-dashboards/voila/blob/4cf7eaf3a81c1c7748a6066f73399e37de1c0650/.github/workflows/benchmark-report.yml#L61-L70

## Context

Noticed in https://github.com/voila-dashboards/voila/pull/993#issuecomment-934547899

trungleduc:
Close #994

### DIFF
--- a/.github/workflows/benchmark-report.yml
+++ b/.github/workflows/benchmark-report.yml
@@ -59,7 +59,8 @@ jobs:
             })
 
             // Find any comment already made by the bot.
-            const botComment = comments.find(comment => comment.user.id === 41898282)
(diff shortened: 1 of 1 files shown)


trungleduc:
This PR need to be merged for the changes take effect since GHA only consider workflow file on the default branch for `workflow_run`  event.

PR #996, opened 2021-10-05
Simplify tests for preheat kernel mode

trungleduc:
Follow up https://github.com/voila-dashboards/voila/pull/936, remove some unnecessary codes in tests for preheat kernel mode.


jtpio (approved):
CI all green, thanks!

Issue #561, opened 2020-03-22
404 : Not Found when running voila from jupyterlab


tylerburleigh:
I've installed voila

`pip install voila`

and I've installed the jupyterlab extension

`jupyter labextension install @jupyter-voila/jupyterlab-preview`

When I try to `Render Notebook with Voila` or `Open with Voila in New Browser Tab`, I get a 404 : Not Found error.

![image](https://user-images.githubusercontent.com/3513754/77256554-6a4a0300-6c45-11ea-8a93-182bbaff971a.png)

However when I launch the same notebook using voila from the command line (`voila index.ipynb`) I am able to then access it in a web browser. I think maybe when I launch voila from Jupyterlab it's looking in the wrong place for the file, but I'm not sure how to fix this.

tylerburleigh:
I saw this issue, but I'm not using JupyterHub so I figured this warranted a new issue:
https://github.com/voila-dashboards/voila/issues/560

tylerburleigh:
Oh I found the issue: voila was installed in `/home/tyler/.local/bin` but jupyterlab was trying to run it as root. I fixed it by adding this line to `~/.bashrc`:

`export PATH=/home/tyler/.local/bin:$PATH`

and then 

`source .bashrc`

28-08-2000:
i got the same error as above( Page you requested does not exist).
Then i followed this
conda install -c conda-forge jupyter
conda install notebook
and then ran jupyter notebook. 
and it worked!!!

jasoriya:
I'm facing this issue and none of the above methods resolves it

SylvainCorlay:
Could you please report the output of

`jupyter serverextension list`?

jasoriya:
`~$ jupyter serverextension list
config dir: /home/cyc/.jupyter
    jupyterlab  enabled
    - Validating...
      jupyterlab 3.0.15 OK
`

Skware-Cyber:
> i got the same error as above( Page you requested does not exist).
> Then i followed this
> conda install -c conda-forge jupyter
> conda install notebook
> and then ran jupyter notebook.
> and it worked!!!

thanks it worked!!

PR #1000, opened 2021-10-14
Add a separate line for `jupyter server extension enable` in README

StanczakDominik:
This proposed change tries to clean up a bit of confusion about the server extension workflow.

I was running into an issue like https://github.com/voila-dashboards/voila/issues/561 . However, I wanted to keep my Jupyter install as a system package on Arch (obligatory "arch, btw!") rather than switching to conda. Thus, the addition of `--sys-prefix` to `jupyter serverextension enable` was not an option for me. I thus tried `jupyter serverextension enable voila`, but to no avail.

There's no Arch/AUR package for Voila just yet (I might make one later), so I'm using Voila via `pip install --user`.

However, I found out in https://github.com/voila-dashboards/voila/blob/88d55eaa418f497cc338a4a0602c961449b76de4/.binder/postBuild that `jupyter serverextension` and `jupyter server extension` are two different concepts (no idea why, honestly). I tried running `jupyter server extension enable --user voila`, and, well... voila! It works now.

So I'd like to propose this change as a "hey, this is confusing in the documentation; perhaps let's change it up in places!" idea.

## References

Addresses similar problems to #561 

## Code changes, User-facing changes, Backwards-incompatible changes

None whatsoever; just a README change. I'd be happy to also improve the docs once I get better understanding of why this is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ Voilà can also be used as a Jupyter server extension, both with the
 To install the Jupyter server extension, run
 
 ```
-jupyter serverextension enable voila --sys-prefix
+jupyter serverextension enable voila
+jupyter server extension enable voila
 ```
 
 When running the Jupyter server, the Voilà app is accessible from the base url


jtpio:
Thanks @StanczakDominik!

> However, I found out in https://github.com/voila-dashboards/voila/blob/88d55eaa418f497cc338a4a0602c961449b76de4/.binder/postBuild that `jupyter serverextension` and `jupyter server extension` are two different concepts (no idea why, honestly). I tried running `jupyter server extension enable --user voila`, and, well... voila! It works now.

`jupyter serverextension` is for the classic notebook server which is currently used on Binder when starting the servers. And `jupyter server extension` is for [Jupyter Server](https://github.com/jupyter/jupyter_server). Jupyter Server originated from the classic notebook server and should now be the preferred option.

jtpio (approved):
Thanks!

README.md
```diff
@@ -74,7 +74,8 @@ Voilà can also be used as a Jupyter server extension, both with the
 To install the Jupyter server extension, run
 
 ```
-jupyter serverextension enable voila --sys-prefix
+jupyter serverextension enable voila
+jupyter server extension enable voila
```
jtpio, line 78: If we want to be more specific, we could even add comments such as:

```suggestion
# if you are using the classic notebook server
jupyter serverextension enable voila

# if you are using Jupyter Server
jupyter server extension enable voila
```

PR #1005, opened 2021-10-20
Add token management

### LINKED ISSUE #825: Token management or any other secured way to serve a Voilà application
Hi,
Maybe I missed some improvements but I didn't find a way to authenticate users before using a Voilà application.
In my case I don't necessarily need a secure authentication but it would be good to be able to use a token (that enough security for my needs).
Is there any reason why this is not implemented ? 
There is some commented code about that in app.py with that line above : 
```
# TODO: do we want to have the token?
```
I made some modifications to add an option `--Voila.token=<unicode>` to let user add his own token, so if you think this feature can be interesting I can create a PR.
Nicolas

brichet:
Fix #825 by adding an optional token parameter when running Voila.
Token must be filled manually in command line, it is not automatically generated.
The token parameter must be provided in VoilaHandler and VoilaTreeHandler URL, otherwise a 404 error is thrown.

### DIFF
--- a/tests/app/token_test.py
+++ b/tests/app/token_test.py
@@ -1,0 +1,39 @@
+# test all objects that should be configurable
+import pytest
+import tornado
+
+import os
+
+from secrets import token_hex
+
+BASE_DIR = os.path.dirname(__file__)
+TOKEN = token_hex()
+
+
+@pytest.fixture
+def base_url():
+    return "/?token=%s" % TOKEN
+
+
+@pytest.fixture
+def voila_args_extra():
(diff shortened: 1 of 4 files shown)


majidaldo:
would this better be deferred to jupyter?

brichet:
Jupyter already handles this in each application :
- NotebookApp.Token for Jupyter Notebook
- ServerApp.Token for Jupyter Lab

This pull request is basically a copy of what is done in Jupyter to bring the feature to Voilà.

brichet:
Superseded by https://github.com/voila-dashboards/voila/pull/1308

PR #1006, opened 2021-10-20
add computed or given ip address or hostname to tornado init

athivillon:
Voila.ip is documented as "--Voila.ip=<Unicode>
    The IP address the notebook server will listen on.
    Default: 'localhost'
"

But this is false, Tornado is listening on all interfaces, including public ones, even if we use 127.0.0.1. In my opinion, this is a huge security risk, at least for using Voila on servers having Internet access.

This patch initialises Tornado with address option filled with ip computed or passed on command-line.

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -566,7 +566,7 @@ class Voila(Application):
         success = False
         for port in self.random_ports(self.port, self.port_retries+1):
             try:
-                self.app.listen(port)
+                self.app.listen(port,address=self.ip)
                 self.port = port
                 self.log.info('Voilà is running at:\n%s' % self.display_url)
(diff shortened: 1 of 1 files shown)


voila/app.py
```diff
@@ -566,7 +566,7 @@ def listen(self):
         success = False
         for port in self.random_ports(self.port, self.port_retries+1):
             try:
-                self.app.listen(port)
+                self.app.listen(port,address=self.ip)
```
trungleduc, line 569: Thank you very much for your contribution! A minor code format problem:

```suggestion
                self.app.listen(port, address=self.ip)
```

PR #1025, opened 2021-11-09
Implement async generator based Voila get handler

davidbrochart:
## References

This change is already part of https://github.com/voila-dashboards/voila/pull/984, which will have to be rebased when this PR gets in.

## Code changes

This PR aims at having a Voila get handler that is less coupled with Tornado. This handler streams HTML snippets as they are produced by the notebook execution. Instead of directly writing the HTML snippets with Tornado-specific methods (`write` and `flush`), we generate (`yield`) them with an async generator. This allows to use this generator e.g. with a FastAPI `StreamingResponse`.

cc @SylvainCorlay

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -8,6 +8,7 @@
 #############################################################################
 
 
+import asyncio
 import os
 from typing import Dict
 
@@ -33,8 +34,7 @@ class VoilaHandler(JupyterHandler):
         # we want to avoid starting multiple kernels due to template mistakes
         self.kernel_started = False
 
-    @tornado.web.authenticated
-    async def get(self, path=None):
+    async def get_generator(self, path=None):
         # if the handler got a notebook_path argument, always serve that
         notebook_path = self.notebook_path or path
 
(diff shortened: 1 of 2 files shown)


trungleduc:
Thank @davidbrochart, I will be very useful for the integration of `fps`!
I quite like your implementation of an independent get generator in #984 and it looks like we can do the same thing for `get_generator` here. What do you think?

davidbrochart:
Do you mean the get handler method being extracted from the class? I so, maybe in another PR?

trungleduc:
Yes, and in the next PR is a good idea too!

davidbrochart:
Ready to merge this one?

trungleduc:
Yes I think so, only question left is how to target this PR to `0.3.1` since we are already in the RC phase for `0.3.0`. Maybe @jtpio will have some insights?

jtpio:
Looking at the diff, it looks like a change affecting the core logic of the main render handler.

So I would say a bit late for an rc (normally focused on bug fixes), unless we consider this change as a bug fix.

davidbrochart:
No, it's not a bug fix.

trungleduc:
So we will wait for `0.3.0` (hopefully in 2 weeks ) before merging this PR?

trungleduc:
@jtpio I'm going to merge this PR, do I need to create a `0.3.x` branch before merging it to `main`?

jtpio:
I think we can keep `main` targeting the `0.3.x` release series for a while before creating a new branch.

trungleduc:
Rebase to get the UI test fixes

trungleduc:
Thanks @davidbrochart !

Issue #1027, opened 2021-11-15
Drop `python 3.6` and `traitlets 4` from CI test 


trungleduc:
`Voila` depends heavily on `nbconvert` and `nbconvert` required minimum `python 3.7` and `traitlets 5` since version `6.1.0`. Should we drop  the configuration for testing `python 3.6` and add the test for `python 3.10`?

trungleduc:
@jtpio should we go for it?

jtpio:
Maybe we could start testing on 3.7 and 3.10 on CI as a first step. I'll open a PR.

martinRenou:
Remove traitlets 4 test in https://github.com/voila-dashboards/voila/pull/1060

PR #1037, opened 2021-12-08
Updating Heroku deploy instructions

fhightower:
In addition to replacing a Unicode character with it's ascii replacement, I've also updated the Python runtime as the one listed is no longer valid. Specifically, I've:

- Updated the Python runtime
- Added a link to Heroku's docs on Python runtimes so if/when the new runtime becomes invalid, it is easy to find the updated one

I chose not to use the latest runtime (`3.10.0`) as I'm not sure that (based on #1027) that voila supports Python 3.10.x yet.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -120,13 +120,13 @@ High level instructions, specific to Voilà can be found below:
 
    .. code:: text
 
-       web: voila —-port=$PORT --no-browser
+       web: voila --port=$PORT --no-browser
 
    Or the following if you only want to show one notebook:
 
    .. code:: text
 
-       web: voila —-port=$PORT —-no-browser your_notebook.ipynb
+       web: voila --port=$PORT --no-browser your_notebook.ipynb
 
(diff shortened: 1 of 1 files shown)


docs/source/deploy.rst
```diff
@@ -108,25 +108,25 @@ High level instructions, specific to Voilà can be found below:
...
-       python-3.7.3
+       python-3.9.9
 
 3. Add a file named Procfile to the project directory with the
    following content if you want to show all notebooks:
 
    .. code:: text
 
-       web: voila —-port=$PORT --no-browser
+       web: voila --port=$PORT --no-browser
```
fhightower, line 123: I replaced this character (and throughout the file) as it was a Unicode character (I believe an [em dash](https://www.unicodepedia.com/unicode/general-punctuation/2014/em-dash/)).

jtpio (approved):
Thanks!

jtpio:
RTD build passes and other CI failures are not related.

jtpio:
Ah looks like this was targeting the `stable` branch,

@meeseeksdev please backport to main

PR #1042, opened 2021-12-09
Automated Changelog Entry for 0.3.0 on main

trungleduc:
Automated Changelog Entry for 0.3.0 on main
```
Python version: 0.3.0
npm version: @voila-dashboards/voila-root: 0.1.0
npm workspace versions:
@voila-dashboards/jupyterlab-preview: 2.1.0
@voila-dashboards/voila: 0.3.0
```

After merging this PR run the "Full Release" Workflow on your fork of `jupyter_releaser` with the following inputs
| Input  | Value |
| ------------- | ------------- |
| Target | voila-dashboards/voila  |
| Branch  | main  |
| Version Spec | release |
| Since | v0.2.16 |


trungleduc:
@jtpio it seems better here, do we need to remove the changelogs of the old alpha, beta, RC... version?

jtpio:
I guess we can yes, since the entries are the same.

jtpio:
For reference the "since last stable" parameter could also have been used:

image

PR #1053, opened 2022-01-12
Lab template: fix body tag

martinRenou:
## References

In combination with https://github.com/jupyter/nbconvert/pull/1704, it fixes an issue where the background for Matplotlib figures is not correct.

## Code changes

Change the body tag to add the missing `data-jp-theme-light` and `data-jp-theme-name` properties.

### Before

before

### After

after

None


trungleduc (approved):

trungleduc:
Thanks @martinRenou!

PR #1054, opened 2022-01-12
Try fixing UI tests

martinRenou:

PR #1055, opened 2022-01-12
Pin nbconvert to 6.3

martinRenou:
Checking if nbconvert is responsible for the UI tests failure fixed in https://github.com/voila-dashboards/voila/pull/1054


martinRenou:
Not merging this PR, this was just a test.

PR #1058, opened 2022-01-14
Fix element style display

martinRenou:
## Code changes

Setting "unset" has some impacts on the way cells are rendered in `voila-retro`. Furthermore "" should be the default, and not 'unset'?

## User-facing changes

None in Voila

None


trungleduc:
Failed tests are not related to PR, the issue is reported here https://github.com/jupyter-server/jupyter_server/issues/661

trungleduc (approved):

PR #1060, opened 2022-01-17
Remove traitlets 4 test, add missing traitlets dependency and drop Python 3.6 support

martinRenou:
## References

Fix https://github.com/voila-dashboards/voila/issues/1027

## User-facing changes

None

None

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,10 +16,10 @@ classifiers =
     License :: OSI Approved :: BSD License
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
     Framework :: Jupyter
     Framework :: Jupyter :: JupyterLab
     Framework :: Jupyter :: JupyterLab :: 3
@@ -30,7 +30,7 @@ classifiers =
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.6
+python_requires = >=3.7
 install_requires =
     jupyter_server>=0.3.0,<2.0.0
     jupyter_client>=6.1.3,<8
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -37,6 +37,7 @@ install_requires =
     nbclient>=0.4.0,<0.6
     nbconvert>=6.0.0,<7
     websockets>=9.0
+    traitlets>=5.0.3,<6
```
trungleduc, line 40: By forcing `traitlets` 5, will `Voila` work with `jupyterlab` 1 and 2?

martinRenou: You mean the labextension?

trungleduc: I mean creating an environment with `Voila` and `JupyterLab` together. Just tested, with `traitlets>=5.0.3`, we need to bump `python_requires = >=3.7` too.

martinRenou: > I mean creating an environment with Voila and JupyterLab together

This does not seem to be a problem (at least I don't see a problem specific to the `traitlets >=5.x` pinning):
- JupyterLab 2.x depends `jupyterlab_server >=1.1.0,<2.0.0` on https://github.com/conda-forge/jupyterlab-feedstock/blob/2.x/recipe/meta.yaml#L26
- `jupyterlab_server 1.2.0` depends on `notebook >=4.2.0` https://github.com/conda-forge/jupyterlab_server-feedstock/blob/56e32c475ac6a26562c7687ebe9b357f0cf7d071/recipe/meta.yaml#L26
- latest notebook depends on `traitlets >=4.2.1` https://github.com/conda-forge/notebook-feedstock/blob/master/recipe/meta.yaml#L46

> Just tested, with traitlets>=5.0.3, we need to bump python_requires = >=3.7 too

Thanks! Will update my PR

trungleduc (approved):

trungleduc:
Thanks @martinRenou !

martinRenou:
Thanks for the review!

PR #1062, opened 2022-01-18
Automatically download labvariables CSS instead of shipping them

martinRenou:
Remove `labvariables.css` and `materialcolors.css` files from the repo. Instead we download them automatically, similar to what nbconvert does: https://github.com/jupyter/nbconvert/blob/main/setup.py

## Code changes

Automatically download CSS files on install

## User-facing changes

None

None


martinRenou:
I don't get the UI test failure. It looks like a font-size difference but the font-size has not been changed using this PR.

This CSS variables are still properly discovered by the reveal template.

I will update the reference screenshots.

trungleduc:
Hi @martinRenou, by fetching CSS files from the internet on installation, users who install `Voila` in an offline environment (e.g directly via `tar.gz` file or via local `pypi` repository) will not be able to do it anymore.

trungleduc:
> > My bad, this PR actually downloads CSS files at build time, which is great!
> 
> No I think you had a point. This CSS, not being in the sdist, will not be part of the `tar.gz` published on Pypi.

I looked at `nbconvert`, both the `tag.gz` and `.whl` have the downloaded CSS files.

martinRenou:
> I looked at nbconvert, both the tag.gz and .whl have the downloaded CSS files.

With this PR in voila, it's **not** included in the `tar.gz`, it is included in the `whl` though.

It's not included in the `tar.gz` because I added an `exclude` in the `MANIFEST.IN` (which nbconvert does not do). We could remove this `exclude` and ship the CSS with the `tar.gz` (I am personally totally ok with this, it would fix the "non internet connection" setup), but the check-manifest CI check is complaining if there are files included in the `sdist` that are not tracked by `git`.

trungleduc:
Can we update `tool.check-manifest` ignore list?

martinRenou:
Yes, we can do that. I wonder if that could be considered bad practice?

trungleduc:
> Yes, we can do that. I wonder if that could be considered bad practice?

I don't know,  but we had the same pattern for ignoring `js` files in `check-manifest`.

martinRenou:
Ok, thanks :) I updated my PR

setup.cfg
```diff
@@ -56,3 +56,10 @@ test =
 [options.entry_points]
 console_scripts =
     voila = voila.app:main
+
+[check-manifest]
```
trungleduc, line 60: We already have `check-manifest` setting in  `pyproject.toml`

martinRenou: Oh I missed it, thanks!

martinRenou:
I accidentally canceled the CI, retriggering it

trungleduc (approved):

trungleduc:
Thanks @martinRenou !

Issue #1049, opened 2021-12-22
File download not routed to correct path


dcnadler:
## Description

For v. 0.3.0, when launching as a directory, clicking on a file link (displayed by `FileLink`) in an app it does not forward to expected path of `/voila/files/my_report.html`.

The user is instead taken to `/voila/render/my_report.html`

This does not happen with v. 0.2.16

## Reproduce

Files:
* docker-compose.yml
* Dockerfile
* requirement.txt
* test.ipynb

requirements.txt
```
voila
notebook
```
or
```
voila<0.3
notebook
```

test.ipynb
```python
from IPython.display import display, FileLink
fpath = './my_report.html'
with open(fpath, "w") as f:
    f.write('<h1>My Report</h1>\n')
    f.write('All the details\n')
display(FileLink(fpath))
```

Dockerfile
```Dockerfile
FROM python:3.8

COPY requirements.txt /app/

WORKDIR /app
RUN pip install --upgrade pip && pip install --no-cache-dir -r requirements.txt

COPY . /app/

CMD voila . --Voila.ip=0.0.0.0 --VoilaConfiguration.file_whitelist="['.*.html']"
```

docker-compose.yml
```yaml
version: '3.7'

services:
  app:
    build: .
    image: voila-test
    ports:
      - 8866:8866
    volumes:
      - .:/app
```

Run app with `docker-compose up`, access at localhost:8866. Click on link test.ipynb link to access that app.

## Expected behavior

For both voila 0.2 and 0.3, a hyperlink shows up after launching the app that points to:
http://localhost:8866/voila/render/my_report.html

For v 0.2.16, it redirects to: http://localhost:8866/voila/files/my_report.html

For v 0.3.0, however, it does not forward so you get a 500 error at http://localhost:8866/voila/render/my_report.html

I didn't notice this issue when launching as an individual notebook, only as a directory.

Thank you!

trungleduc:
Thanks for your report, I can reproduce this bug, I'll take a look at it.

trungleduc:
Fixed by #1073

PR #1073, opened 2022-01-24
Fix file redirect error

### LINKED ISSUE #1071: Voila doesn't load static HTML files despite whitelisting
Following up the conversation in https://github.com/voila-dashboards/voila/issues/1044#issuecomment-1019301192:

## Description

My Jupyter notebook loads `ipywidgets` some of them instantiating `IFrame` objects loading static html files.

Despite whitelisting html files in Voila (v0.3.0), the Voila server fails with a 500 error:
```
ERROR:tornado.application:Uncaught exception GET /voila/render/hello.html (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:8866', method='GET', uri='/voila/render/hello.html', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/tilushq/opt/miniconda3/envs/jupyter/lib/python3.7/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/home/tilushq/opt/miniconda3/envs/jupyter/lib/python3.7/site-packages/voila/handler.py", line 150, in get
    kernel_name=gen.notebook.metadata.kernelspec.name,
AttributeError: 'NoneType' object has no attribute 'metadata'
ERROR:tornado.access:500 GET /voila/render/hello.html (127.0.0.1) 18.03ms
```

## Reproduce

All files are in working directory X:
* voila.json
* test.ipynb
* hello.html

1. Create a `voila.json` file with configuration:

```json
{
    "VoilaConfiguration": {
        "file_whitelist": [
            ".*\\.(html|js)"
        ]
    }
}
```

2. Create a `test.ipynb` with the minimal code:

```python
hello_file = 'hello.html'
hello_body = '<body><h1>Hello World!</h1></body>'

with open(hello_file, 'w') as fw:
    _ = fw.write(hello_body)
    
from IPython.display import display, display_html, IFrame

my_html_widget = IFrame(
    src=hello_file,
    width=300,
    height=100
)

import ipywidgets as w

widget_output = w.Output()
with widget_output:
    display(my_html_widget)

widget_output
```

3. Invoke Voila in directory X:
`voila .`

4. Open `test.ipynb` in the browser and let it run.

The rendering fails:

image

## Expected behavior

image

## Context

- voila version 0.3.0
- Operating System and version: Ubuntu 20.04
- Browser and version: Chrome 97
- JupyterLab version: 3

trungleduc:
## References

Close #1071 #1049 

## Code changes
Correctly redirect `voila/render/` to `voila/files/` for non-notebook items.


voila/handler.py
```diff
@@ -127,7 +127,10 @@ async def get_generator(self, path=None):
                 kernel_spec_manager=self.kernel_spec_manager,
             )
 
-            await gen.initialize(template=template_arg, theme=theme_arg)
+            done = await gen.initialize(template=template_arg, theme=theme_arg)
+            if not done:
+                self.redirect_to_file(path)
```
martinRenou, line 132: I am not sure I understand the PR, if the `path` was not a path to a Notebook file in the first place, maybe we should not use the `NotebookRenderer` class?

trungleduc: Indeed, it makes no sense to access non-notebook files through `/voila/render/` URL. But in some corner cases, where users have `IFrame` or `FileLink` widgets in their notebooks, these widgets will detect the URL to the local files from the current URL of the notebook, which is `/voila/render/...` 
So this PR will redirect these requests to the normal `/voila/files/`

martinRenou: I see. Could we detect this earlier in the code? Instead of trying to initialize a `NotebookRenderer` generator, we could probably redirect to the FileHandler somewhere around [here](https://github.com/voila-dashboards/voila/pull/1073/files#diff-7bc454003168bbff49060f6b1bef6e8d149654b85c18edbf5ca75f1720fdeeffR40)? 

I guess if the path does not end with `.ipynb` we are safe to try to redirect?

trungleduc: `Voila` can run text files by setting the mapping `VoilaConfiguration.extension_language_mapping`, so the file extension check would be more complicated.

martinRenou: Ok I see, thanks!

martinRenou (approved):

voila/handler.py
```diff
@@ -127,7 +140,9 @@ async def get_generator(self, path=None):
                 kernel_spec_manager=self.kernel_spec_manager,
             )
 
-            await gen.initialize(template=template_arg, theme=theme_arg)
+            done = await gen.initialize(template=template_arg, theme=theme_arg)
+            if not done:
+                raise tornado.web.HTTPError(404, 'NotebookRenderer has failed to initialize!')
```
martinRenou, line 145: Instead of doing this, it might be nicer that `initialize` fails properly. I'm thinking [this method](https://github.com/voila-dashboards/voila/blob/63e0f91cde88a62af9130b2008b06ee01a3e12a5/voila/notebook_renderer.py#L298) could throw a proper error if it fails:

```python
    async def load_notebook(self, path):

        model = await ensure_async(self.contents_manager.get(path=path))
        if 'content' not in model:
            raise tornado.web.HTTPError(404, 'file not found')
        __, extension = os.path.splitext(model.get('path', ''))
        if model.get('type') == 'notebook':
            notebook = model['content']
            notebook = await self.fix_notebook(notebook)
            return notebook
        elif extension in self.voila_configuration.extension_language_mapping:
            language = self.voila_configuration.extension_language_mapping[extension]
            notebook = await self.create_notebook(model, language=language)
            return notebook
        
         ## THROW HERE
         else:
            raise tornado.web.HTTPError(404, 'NotebookRenderer has failed to initialize')
```

?

Also I'm wondering if a `404` is the right error? It feels like a file was found, but we failed to process it, should it be a `500`?

trungleduc: Updated

voila/notebook_renderer.py
```diff
@@ -46,8 +45,14 @@ def __init__(self, **kwargs):
...
         self.rendered_cache: List[str] = []
 
-    async def initialize(self, **kwargs):
+    async def initialize(self, **kwargs) -> None:
+        """ Initialize the notebook generator, this method will
+        return `True` if the generator is initialized, return `False`
+        otherwise.
 
+        Returns:
+            Boolean: The flag to check if the generator is initialized.
```
martinRenou, line 54: This method does not return anything anymore

voila/handler.py
martinRenou: Thanks!

martinRenou:
Thanks!

PR #1083, opened 2022-01-28
Add bot for updating Galata references

martinRenou:
Add a bot for automatically updating Galata references, the bot can be triggered by commenting `Please update galata references` in the PR.


martinRenou:
Testing the bot in https://github.com/martinRenou/voila/pull/2

trungleduc:
Great @martinRenou!, and it has been a long time to see an all green voila PR 😄

jtpio:
Thanks!

If we start using this workflow in several places, maybe it could be interesting to make it a reusable action like the ones in https://github.com/jupyterlab/maintainer-tools?

.github/workflows/update_galata_references.yaml
```diff
@@ -0,0 +1,73 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Checkout the branch from the PR that triggered the job
+        run: hub pr checkout ${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jtpio, line 24: As noticed in https://github.com/jupyterlab/maintainer-tools/pull/39#issue-1111688219, it looks like this would make it possible for anyone to commit and push to the PR?

jtpio:
> If we start using this workflow in several places, maybe it could be interesting to make it a reusable action like the ones in https://github.com/jupyterlab/maintainer-tools?

Or maybe this needs we could even reuse the existing PR script action from https://github.com/jupyterlab/maintainer-tools#pr-script-1.

And encapsulate the logic to update the Galata screenshots so it's callable via a script.

.github/workflows/update_galata_references.yaml
martinRenou: Yes, but I am not sure the concerns about adding/executing arbitrary code apply here. The action really just commits the test references (it's hardcoded), so yeah anyone could comment "update galata references" and update the references, but that should not do much harm.

martinRenou: And I think it's actually important for anyone to be able to do that. Because new contributors should be able to trigger that action in their own PR.

jtpio: Alright, this workflow is indeed more specific than the "PR Script" workflow provided by https://github.com/jupyterlab/maintainer-tools

martinRenou: Your comment about adding such a tool in maintainer-tools makes sense, we should probably not copy paste this everywhere Galata is used. I'll see if I can make something.

jtpio:
Is this PR still relevant now that https://github.com/jupyterlab/maintainer-tools/pull/41 has been merged?

.github/workflows/update_galata_references.yaml
martinRenou: Done in https://github.com/jupyterlab/maintainer-tools/pull/41

martinRenou:
@jtpio I guess it's ready now :)

trungleduc (approved):

PR #1089, opened 2022-02-03
Lab theme handling

martinRenou:
## References

Builds on top of https://github.com/jupyter/nbconvert/pull/1703

## Code changes

Add handlers for the JupyterLab federated extension themes. This adds a dependency on `jupyterlab_server` which provides the `ThemesHandler`.

## User-facing changes

```bash
voila --theme=jupyterlab_miami_nights
```

https://user-images.githubusercontent.com/21197331/152327655-5cab27ad-7147-4730-946a-2f8dff1cd304.mp4

None


martinRenou:
I need to update the visual regression test references (due to nbconvert change). Will wait for the bot to be added in https://github.com/voila-dashboards/voila/pull/1083.

martinRenou:
One of the advantages of properly serving the theme assets (CSS, images, fonts) is that it allows the browser to cache them. While if we embed them in the HTML the browser will not be able to cache them.

In the case of nbconvert we want the resulting HTML file to be portable (we want to be able to render the HTML file without having to look for the CSS assets somewhere else), that's why embedding makes sense. But in the case of Voila we have a proper server, so we should probably use it.

martinRenou:
Also, adding `jupyterlab_server` might seem overkill just for this PR, but it will also be useful later for when Voila is a JupyterLab remix.

trungleduc:
Thanks for the explanation! I forgot about the caching, it totally makes sense to serve assets via URL.

martinRenou:
Sure! :)

I should probably add a visual regression test to this PR

trungleduc:
> Sure! :)
> 
> I should probably add a visual regression test to this PR

Indeed, I think visual regression tests are needed for this kind of change.

martinRenou:
Update galata references

martinRenou:
Bot running in https://github.com/voila-dashboards/voila/actions/runs/1828426920

martinRenou:
Update galata references

martinRenou:
Added support for query parameters on the TreeHandler, it will keep the query when changing directory and rendering Notebooks:


https://user-images.githubusercontent.com/21197331/153574688-dd631db6-ec30-43dc-b939-81d576aee275.mp4

martinRenou:
Update galata references

ui-tests/tests/voila.test.ts
```diff
@@ -36,13 +123,55 @@ test.describe('Voila performance Tests', () => {
...
 
+  test('Render basics.ipynb with dark theme', async ({
+    page,
+    browserName
+  }, testInfo) => {
+    const notebookName = 'basics';
+    const testFunction = async () => {
+      await page.goto(`/voila/render/${notebookName}.ipynb?voila-theme=dark`);
+      // wait for the widgets to load
+      // await page.waitForSelector('.jupyter-widgets');
```
jtpio, line 134: Maybe we can remove these comments if they are not used anymore?

voila/treehandler.py
```diff
@@ -76,7 +87,8 @@ def allowed_content(content):
                        contents=contents,
                        terminals_available=False,
                        server_root=get_server_root_dir(self.settings),
-                       theme=self.voila_configuration.theme,
+                       theme=theme_arg,
+                       query=self.request.query,
```
trungleduc, line 91: I'm not sure that we want to pass all queries of the tree page to the notebook URLs, maybe just the theme and template?

martinRenou: As of this PR, that's the only two query arguments we support, or do we support more arguments?

trungleduc: `self.request.query` contains all the queries passed to the tree handler, and then it is appended to the notebook URLs. I'm thinking about a more limited scope where only  `voila-theme` and `voila-template` are transferred to the URL

trungleduc (approved):

voila/treehandler.py
martinRenou: I think your original comment will make sense if we add some query arguments specific to the TreeHandler at some point. But we can probably filter the query arguments here only when that happens?

trungleduc: If `Voila` is served under proxy, there might be some sort of token query for the tree handler, but I think there is no harm in transferring these tokens to the notebook URLs. We can filter the query when we actually use them in the tree handler.

trungleduc:
Thanks @martinRenou

PR #1094, opened 2022-02-14
small fixes to jupyterlab voila extension  

legendb317:
## References
n/a

#### Change 1
Issue: If a user makes an edit without saving and refreshes voila, voila does not reflect the updated cells
Fix: Save the notebook on refresh
(Note: Saving the notebook when using voila already occurs when the voila extension is launched)

#### Change 2
Issue: "Render on Refresh" checkbox is misaligned 
Fix: Update style
(Note: Unless something is wrong with my dev environment, I noticed that changes to `jp-VoilaPreview-renderOnSave` in the `base.css` were not reflected)

#### Change 1
Users will no longer need to save a notebook before refreshing voila

##### Before 
image

##### After 
image

n/a


jtpio:
Thanks @legendb317 :+1: 

> (Note: Unless something is wrong with my dev environment, I noticed that changes to jp-VoilaPreview-renderOnSave in the base.css were not reflected)

Normally the changes should be reflected and there seems to be some CSS applied to that class already:

https://github.com/voila-dashboards/voila/blob/6be66cc07e18a7a7caf44b11e8c5af24dbe5b8de/packages/jupyterlab-preview/style/base.css#L5-L7

If you run the watch script, does `webpack` incrementally recompiles the extension after making a change to the css file?

packages/jupyterlab-preview/src/preview.tsx
```diff
@@ -98,14 +98,18 @@ export class VoilaPreview extends DocumentWidget<IFrame, INotebookModel> {
       icon: refreshIcon,
       tooltip: 'Reload Preview',
       onClick: () => {
-        this.reload();
+        context.save().then(() => {
```
jtpio, line 101: Minor styling note, we could also use `async` / `await` here for readability:

```ts
onClick: async () => {
  await context.save();
  this.reload();
}
```

legendb317: 👍

legendb317:
> Thanks @legendb317 👍
> 
> > (Note: Unless something is wrong with my dev environment, I noticed that changes to jp-VoilaPreview-renderOnSave in the base.css were not reflected)
> 
> Normally the changes should be reflected and there seems to be some CSS applied to that class already:
> 
> https://github.com/voila-dashboards/voila/blob/6be66cc07e18a7a7caf44b11e8c5af24dbe5b8de/packages/jupyterlab-preview/style/base.css#L5-L7
> 
> If you run the watch script, does `webpack` incrementally recompiles the extension after making a change to the css file?

The `webpack` recompiles the extension after a change to the css file. However, the styling that is in the `css` file does not show up when I inspect the page.

trungleduc:
I think we have some issues with the CSS of the extension, the CSS file is compiled with `webpack` but the content is not included on the page. I'm working on it.

jtpio:
Ah ok, thanks both for looking into it!

legendb317:
> Since the CSS issue is not the main point of this PR, I think we can merge this one and open another issue for the CSS file bug. What do you think @jtpio @legendb317 ?

Sounds good to me.

jtpio:
This should be fixed in https://github.com/voila-dashboards/voila/pull/1096 now.

@legendb317 mind updating this PR and check if that fixes it? Thanks!

legendb317:
@jtpio I have updated the PR and the CSS still seems broken

legendb317:
> @legendb317 I just tested locally, the CSS file seems to be working fine.
> 
> voila-css

sorry, I forgot to do an install after the update to https://github.com/voila-dashboards/voila/pull/1096. The fix does work and I have pushed the changes to this PR.

jtpio (approved):
Thanks!

PR #1096, opened 2022-02-17
Bump `css-loader` and `style-loader``

### LINKED ISSUE #1095: Styles defined in  `jupyterlab-preview`  do not work.
## Description

As stated in #1094  the styles defined in the `base.css` of  `jupyterlab-preview`  are not taken into account in Jupyter Lab. For some reason, the content of this file is not included on the lab page.

## Reproduce

1. Open jupyter lab
2. Open the `Elements` tab of the browser's inspection tool.
3. Search for the classes defined in `base.css`
4. Found nothing

## Context

- voila version  0.3.1

trungleduc:
Fix #1095


jtpio (approved):
Thanks for catching this!

jtpio:
A follow-up to this would be to check if we can realign the dependencies with the ones specified in the cookiecutter: https://github.com/jupyterlab/extension-cookiecutter-ts/blob/3.0/%7B%7Bcookiecutter.python_name%7D%7D/package.json

We might also need to double check the dependencies of the voila frontend app: https://github.com/voila-dashboards/voila/blob/main/packages/voila/package.json

PR #1106, opened 2022-03-07
Add timeout on the control comm target

martinRenou:
So that if we don't get a response in time, we fallback to the old implementation of fetching the widgets one by one.

## References

Similar to what has been done in https://github.com/jupyter-widgets/ipywidgets/pull/3335/files. Once we reuse the base manager, we can get rid of this code in Voila.

## Code changes

Add a timeout so that the Promise gets rejected if we didn't get a response on the control comm target in time.

This can happen with Notebook <=6.1, if the `comm_open` message is not allowed, the handler will simply ignore our request to open on the control comme target (see [this](https://github.com/jupyter/notebook/blob/6.1.x/notebook/services/kernels/handlers.py#L432-L435)), resulting in never getting a response/close on the front-end, and the promise gets stuck.

## User-facing changes

None

None


packages/voila/src/manager.ts
```diff
@@ -238,6 +243,12 @@ export class WidgetManager extends JupyterLabManager {
 
         // Send a states request msg
         initComm.send({ method: 'request_states' }, {});
+
+        // Reject if we didn't get a response in time
```
trungleduc, line 247: Could you add `TODO` here to remove this code in the future?

martinRenou: Are you referring to my comment?

> Once we reuse the base manager, we can get rid of this code in Voila.

The entire method `_build_models` will be removed once https://github.com/jupyter-widgets/ipywidgets/pull/3335 gets released.

trungleduc: Thanks for clarifying!

trungleduc (approved):

trungleduc:
Thanks @martinRenou!

martinRenou:
Thank you for reviewing and merging!

Issue #1107, opened 2022-03-08
Using --enable_nbextensions=True in v0.3.2 shows an error in console


mariobuikhuizen:
## Description

Using `--enable_nbextensions=True` in v0.3.2 shows an error in console:
```
ERROR:tornado.application:Uncaught exception in write_error
Traceback (most recent call last):
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/tornado/web.py", line 2576, in get
    self.absolute_path = self.validate_absolute_path(self.root, absolute_path)
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/jupyter_server/base/handlers.py", line 860, in validate_absolute_path
    raise web.HTTPError(404)
tornado.web.HTTPError: HTTP 404: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/tornado/web.py", line 1217, in send_error
    self.write_error(status_code, **kwargs)
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/jupyter_server/base/handlers.py", line 590, in write_error
    html = self.render_template("%s.html" % status_code, **ns)
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/jupyter_server/base/handlers.py", line 521, in render_template
    return template.render(**ns)
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/jinja2/environment.py", line 1291, in render
    self.environment.handle_exception()
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/lib/python3.10/site-packages/jinja2/environment.py", line 925, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/share/jupyter/voila/templates/base/404.html", line 1, in top-level template code
    {% extends "error.html" %}
  File "/Users/mariobuikhuizen/miniconda3/envs/voila32/share/jupyter/voila/templates/base/error.html", line 1, in top-level template code
    {% extends "page.html" %}
[...]

PR #1108, opened 2022-03-09
Add theme handling to error pages

martinRenou:
## Code changes

Add theme handling to error pages. Also fixes an issue mentioned in https://github.com/voila-dashboards/voila/issues/1107 where the error page fails to render due to `include_css` function not being defined.

## User-facing changes

Error pages should be improved visually.

None


martinRenou:

martinRenou:
[1ba7791](https://github.com/voila-dashboards/voila/pull/1108/commits/1ba77919f9063055f0e93e6fe7c94586b2b3f502) commit should fix https://github.com/voila-dashboards/voila/issues/1107. It is unclear to me why the `jupyter_server.base.handlers.FileFindHandler` is even trying to render the `base.html` template, that seems wrong?

voila/handler.py
```diff
@@ -192,6 +192,30 @@ async def get(self, path=None):
             self.write(html)
             self.flush()
 
+    def render_template(self, name, **ns):
```
trungleduc, line 195: Since `render_template` is not used directly in `get`, can you add some docstrings explaining why we need to add this method?

martinRenou: Yes! Thanks for the review.

martinRenou:
@trungleduc I added a small docstring and did a bit of refactoring

jtpio (approved):
Looks good, thanks!

trungleduc (approved):

PR #1109, opened 2022-03-10
feat: make request info available in env for preheated kernels

mariobuikhuizen:
## Code changes

The mechanism that was used for 'get_query_string()' is changed so
that all request info is sent. For backward compatibility
'get_query_string()' is reimplemented by using the changed code.

## User-facing changes

Request info is now available in preheated kernels in the same way as in on-demand kernels after calling `voila.utils.wait_for_request()` in the notebook.


voila/handler.py
```diff
@@ -150,6 +150,7 @@ def time_out():
 
                 return '<script>voila_heartbeat()</script>\n'
 
+            kernel_env = {**os.environ.copy(), **request_info}
```
maartenbreddels, line 153:
```suggestion
            kernel_env = {**os.environ, **request_info}
```
I don't think the copy adds anything, since we already create a new dict

trungleduc:
@mariobuikhuizen thanks a lot! I found it more natural to use than the old implementation. Since `get_query_string` is quite new, I think we could remove it and use only the new method.

voila/request_info_handler.py
```diff
@@ -26,28 +26,28 @@ def open(self, kernel_id: str) -> None:
...
-    def send_updates(cls: 'QueryStringSocketHandler', msg: Dict) -> None:
-        """Class method used to dispath the query string to the waiting
-        notebook. This method is called in `VoilaHandler` when the query
-        string becomes available.
+    def send_updates(cls: 'RequestInfoSocketHandler', msg: Dict) -> None:
+        """Class method used to dispath the request info to the waiting
+        notebook. This method is called in `VoilaHandler` when the request
+        info becomes available.
         If this method is called before the opening of websocket connection,
         `msg` is stored in `_cache0` and the message will be dispatched when
```
trungleduc, line 45:
```suggestion
        `msg` is stored in `_cache` and the message will be dispatched when
```

mariobuikhuizen: This is not a line I've changed or part of this PR, but it's a small enough change to put in. Just checking if you're sure this should be changed in this PR.

trungleduc: Thanks for fixing it, this is the repo I left from previous PR 😄

trungleduc (approved):

trungleduc:
Thanks @mariobuikhuizen

PR #1129, opened 2022-03-28
feat: support all kernel ServerConnection options

maartenbreddels:
We only exposed baseUrl and kernelId in connectKernel, with this we expose all ServerConnection options.


trungleduc:
@maartenbreddels can you rebase the PR to get the `jinja2` fix from `main` ?

trungleduc:
Thanks @maartenbreddels

PR #1136, opened 2022-03-30
Fix UI-tests detached script

martinRenou:


trungleduc:
Thanks @martinRenou !

Issue #1047, opened 2021-12-18
Deploying to Heroku not working on v0.3


samharrison7:
## Description

I've been struggling to get Voila to deploy to Heroku recently, and have just figured that I am only having issues where I'm using version 0.3. For example, using the template repo (https://github.com/voila-dashboards/voila-heroku) now doesn't work on Heroku as it installs the 0.3 version (I had to update the runtime.txt to use 3.8.10 as 3.7.5 isn't available on the default Heroku-20 stack).

Specifically I get a timeout error in the Heroku logs. It looks like the Voila server starts up okay but doesn't manage to send anything when a page is requested:

```
2021-12-18T13:09:03.238707+00:00 heroku[web.1]: Starting process with command `voila --port=26507 --no-browser --template=material --enable_nbextensions=True notebooks/bqplot.ipynb`
2021-12-18T13:09:05.949936+00:00 app[web.1]: [Voila] Using /tmp to store connection files
2021-12-18T13:09:05.950308+00:00 app[web.1]: [Voila] Storing connection files in /tmp/voila_l56jo0cr.
2021-12-18T13:09:05.950375+00:00 app[web.1]: [Voila] Serving static files from /app/.heroku/python/lib/python3.8/site-packages/voila/static.
2021-12-18T13:09:06.180485+00:00 app[web.1]: [Voila] Voilà is running at:
2021-12-18T13:09:06.180486+00:00 app[web.1]: http://localhost:26507/
2021-12-18T13:09:13.000000+00:00 app[api]: Build succeeded
2021-12-18T13:10:03.342378+00:00 heroku[web.1]: Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch
2021-12-18T13:10:03.368529+00:00 heroku[web.1]: Stopping process with SIGKILL
2021-12-18T13:10:03.528533+00:00 heroku[web.1]: Process exited with status 137
2021-12-18T13:10:03.609381+00:00 heroku[web.1]: State changed from starting to crashed
```

If I change the `requirements.txt` to `voila==0.2.16`, it works as expected. Running Heroku locally also works (`heroku local`) as expected, even with version 0.3. I tried to set max boot time to 180 secs and it still didn't work.

## Reproduce

To reproduce, follow the instructions in https://github.com/voila-dashboards/voila-heroku, but change the `runtime.txt` file to `python-3.8.10` (or anything else on Heroku-20).

trungleduc:
[...]

PR #1158, opened 2022-05-31
Update deploy to Heroku instructions

samharrison7:
As per issue #1047, Voila v0.3 requires `--Voila.ip=0.0.0.0` to be explicitly set when deploying to Heroku. I have updated the deploy.rst docs to reflect this. 

I also updated the Python runtime to the latest version that is supported by all Heroku stacks.

## References

#1047 

## Code changes

Only changes to docs as described above.

## User-facing changes

No user facing changes (except to documentation).

No backwards incompatible changes.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -113,20 +113,20 @@ High level instructions, specific to Voilà can be found below:
 
    .. code:: text
 
-       python-3.9.9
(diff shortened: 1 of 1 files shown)


trungleduc:
Thanks!

trungleduc (approved):

PR #1161, opened 2022-06-16
Update nbconvert pinning

### LINKED ISSUE #1160: Test against nbconvert master
### Problem

nbconvert 7.x is coming, we should probably run the CI against the master branch just to make sure everything is working fine.

https://github.com/jupyter/nbconvert/releases

martinRenou:
Update nbconvert pinning

## References

Resolve #1160


jtpio:
Maybe this PR could now be about raising the upper bounds for `nbclient` and `nbconvert`?

https://github.com/voila-dashboards/voila/blob/55d0cb005d1c89dbf42c680ece7d682ce4d7b042/setup.cfg#L39-L40

pyproject.toml
```diff
@@ -38,8 +38,8 @@ dependencies = [
     "jupyter_core>=4.11.0",
     "jupyter_server>=1.18,<2.0.0",
     "jupyterlab_server>=2.3.0,<3",
-    "nbclient>=0.4.0,<0.6",
-    "nbconvert>=6.4.5,<7",
+    "nbclient>=0.4.0,<0.7",
```
martinRenou, line 41:
```suggestion
    "nbclient>=0.4.0,<0.6",
```

martinRenou:
I can't understand yet why the tests fail in the CI with this `AttributeError: module 'mistune' has no attribute 'Renderer'`. 

I'm also runtime Mistune 2.0.4 locally and it works just fine to run the tests.

jtpio:
I just tried in a clean environment (Gitpod) and also getting the same error as in the CI tests:

image

image

martinRenou:
I was in this weird installation where `conda list` would show me Mistune 2.0.4 while I was actually using 0.8.

martinRenou:
The failing test seems to be due to the change in validate/normalize logic introduced by https://github.com/jupyter/nbconvert/commit/8c1146d756175824dfb1e939d86fddfcc3979fdd and https://github.com/jupyter/nbformat/pull/282.

martinRenou:
> The failing test seems to be due to the change in validate/normalize logic introduced by https://github.com/jupyter/nbconvert/commit/8c1146d756175824dfb1e939d86fddfcc3979fdd and https://github.com/jupyter/nbformat/pull/282.

Indeed the failing test was originated from those, because nbconvert is now less permissive with invalid Notebooks, and papermill generates an invalid Notebook (see https://github.com/nteract/papermill/issues/686#issuecomment-1266508548). Bumping our test Notebook version fixes it.

voila/exporter.py
```diff
@@ -16,6 +16,11 @@
 except ImportError:
     from jinja2 import contextfilter as pass_context
 
+try:
+    from mistune import Renderer as MistuneRenderer
```
jtpio, line 20: Maybe we can add a quick comment to specify which version of `mistune` requires this import?

So it can be removed at some point later when the minimum of `mistune` exports `HTMLRenderer`.

martinRenou: Actually we should probably require `markdown_renderer_class` to be: 
```python
    markdown_renderer_class = traitlets.Type(VoilaMarkdownRenderer).tag(config=True)
```

Because the `markdown_renderer_class` class must have a constructor which accepts a `contents_manager` as parameter. This would have a side effect that we don't need to import anything from Mistune.

jtpio: Nice, even better.

pyproject.toml
```diff
@@ -38,8 +38,8 @@ dependencies = [
     "jupyter_core>=4.11.0",
     "jupyter_server>=1.18,<2.0.0",
     "jupyterlab_server>=2.3.0,<3",
-    "nbclient>=0.4.0,<0.7",
-    "nbconvert>=6.4.5,<7",
+    "nbclient>=0.4.0,<0.6",
```
martinRenou, line 41:
```suggestion
    "nbclient>=0.4.0,<0.7",
```

martinRenou: Issue while rebasing

jtpio (approved):
Thanks!

jtpio:
Starting a new pre-release now: https://github.com/voila-dashboards/voila/pull/1225

pyproject.toml
bnavigator: nbclient 0.7.0 is out. Please make it compatible before releasing voila 0.4

martinRenou: Thanks for the heads up!

PR #1163, opened 2022-06-27
Update jupyter_server

martinRenou:
## References

This brings in multiple security fixes


trungleduc:
The failed tests on `Windows` can be fixed by https://github.com/jupyter/jupyter_core/pull/271, we need to wait for a new release of `jupyter_core` and then pin its latest version.

PR #1167, opened 2022-07-13
adding nbextension path to voila config object

dharmaquark:
## Code changes

This PR introduces `nbextensions_path` as a configurable value that we can set to override the default Jupyter Server value (as a workaround until Voila moves away entirely from nbextensions to adopt the lab extension system)

### Background on Motivation (credit @SylvainCorlay)
While Voilà is built with JupyterLab components, we use the classic notebook extension system for the widgets, because when Voilà came out, JupyterLab could not use pre-built extensions.
By default, Voilà uses the webpack bundles that are published on npmjs.
But in order to support extensions that have never been published, but are installed locally, we added support for using local notebook extensions.
Jupyter server is the raw server, and it does not come with the notebook frontend, hence it does not have a notion of notebook extensions. As a workaround for applications that have a nbextension path different from the default, we are adding an option `nbextensions_path` to the Voila Configuration object, so that we are not tied to the server default. 

## User-facing changes

introduces a new optional configuration value that users can set to override default server nbextensions path

N/A

### DIFF
--- a/voila/app.py
+++ b/voila/app.py
@@ -131,6 +131,7 @@ class Voila(Application):
         'base_url': 'Voila.base_url',
         'server_url': 'Voila.server_url',
         'enable_nbextensions': 'VoilaConfiguration.enable_nbextensions',
+        'nbextensions_path': 'VoilaConfiguration.nbextensions_path',
         'show_tracebacks': 'VoilaConfiguration.show_tracebacks',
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Thanks for making the PR. I am happy to merge as it looks good to me.

PR #1175, opened 2022-07-22
Add Voila configuration option for default kernel environmental variables

dcnadler:
Currently, to use [partially pre-rendered notebooks](https://voila.readthedocs.io/en/stable/customize.html#partially-pre-render-notebook) a user may have to set environmental variables in notebook kernels to access the correct websocket address.

The docs give the example of using `voila.json` to set these. However, for some deployment platforms these values (e.g. base_url or port) may not be known ahead of time or differ between deployments. Having a command line option provides easier passing of env variables (as opposed to modifying `voila.json` in-place).

## Code changes

A `default_kernel_env_variables` attribute is added to `VoilaConfiguration`, which is then passed to the default `kernel_pools_config` of `VoilaKernelManager`.

### DIFF
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -323,7 +323,7 @@ Preheated kernels
 ==================
 
 Since Voilà needs to start a new jupyter kernel and execute the requested notebook in this kernel for every connection, this would lead to a long waiting time before the widgets can be displayed in the browser. 
-To reduce this waiting time, especially for the heavy notebooks, users can activate the preheating kernel option of Voilà, this option will enable two features:
+To reduce this waiting time, especially for heavy notebooks, users can activate the preheating kernel option of Voilà, this option will enable two features:
 
 - A pool of kernels is started for each notebook and kept in standby, then the notebook is executed in every kernel of its pool. When a new client requests a kernel, the preheated kernel in this pool is used and another kernel is started asynchronously to refill the pool.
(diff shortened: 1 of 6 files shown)


trungleduc:
Thanks @dcnadler for your contribution, for now, setting the default environmental variables for the pre-heated kernel is possible by configuring the `VoilaKernelManager` class. For example, this command will start kernels with the variable `foo`

```
voila  --preheat_kernel=true --VoilaKernelManager.kernel_pools_config="{'default':{'kernel_env_variables':{'foo':'bar'}}}"
```

It's pretty verbose so I think we can make `default_kernel_env_variables`  a setting of `VoilaKernelManager` instead of `VoilaConfiguration`

dcnadler:
Thanks @trungleduc! Did not realize that was an option. I am new to trailets.

Seems like trailets may not let you pass a nested dict in the future?

```
/anaconda3/envs/voila/lib/python3.10/site-packages/traitlets/traitlets.py:3258: FutureWarning: --kernel_pools_config={"default":{"kernel_env_variables":{"FOO":"BAR"}}} for dict-traits is deprecated in traitlets 5.0. You can pass --kernel_pools_config <key=value> ... multiple times to add items to a dict.
```

I moved the `default_env_variables` to the `VoilaKernelManager` as you suggested. Let me know what you think.

trungleduc (approved):

trungleduc:
Thanks!

PR #1141, opened 2022-04-08
Fix default VOILA_WS_BASE_URL value in preheating mode

vkaidalov-rft:
## References

Resolves #1140.

## Code changes

Use `VOILA_BASE_URL` as the default value if `VOILA_WS_BASE_URL` has not been specified to make the `wait_for_request` and `get_query_string` functions backwards-compatible in the preheating mode.

PR #1176, opened 2022-07-23
Add Voila app port to kernel env

dcnadler:
Similar to https://github.com/voila-dashboards/voila/pull/1141, which added `VOILA_SERVER_URL` to the kernel env, this adds `VOILA_APP_PORT`.

The purpose is to make using preheated kernels with `wait_for_request()` easier when not using the default port.

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -187,6 +187,7 @@ class VoilaHandler(BaseVoilaHandler):
             kernel_env[ENV_VARIABLE.VOILA_PREHEAT] = 'False'
             kernel_env[ENV_VARIABLE.VOILA_BASE_URL] = self.base_url
             kernel_env[ENV_VARIABLE.VOILA_SERVER_URL] = self.settings.get('server_url', '/')
+            kernel_env[ENV_VARIABLE.VOILA_APP_PORT] = str(port) if port else ''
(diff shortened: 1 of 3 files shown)


voila/handler.py
```diff
@@ -187,6 +187,7 @@ def time_out():
             kernel_env[ENV_VARIABLE.VOILA_PREHEAT] = 'False'
             kernel_env[ENV_VARIABLE.VOILA_BASE_URL] = self.base_url
             kernel_env[ENV_VARIABLE.VOILA_SERVER_URL] = self.settings.get('server_url', '/')
+            kernel_env[ENV_VARIABLE.VOILA_APP_PORT] = str(port) if port else ''
```
trungleduc, line 190:
```suggestion
```

The value of `port` here is the port that `Voila` gets from the incoming request, not the port on which `Voila` is listening locally (for example in the case of `Voila` is served behind a proxy server). 
But we don't need to set the variable `ENV_VARIABLE.VOILA_APP_PORT` here since we are not in the preheated kernel mode.

dcnadler:
Thanks for clarifying @trungleduc! I updated the code

trungleduc:
Thanks!

PR #1184, opened 2022-08-10
Cleanup exporter

martinRenou:
Some code was duplicated from `TemplateExporter.from_notebook_node` that we actually call.

See https://github.com/jupyter/nbconvert/blob/main/nbconvert/exporters/templateexporter.py#L386-L398

Maybe @maartenbreddels knows why this code was around? Or is this a leftover?


martinRenou:
My bad, I didn't read the code carefully. Closing.

PR #1185, opened 2022-08-10
Fix inifinite  loop in property access

martinRenou:
A subclass of VoilaExporter would create an infinite loop with this

```python
class Foo(VoilaExporter):
    pass
```

Causes:
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```


martinRenou:
UI-tests failures are related to https://github.com/jupyter/nbconvert/pull/1829

martinRenou:
UI-tests failures are related to https://github.com/jupyter/nbconvert/pull/1829

PR #1187, opened 2022-08-12
Voilite POC

martinRenou:
POC implementation for Voilite. Building on top of the lab-based PR https://github.com/voila-dashboards/voila/pull/846

This PR implements a Voilite nbconvert exporter that will execute the code in the client using a client-side kernel.

```
jupyter nbconvert --to voilite_dashboard notebooks/basics.ipynb
```

## Update 18/11
The `nbconvert` template is replaced by the `voilite` command. Users can convert a notebook into a static dashboard  page with 
 
```
voilite notebook.ipynb --theme=dark --template=gridstack --packages=ipywidgets --packages=ipyleaflet
```
Supported options: 
- `--theme`: Change the theme , can be `dark`, `light` or the full name of the JupyterLab themes. 
- `--template`: Change the template, for now `Voilite` supports `lab`, `gridstack`, `reveal` and `classic` theme
- `--packages`: List of packages to be installed at run time in the lite kernel. 

`Voilite` can be configured via the CLI or the `voilite.py/voilite.json` file.

Remaining tasks 

- [x]  Update `classic` and `reveal` templates to support `voilite`.
- [x] Handle static files
- [x]  Generate a tree page for the case of multiple notebooks.
- [x]  Improve loading message.
- [ ] ~Add tests and documentation~ (will be addressed in the follow-up PR )

voilite

PR #1190, opened 2022-08-22
Troubleshoot docs build

jtpio:
## References

Over in https://github.com/voila-dashboards/voila/pull/846 it seems the docs build has started failing on ReadTheDocs:

```
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
Collecting package metadata: ...working... failed

UnavailableInvalidChannel: The channel is not accessible or is invalid.
  channel name: pkgs/main
  channel url: https://repo.anaconda.com/pkgs/main
  error code: 404

You will need to adjust your conda configuration to proceed.
Use `conda config --show channels` to view your configuration's current state,
and use `conda config --show-sources` to view config file locations.
```

## Code changes

Update docs dependencies.

## User-facing changes

None

None

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -4,4 +4,5 @@ channels:
   - conda
 dependencies:
   - sphinx
+  - pip
   - pydata-sphinx-theme
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,6 +1,7 @@
 name: voila
 channels:
   - conda-forge
+  - nodefaults
 dependencies:
   - sphinx
   - pip
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,7 +1,6 @@
 name: voila
 channels:
   - conda-forge
-  - conda
 dependencies:
   - sphinx
   - pip
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -7,6 +7,5 @@ dependencies:
   - bqplot
   - scipy
   - ipympl
-  - ipympl
   - xleaflet=0.16.0
   - xeus-cling=0.13.0
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -72,7 +72,7 @@ jobs:
             py_cmd: python
     steps:
       - name: Install Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
           architecture: 'x64'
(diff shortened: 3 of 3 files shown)


jtpio:
This seems to be happening on this PR as well, which means it's reproducible:

image

jtpio:
`nodefaults` did the trick: https://github.com/voila-dashboards/voila/pull/1190/commits/dbd2b514841de8af58236406095ea4bb5f483dd0

jtpio:
The failing macos check looks unrelated: https://github.com/voila-dashboards/voila/runs/7952356006?check_suite_focus=true

```
ModuleNotFoundError: No module named 'jupyter_packaging'
```

For some reasons it seems to be picking up pip for Python 3.10 instead of 3.7:

image

Not sure what's happening. It might be a change in GitHub Actions runners :shrug: 
Also seeing this in https://github.com/jtpio/ipylab/pull/100

docs/environment.yml
```diff
@@ -1,7 +1,8 @@
 name: voila
 channels:
   - conda-forge
-  - conda
+  - nodefaults
```
martinRenou, line 4: I'm really surprised by this. Shouldn't we only use the `conda-forge` channel?

martinRenou:
> For some reasons it seems to be picking up pip for Python 3.10 instead of 3.7:

It's probably an issue with the setup-python action? Should we only rely on micromamba to provide our Python?

jtpio:
The latest commit on master shows something similar with mismatched Python versions, although it passed: https://github.com/voila-dashboards/voila/runs/7783033666?check_suite_focus=true

image

docs/environment.yml
jtpio: No idea. It could have been a temporary issue on ReadTheDocs as well.

I also saw this change in https://github.com/jupyterlab/jupyterlab_server/pull/296

jtpio:
Closing as the docs build seems to be working on ReadTheDocs now (for example in https://github.com/voila-dashboards/voila/pull/1187). Probably something on ReadTheDocs changed in the meantime.

Not sure how to fix the failing macos check. Over in https://github.com/jupyter/notebook/pull/6523 switching to `bash -e {0}` was enough. I'll open a separate PR to continue investigating.

PR #1191, opened 2022-08-26
Troubleshoot failing `macos` check on CI

jtpio:
## References

Continuation of #1190 

## Code changes

CI fixes.

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -72,7 +72,7 @@ jobs:
             py_cmd: python
     steps:
       - name: Install Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
           architecture: 'x64'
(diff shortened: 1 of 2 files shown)


jtpio:
OK so all the tests are now failing, this looks new...

But the macos install check now passes:

image

With the correct Python version now being used across steps:

image

.github/workflows/packaging.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           path: ./dist
       - name: Install the prerequisites
         run: |
-          ${{ matrix.py_cmd }} -m pip install pip wheel
+          ${{ matrix.py_cmd }} -m pip install pip wheel jupyterlab~=3.0 notebook~=6.4
```
jtpio, line 85: Not sure how the extension checks could have worked before without these packages :thinking:

jtpio:
Marking as ready as this at least fixes the macos install check.

The other failures still need a fix too but they look unrelated to this change: https://github.com/voila-dashboards/voila/pull/1192#issuecomment-1228433656

trungleduc:
Thanks @jtpio !

PR #1193, opened 2022-08-29
Fix CI 

trungleduc:
## References
Follow up on https://github.com/voila-dashboards/voila/pull/1191

### DIFF
--- a/tests/notebooks/output.ipynb
+++ b/tests/notebooks/output.ipynb
@@ -16,8 +16,9 @@
        "Output()"
       ]
      },
+     "execution_count": 1,
      "metadata": {},
-     "output_type": "display_data"
+     "output_type": "execute_result"
     }
    ],
    "source": [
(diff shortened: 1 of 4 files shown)


tests/notebooks/output.ipynb
```diff
@@ -16,8 +16,9 @@
        "Output()"
       ]
      },
+     "execution_count": 1,
      "metadata": {},
-     "output_type": "display_data"
+     "output_type": "execute_result"
```
jtpio, line 21: Since the diff also includes formatting changes, wondering if this switch of `output_type` is the one that did the trick?

trungleduc: Yes, since we are testing against `ipywidgets 8`, the output of the widget is now of type `execute_result` instead of `display_data`.

jtpio: Nice thanks :+1:

PR #1046, opened 2021-12-16
feat: support for ipywidgets 8

mariobuikhuizen:
## Code changes
- update dependencies to ipywidgets 8
- use renamed method and property

PR #1197, opened 2022-09-06
Migrate to `hatch`

jtpio:
## References

Kickstarting the migration to `hatch` using the migration script: https://hatch-jupyter-builder.readthedocs.io/en/stable/source/how_to_guides/migrating_javascript_projects.html

```
pip install hatch_jupyter_builder
python -m hatch_jupyter_builder.migrate .
```

More info: https://blog.jupyter.org/packaging-for-jupyter-in-2022-c7be64c38926

## Code changes

- [x] Update to `hatch`.
- [x] Use `hatch` to bump the Python version
- [x] Handle CSS files downloaded in `setup.py`:

https://github.com/voila-dashboards/voila/blob/b16d377ce5a40edef5f7dff704b9143c4562f72d/setup.py#L112-L118

## User-facing changes

None

None


jtpio:
> Seems to be having troubles installing the `visual_test` `optional-dependencies` with `python -m pip install --upgrade ".[test,visual_test]"`

So https://github.com/voila-dashboards/voila/pull/1197/commits/55ab7edc1ab45bac08b1f4b634225831505e474d seems to be fixing it, by updating `visual_test` to `visual-test`.

jtpio:
CI is now passing and it is also looking good on Binder, marking as ready.

image

jtpio:
cc @martinRenou @trungleduc for review.

I think it should be fine to get this in a `0.3.x` release. Although we could also ship that in a `0.4.0` release with https://github.com/voila-dashboards/voila/pull/1046.

martinRenou:
I took the liberty to rebase and fix the conflict

setup.py
```diff
@@ -1,139 +1,2 @@
...
-        "https://unpkg.com/@jupyterlab/apputils@%s/style/materialcolors.css"
-        % jupyterlab_apputils_version,
-        "materialcolors.css",
-    ),
-    (
-        "https://unpkg.com/@jupyterlab/theme-light-extension@%s/style/variables.css"
-        % jupyterlab_theme_light_version,
-        "labvariables.css",
-    ),
-]
```
martinRenou, line 67: It looks like some data files and some CSS is not installed anymore? At least that's what I see from the code.

jtpio: Data files should still be handled by the new `tool.hatch.build.targets.wheel.shared-data` config.

Good catch for the CSS files, need to check how to keep them.

jtpio:
Converting back to a draft for now to check the handling of extra CSS files that used to happen in `setup.py`.

setup.py
jtpio: Looks like we can indeed keep the same logic with a custom hook in `hatch_build.py`.

jtpio:
The CSS files now seem to be correctly downloaded and available when looking at the releaser assets from https://github.com/voila-dashboards/voila/actions/runs/315

image

jtpio:
The size of the sdist seems to be bigger than before:

image

Which might be related to the gif that are included:

image

These gifs are not in the assets on the latest `main`:  https://github.com/voila-dashboards/voila/actions/runs/3150516185

So we should be able to filter them out.

jtpio:
Marking as ready for a second round of review.

The UI tests failure might not be related. The "Expected" snapshot currently looks broken and might need to be updated separately:

https://github.com/voila-dashboards/voila/blob/main/ui-tests/tests/voila.test.ts-snapshots/basics-classic-linux.png

https://user-images.githubusercontent.com/591645/193030780-d9a86b23-dd28-491b-ac57-4846969ad198.mp4

pyproject.toml
```diff
@@ -1,41 +1,159 @@
...
+[tool.hatch.build.hooks.jupyter-builder.build-kwargs]
 build_cmd = "build:prod"
-npm = ["jlpm"]
+npm = [
+    "jlpm",
+]
+
+# This is only so that we can pin to an ipyvolume branch, we should remove this when we can
+[tool.hatch.metadata]
+allow-direct-references = true
```
martinRenou, line 138: This can probably be removed now that the ipyvolume dep is in a separate file

```suggestion
```

jtpio: Good point.

jtpio:

jtpio:
The snapshot update for the `classic` template in https://github.com/voila-dashboards/voila/pull/1197/commits/5cd10eac6eac41f54d7d3463c266fd6b24e8742a looks legit. Not sure about `reveal` though.

martinRenou:
Looking at another build like for this PR, the UI-tests seems to pass. So I'm not sure I understand why this PR would need to change the reference screenshots?

jtpio:
Agree it's odd. One thing for sure is the current classic snapshot on `main` which does not look correct (https://github.com/voila-dashboards/voila/pull/1197#issuecomment-1262205610).

Wondering if this switch to hatch fixes some things to the way Voila is built and packaged for the UI tests.

martinRenou:
That could be it indeed.

Your PR looks good to me. Let's probably move forward and merge?

martinRenou (approved):

PR #1204, opened 2022-09-23
Error logging at runtime if we detect incompatible ipywidgets version

martinRenou:
## References

Log the following error when we detect that the user is not using a compatible ipywidgets version:

```
[Voila] ERROR | This Voila version is only compatible with ipywidgets version 7, though we detected that you are using ipywidgets version 8.0.2
```

This version is evaluated by checking if ipywidgets is being used in the Notebook (at least imported), and the error is shown in the terminal, but does not prevent the app to continue running.

### DIFF
--- a/voila/execute.py
+++ b/voila/execute.py
@@ -7,11 +7,28 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
+from nbformat import NotebookNode
 from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbclient.exceptions import CellExecutionError
 from nbclient import NotebookClient
 
 from traitlets import Bool, Unicode
+
+
+SUPPORTED_IPYWIDGETS_VERSION = "7"
+IPYWIDGETS_CELL_CHECK = NotebookNode(
+    id="ipywidgets-check",
+    cell_type="code",
+    metadata=NotebookNode(),
+    execution_count=None,
+    source=f"""
+import sys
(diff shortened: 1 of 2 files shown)


voila/execute.py
```diff
@@ -7,13 +7,30 @@
...
 
+from nbformat import NotebookNode
 from nbconvert.preprocessors import ClearOutputPreprocessor
 from nbclient.exceptions import CellExecutionError
 from nbclient import NotebookClient
 
 from traitlets import Bool, Unicode
 
 
+SUPPORTED_IPYWIDGETS_VERSION = "7"
```
martinRenou, line 18: This is a bit annoying to maintain, but I could not find a better approach.

trungleduc, line 18:
```suggestion
SUPPORTED_IPYWIDGETS_VERSION = 7
```

martinRenou:
This approach is actually a no-go. It only works with Python Notebooks 😅

martinRenou:
@trungleduc @SylvainCorlay 

We should check that the Notebook is a Python notebook prior to doing this. But I'm actually tempted to not show any error... Because this is a bit hacky.

trungleduc:
Maybe we can check it on the front-end side? Depending on the versions of `@jupyter-widgets/*`,  in our `WidgetManager` we can log a more explicit error or even raise an alert dialog.

martinRenou:
Right, and we have the hand on the manager itself, so we can definitely do that.

martinRenou:
Let's close this then :)

PR #1207, opened 2022-09-26
Try fixing check-release workflow

martinRenou:
## References

This workflow currently fails on master due to some cryptic TypeScript issue.


martinRenou:
Closing as we may not be able to fix it here, this is probably a bug in the releaser itself

PR #1211, opened 2022-09-28
Debug UI tests

martinRenou:
## References

UI tests are currently failing on the CI, but I'm not able to reproduce locally (with what looks like the same packages versions)


martinRenou:
Looks like the issue with the UI-tests not able to render ipympl and ipyvolume automagically fixed itself....

PR #1224, opened 2022-10-03
Update nbclient pinning

martinRenou:
## References

Update `nbclient` pinning


jtpio (approved):
Looking good and CI is all passing.

Thanks!

PR #1239, opened 2022-10-25
Add upper bound on jupyter_client dependency

martinRenou:
## References

Latest jupyter-client seems to break our tests. We will pin this dependency so we can finally make a release with ipywidgets 8 support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
-    "jupyter_client>=6.1.3,<=7.4.2",
+    "jupyter_client>=6.1.3,<=7.4.1",
     "jupyter_core>=4.11.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -34,7 +34,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
-    "jupyter_client>=6.1.3,<8",
+    "jupyter_client>=6.1.3,<=7.4.2",
```
trungleduc, line 37:
```suggestion
    "jupyter_client>=6.1.3,<=7.4.1",
```

martinRenou: 7.4.2 seems to be ok? Tests are passing

trungleduc: Nope. Some tests have been hanging for more than 1h now. Last time I checked, only `7.4.1` works

martinRenou: You're right! Thanks

PR #1249, opened 2022-10-31
Make use of JupyterLab mimetype renderers

martinRenou:
## Code changes

This PR changes the logic for rendering cell outputs, it switches from relying on `nbconvert` for handling different mimetypes to relying on `JupyterLab`.

The PR also has the benefit of removing the custom Widget Manager from Voila! It will load the `@jupyter-widgets/controls` package just as any other widget extension.

## User-facing changes

This adds support for many custom JupyterLab renderers: `jupyterlab-fasta`, `vega`, `jupyterlab-geojson`.

Screenshot from 2022-10-31 13-56-33

This might break some assumptions in the template, as it's overwriting the `data_priority` block in the `nbconvert` template.


martinRenou:
`HTML` and `SVG` do not seem to render properly yet

martinRenou:
Panel seems to work with this PR. Though the app is not updating when clicking on buttons (unlike in JupyterLab).

I suspect panel uses the `update_display` logic that Voila does not support (yet?).

Friendly ping @philippjfr, is that right that panel uses `IPython.display.update_display`? I don't see any reference to that in Panel.

Screenshot from 2022-10-31 16-24-08

martinRenou:
Bokeh works but suffers from the same issue as Panel, which I suppose makes sense?

Screenshot from 2022-10-31 16-42-53

philippjfr:
We do not use update_display, instead our Jupyter(Lab) extension lets us access a comm. As an alternative you can try wrapping the Panel app in `pn.ipywidget(...)`. That probably requires downgrading to ipywidgets<8 though.

philippjfr:
I suppose that doesn't really help you test out the render mime situation. I can take a look at what the problems on that side of things are when I get back from vacation.

maartenbreddels:
> We do not use update_display, instead our Jupyter(Lab) extension lets us access a comm

So if comm works, bokey should work? Could you give us some pointers to the code where this is done?

martinRenou:
I got bokeh to work in Voila using this PR:

[bokeh.webm](https://user-images.githubusercontent.com/21197331/201053476-10e12e0c-debe-47c2-b6d9-59ccda854fc5.webm)

The issue was that I was not serving Voila at the right port (bokeh expected me to serve on `localhost:8888`).

@philippjfr I wonder if this UX could be improved using only kernel comms instead of custom web sockets?

philippjfr:
Sorry I've been traveling for several weeks now. Just to summarize the current approaches for working with Bokeh/Panel interactively in notebooks:

1. **Embedding a server**: This is indeed fiddly due to setting up ports correctly and I would not recommend it as the default approach.
2. **Wrapping in an ipywidget**: The second approach we support is to wrap Bokeh model in the `jupyter_bokeh.widget.BokehModel` ipywidget. This uses the regular IPyWidget comms to shuttle messages across the websocket. Overall this approach is pretty clean but I've always found performance slightly less than optimal for some reason. There may also still be some lingering compatibility issues with ipywidgets 3.0 for this approach. The easiest way to invoke this is to wrap your Bokeh/Panel component with `pn.ipywidget(...)`
3. **Using a global comm**: This is the default approach we use for Panel and HoloViews and although conceptually it's very hacky it has been the most robust approach for us for several years now. The basic idea is that we create a Proxy object around the Jupyter comm manager which we patch onto a global object (`window.PyViz.comm_manager`) both in JupyterLab (via our JLab extension) and in classic notebook. While rendering Panel objects we then connect to a Jupyter Comm via this proxy object. The relevant code for this lives [here](https://github.com/holoviz/pyviz_comms/blob/master/src/renderer.ts) and [here](https://github.com/holoviz/panel/blob/master/panel/models/comm_manager.ts#L51).

maartenbreddels:
> 2\. Overall this approach is pretty clean but I've always found performance slightly less than optimal for some reason

I wonder if that's the server parsing all json, and encoding it again (that should improve due to @davidbrochart work on jupyter server I guess). My guess is that in situation 1 you just have a WebSocket, and skip a few layers in ipykernel/jupyter_server.

philippjfr:
>So, if I understand correctly, the @pyviz/jupyterlab_pyviz labextension should probably work with this PR,

Yup, that sounds right. Also happy to take a look.

jtpio:
> The PR also has the benefit of removing the custom Widget Manager from Voila! It will load the `@jupyter-widgets/controls` package just as any other widget extension.

Really nice! :+1:

jtpio:
> - [ ] HTML do not seem to render properly yet
> - [ ] SVG do not seem to render properly yet

This might need adding the corresponding lab extensions to the app.

---

Some additional observations after a quick check on the UI tests failures:

- The tree snapshots can be updated to include the new example notebook
- The `interactive` test might now be failing because the fix from https://github.com/voila-dashboards/voila/pull/846#issuecomment-1222047852 should now be done upstream: https://github.com/jupyter-widgets/ipywidgets/pull/3561

share/jupyter/voila/templates/classic/index.html.j2
```diff
@@ -32,6 +32,15 @@
 {%- block html_head_css -%}
   {{ resources.include_css("static/materialcolors.css") }}
   {{ resources.include_css("static/labvariables.css") }}
+
+  {#
+    This is a temporary solution, when Notebook 7 is out, the "classic"
+    template should inherit from the "lab" template and get the lab CSS
+    automatically
+  #}
+  {{ resources.include_css("../lab/static/index.css") }}
```
martinRenou, line 41: Although this fixes the missing CSS for some mimetype renderning, it's breaking the global look of the classic template (and opinionated templates like the AdminLTE one and the ipyvuetify one would be broken too) .

I'll see if I can be more specific about the CSS that is included in the page, we should only include the CSS for the mimetype renderers.

martinRenou:
>  HTML do not seem to render properly yet
> SVG do not seem to render properly yet
> > This might need adding the corresponding lab extensions to the app.

@jtpio which lab extensions are you referring to? I have not found them

packages/voila/src/plugins.ts
```diff
@@ -22,14 +22,26 @@ import { KernelConnection } from '@jupyterlab/services/lib/kernel/default';
...
   IJupyterWidgetRegistry,
   IWidgetRegistryData
 } from '@jupyter-widgets/base';
 
 import { VoilaApp } from './app';
 
-import { WidgetManager as VoilaWidgetManager } from './manager';
+import { Widget } from '@lumino/widgets';
+
+let resolveManager: (value: KernelWidgetManager) => void;
```
trungleduc, line 39: I'm not entirely sure but does the `PromiseDelegate` help simplify this logic?

voila/exporter.py
```diff
@@ -114,6 +114,13 @@ async def generate_from_notebook_node(
         )
         self.register_filter("highlight_code", highlight_code)
 
+        # Why do we need this? Should we make this configurable?
```
trungleduc, line 117:
```suggestion
```
Output can be rendered only if its flag `trusted` is `true`, and JupyterLab handles this flag from the frontend.

jtpio (commented):
Nice thanks @martinRenou changes look good!

Trying to test the PR on Binder gives the following error:

image

Not sure it's related to this change though.

voila/handler.py
```diff
@@ -198,12 +198,18 @@ def time_out():
...
+            if server_url[-1] == "/":
+                server_url = server_url[:-1]
+
             kernel_env = {**os.environ, **request_info}
             kernel_env[ENV_VARIABLE.VOILA_PREHEAT] = "False"
             kernel_env[ENV_VARIABLE.VOILA_BASE_URL] = self.base_url
-            kernel_env[ENV_VARIABLE.VOILA_SERVER_URL] = self.settings.get(
-                "server_url", "/"
-            )
+            kernel_env[ENV_VARIABLE.VOILA_SERVER_URL] = server_url
```
trungleduc, line 209: We also set these variables in the preheated kernels, should we update it to make it consistent with this case?

packages/voila/src/plugins.ts
jtpio: Also using something like this at the module level can make it more difficult to reuse and decouple plugins later if needed (looks similar to https://github.com/voila-dashboards/voici/pull/38).

voila/exporter.py
martinRenou: I meant that, maybe, it may be missing in the nbconvert execute logic. I wonder why we need to manually mark outputs as trusted ourselves.

voila/handler.py
trungleduc: I mean the value of the variables should be the same in both cases, we can keep your implementation in `voila/handler.py` and then update the corresponding variable in `VoilaKernelManager`

packages/voila/src/plugins.ts
martinRenou: Done!

trungleduc:
pre-commit.ci run

trungleduc (approved):

jtpio:
Nice, thanks!

Issue #827, opened 2021-02-12
Doc update needed in 'An example custom template' section


marckassay:
This section, ['An example custom template'](https://voila.readthedocs.io/en/stable/customize.html#an-example-custom-template), is referring to, I believe, previous generation of Jinja templates. Which is not aligned with what I'm seeing here in master: [share/jupyter/voila/templates](https://github.com/voila-dashboards/voila/tree/master/share/jupyter/voila/templates).

bmorris3:
👍🏻 This would be handy for those of us maintaining custom templates like jdaviz.

PR #126, opened 2019-04-25
updating documentation

choldgraf:
This updates a bit of the documentation that's contained in the readthedocs folder...it doesn't try to do too much other than adding a few new pieces of content. There's probably some information in there that isn't quite correct, so please feel free to give comments and/or edits!

cc @SylvainCorlay who I chatted about this with!

PR #1250, opened 2022-11-02
Install function for custom templates, template docs update

bmorris3:
## References

This PR addresses #827, as well as spacetelescope/jdaviz#1729.

## Code changes

This PR introduces a convenience function for installing custom templates, called `install_custom_template`. A user specifies a path to a custom template's directory structure and the name of the custom template. The custom template is then symlinked (preferred) or copied (fallback option) to the same directory where the voila `base` template is located. By default, this function does not overwrite existing templates. Also by default, `install_custom_template` will ignore any ``share/`` dirs in the current working directory (so if you run `install_custom_template` in your local `voila` repo, it won't install your custom template in the voila repo's ``share/`` dir.

Background: this PR was motivated by work on spacetelescope/jdaviz#1729. jdaviz, among other tools, contains a custom voilà template which gets installed with the package. We had to [carefully symlink or copy the template files from within our `setup.py`](https://github.com/spacetelescope/jdaviz/blob/main/setup.py#L123-L164) in order to make the jdaviz custom template available for voilà. In spacetelescope/jdaviz#1792 I developed the functions in this PR, and we realized they might belong upstream in voilà. If this PR is merged, it supersedes spacetelescope/jdaviz#1792.

## User-facing changes

I've edited docs page on custom templates, mostly by removing out-of-date content.

None.

### DIFF
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -106,101 +106,28 @@ A few example voila/nbconvert template projects are:
 * https://github.com/voila-dashboards/voila-vuetify
 
 
-Where are Voilà templates located?
-----------------------------------
+Install a custom template
+-------------------------
 
-All Voilà templates are stored as folders with particular configuration/template files inside.
-These folders can exist in the standard Jupyter configuration locations, in a folder called ``voila/templates``.
-For example:
+Suppose you created a custom template called ``mytemplate``, defined in a set of
+directories located on your machine at ``/path/to/share/``.
+You can install the custom template for use with Voilà like so:
 
-.. code-block:: bash
+.. code-block:: python
 
-   ~/.local/share/jupyter/voila/templates
-   ~/path/to/env/dev/share/jupyter/voila/templates
-   /usr/local/share/jupyter/voila/templates
-   /usr/share/jupyter/voila/templates
+   from voila.paths import install_custom_template
 
-Voilà will search these locations for a folder, one per template, where
-the folder name defines the template name.
+   custom_template_name = 'mytemplate'
+   share_path = '/path/to/share/'
 
-The Voilà template structure
-----------------------------
+   install_custom_template(share_path, custom_template_name)
 
-Within each template folder, you can provide your own nbconvert templates, static
-files, and HTML templates (for pages such as a 404 error). For example, here is
-the folder structure of the base Voilà template (called "default"):
-
-.. code-block:: bash
-
-    tree path/to/env/share/jupyter/voila/templates/default/
-    ├── nbconvert_templates
-    │   ├── base.tpl
-    │   └── voila.tpl
-    └── templates
-        ├── 404.html
-        ├── error.html
-        ├── page.html
-        └── tree.html
-
-**To customize the nbconvert template**, store it in a folder called ``templatename/nbconvert_templates/voila.tpl``.
-In the case of the default template, we also provide a ``base.tpl`` that our custom template uses as a base.
-The name ``voila.tpl`` is special - you cannot name your custom nbconvert something else.
-
-**To customize the HTML page templates**, store them in a folder called ``templatename/templates/<name>.html``.
-These are files that Voilà can serve as standalone HTML (for example, the ``tree.html`` template defines how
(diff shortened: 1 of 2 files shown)


bmorris3:
I wasn't sure if this can/should be tested, so I simply tested it locally on the jdaviz custom template by running: 
```python
from voila.paths import install_custom_template

install_custom_template('../jdaviz', 'jdaviz-default', overwrite=True)
```
I can confirm that the jdaviz template directories get symlinked within the correct directories, alongside voilà's "base" template dirs.

pllim:
cc @maartenbreddels @mariobuikhuizen

maartenbreddels:
Hi @bmorris3 ,

the PR here looks much larger than what was needed in https://github.com/spacetelescope/jdaviz/blob/main/setup.py#L123-L164 , could you describe the large discrepancy ?

Also, I am not sure why moving from setuptools to hatch doesn't allow for the current approach, I believe hatch has a good plugin system that allows custom code to be executed, like what we currently do in setup.py.

Also, the reason for the change in documentation is not clear to me, I see a large portion of documentation being replaced by documentation that doesn't seem to replace it.

Regards,

Maarten

bmorris3:
Hi @maartenbreddels,

> the PR here looks much larger than what was needed in https://github.com/spacetelescope/jdaviz/blob/main/setup.py#L123-L164 , could you describe the large discrepancy ?

If you compare [`jdaviz/setup.py:DevelopCmd.run`](https://github.com/spacetelescope/jdaviz/blob/main/setup.py#L123-L164) with the proposed function [`voila.paths.install_custom_template` in this PR](https://github.com/voila-dashboards/voila/pull/1250/files#diff-82b6d42a5f2677c71fdd29a303d9fb1a3641b086ffd45b4ca8e8bde1bda217dbR161), you'll see the contents are quite similar. The differences are:
* The jdaviz implementation is specific to the `jdaviz-default` template. The implementation in this voila PR is generic
* The voila PR has a detailed docstring, and comments
* The voila PR also adds a function called [`get_existing_template_dirs`](https://github.com/voila-dashboards/voila/pull/1250/files#diff-82b6d42a5f2677c71fdd29a303d9fb1a3641b086ffd45b4ca8e8bde1bda217dbR141) which is handy for identifying the available voila templates in a given environment.

> Also, I am not sure why moving from setuptools to hatch doesn't allow for the current approach, I believe hatch has a good plugin system that allows custom code to be executed, like what we currently do in setup.py.

This PR is somewhat independent from the setuptools/hatch conversation in https://github.com/spacetelescope/jdaviz/issues/1729. In the related PR  https://github.com/spacetelescope/jdaviz/pull/1792, I outlined that the only convenient way for a developer to "install" a custom voila template is to `pip install` the package containing the custom template. But the custom template will not be copied/symlinked when a developer uses editable install mode. Most active devs work with editable installs, so the current dev install would look like: 
```
cd jdaviz
pip install .  # "install" the voila template
pip uninstall jdaviz  # uninstall the static jdaviz installation
pip install -e .  # install the editable package
```
There's no convenient way to re-install further revisions to the custom template without repeating this process. (More details and discussion are over in that jdaviz PR.) 

> Also, the reason for the change in documentation is not clear to me, I see a large portion of documentation being replaced by documentation that doesn't seem to replace it.

As pointed out in #827, it seems that the template docs are out of date. The existing documentation was contributed several years ago in #126. If you look at [the file/directory structure explained in the voila docs](https://voila.readthedocs.io/en/stable/customize.html#the-voila-template-structure), you'll see it's out of sync with [the structure in the voila source code](https://github.com/voila-dashboards/voila/tree/main/share/jupyter/voila/templates). I've spent some time attempting the git archaeology to sort out how/when this happened, but I haven't found the PR. 

In this PR, I've simply [removed the section on custom templates](https://github.com/voila-dashboards/voila/pull/1250/files#diff-38827b3b5142629ea41c9b3e1765a98ece2d5e8f6d63b05412834037baca4cdeL160-L204), which is no longer accurate. I would contribute corrected docs in their place, but I don't think I am qualified to write out what voila expects.

Thanks!

PR #1268, opened 2022-11-30
Refactor styling

martinRenou:
## References

Make a separate style bundle to inject in the page.

If we were to use the style bundle produced by the Voila app we would pull the entire JupyterLab application CSS, breaking custom templates like classic/vuetify/reveal.

This will make https://github.com/voila-dashboards/voila/pull/1249 easier by providing a way to inject the CSS needed for JupyterLab mime renderers without breaking template's CSS.


martinRenou:
Some font-size changes are legit and references should be updated. This is because the CSS nbconvert pulls contains `font-size: 12px;`:

```
.jp-RenderedHTMLCommon table {
  border-collapse: collapse;
  border-spacing: 0;
  border: none;
  color: var(--jp-ui-font-color1);
  font-size: 12px;
  table-layout: fixed;
  margin-left: auto;
  margin-right: auto;
}
```

But the latest JupyterLab sets the font-size with proper lab variables: `font-size: var(--jp-ui-font-size1);` which is 13px.

martinRenou:
The reveal template visual changes are legit, we were not respecting the font sizes specified by nbconvert: https://github.com/jupyter/nbconvert/blob/main/share/templates/reveal/index.html.j2#L80 we now do

trungleduc (approved):

PR #1290, opened 2023-02-21
Extendable voila

martinRenou:
## References

This is the part of https://github.com/voila-dashboards/voila/pull/1187 (PR that was later reverted), that applies to Voila only. The rest of the PR (introduction to Voilite/Voici) is now in a separate repository: https://github.com/voila-dashboards/voici

### DIFF
--- a/packages/voila/src/index.ts
+++ b/packages/voila/src/index.ts
@@ -10,7 +10,5 @@
 export * from './app';
 export * from './manager';
 export * from './shell';
-
-import * as plugins from './plugins';
-
-export { plugins };
+export * from './output';
+export * from './plugins';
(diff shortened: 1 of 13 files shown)


trungleduc:
update galata references

martinRenou:
Not sure the references should be updated, the failure seems legit: all outputs are duplicated.

trungleduc:
> Not sure the references should be updated, the failure seems legit: all outputs are duplicated.

And the bot is also broken

martinRenou:
> And the bot is also broken

Indeed, it looks like a NodeJS version issue

martinRenou:
@trungleduc UI-tests are green now!

trungleduc:
Thanks @martinRenou

PR #1292, opened 2023-02-21
Update check-release workflow

martinRenou:


pyproject.toml
```diff
@@ -134,6 +134,7 @@ npm = [
 [tool.jupyter-releaser]
 skip = [
     "check-links",
+    "check-wheel-contents",
 ]
 
 [tool.jupyter-releaser.hooks]
```
jtpio, line 140: We'll also need to update the name of the `after-draft-release` hook.

Need to double check which one to use.

jtpio:
This error:

```
dist/voila-0.5.0a1-py3-none-any.whl: W002: Wheel contains duplicate files:
```

Can probably be fixed like in the JupyterLab extension cookiecutter: https://github.com/jupyterlab/extension-cookiecutter-ts/blob/8dca7b37e31bc149ebaae29cc3f942aebf19a88d/%7B%7Bcookiecutter.python_name%7D%7D/pyproject.toml#L85

martinRenou:
Thanks @jtpio !

jtpio:
We could probably also remove the skip section here:

https://github.com/voila-dashboards/voila/blob/eab947ec004a596864dcf93a6eb93867610356bc/pyproject.toml#L134-L137

And add a `check-links.yml` workflow instead: https://github.com/jupyterlab/maintainer-tools#check-links

martinRenou:
> We could probably also remove the skip section here:

Let's do that in a separate PR.

jtpio:
Looks like there is also an old reference to `jupyter-packaging` that could be removed (since Voila is now using hatch).

PR #1296, opened 2023-03-02
refactor: typo + rst styling

12rambau:
no need for fancy message, there was a typo in the file and the reference to Python objects was using 1 quotes when rst requires 2 of them.


trungleduc:
Thanks @12rambau !

PR #1301, opened 2023-03-14
Reveal template: Use proper routine for including JS asset

martinRenou:
This fixes support for the reveal template in Voici (and nbconvert?)


trungleduc:
Thanks @martinRenou

PR #1310, opened 2023-03-27
Clean up the check release workflow

jtpio:
## References

With Releaser v2 it should be possible to make a Voila release without installing the `voila` package, and without `jupyter-packaging` (Voila uses `hatch` now)

## Code changes

Clean up `check_release.yml`. This should avoid some potential issues that could arise during a real release but not with the check release.

## User-facing changes

None

None


trungleduc (approved):

trungleduc:
Thanks @jtpio

PR #1316, opened 2023-04-12
Remove unused `_get_query_string` function

jtpio:
## References

This `_get_query_string` util function appears to not be used anywhere else in the code base.

It is also prefixed with a `_` which suggests it was as a private utility function within the module before.

## Code changes

Remove unused code.

## User-facing changes

None

None

### DIFF
--- a/voila/utils.py
+++ b/voila/utils.py
@@ -104,12 +104,6 @@ def get_page_config(base_url, settings, log):
     return page_config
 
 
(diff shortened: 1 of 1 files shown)


trungleduc (approved):

maartenbreddels (approved):
Yeah, seems to be added in #846 but not used

martinRenou (approved):
Thanks!

PR #1320, opened 2023-04-21
Inject react-dom in the shared scope

martinRenou:
## References

Similar to https://github.com/voila-dashboards/voici/pull/67

This fixes support for ipyflex

### DIFF
--- a/packages/voila/package.json
+++ b/packages/voila/package.json
@@ -40,7 +40,8 @@
     "@lumino/signaling": "^1.7.2",
     "@lumino/virtualdom": "^1.11.2",
     "@lumino/widgets": "^1.26.2",
-    "react": "^17.0.1"
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1"
   },
   "devDependencies": {
     "@babel/core": "^7.2.2",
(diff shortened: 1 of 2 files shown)


trungleduc:
Thanks @martinRenou

trungleduc (approved):

maartenbreddels:
Why is this needed? What was the issue with ipyflex?
I don't think a widget library should depends on the environment providing react and react-dom. If a widget library wants to use a specific react version, https://github.com/jupyter-widgets/widget-ts-cookiecutter/issues/130#issuecomment-1490179846 should do that.
This PR might still be useful btw, but I've been fighting the battle recently. Maybe this information helps.

martinRenou:
That's due to the whole "shared package" logic in the webpack federated thingy https://github.com/trungleduc/ipyflex/blob/master/package.json#L99.

I think react needs to be specified as shared, we don't want all packages to provide their own react in the page, do we?
Also, does it actually work to have multiple react in the page? Wouldn't it break entirely?

maartenbreddels:
Multiple react version on the same page is no problem at all. And react is quite small: 
 * https://bundlephobia.com/package/react@18.2.0
 * https://bundlephobia.com/package/react-dom@18.2.0

The problem actually occurs when you get a wrong version of react and react-dom. Maybe a better strategy is `strictVersion =true` ( https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#deduplication-of-dependencies ) since I believe the default of that is false (the documentation doesn't say that, but otherwise I wouldn't see the issue).

PR #1330, opened 2023-06-20
Separate Tornado handlers into their own class

davidbrochart:
## References

This will make it easier to support Jupyverse (see #984).

## Code changes

This PR moves the Tornado-specific parts (the GET handlers) into their own class.

## User-facing changes

None.

None.

### DIFF
--- a/voila/tornado_handler.py
+++ b/voila/tornado_handler.py
@@ -1,0 +1,20 @@
+#############################################################################
+# Copyright (c) 2018, Voilà Contributors                                    #
+# Copyright (c) 2018, QuantStack                                            #
+#                                                                           #
+# Distributed under the terms of the BSD 3-Clause License.                  #
+#                                                                           #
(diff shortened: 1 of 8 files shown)


davidbrochart:
CI failures are unrelated, see #1331.

trungleduc:
Hey @davidbrochart, could you rebase your PR, the CI is fixed.

voila/tornado_handler.py
```diff
@@ -0,0 +1,20 @@
+#############################################################################
```
trungleduc, line 1: I think we could move these 2 files into the `tornado` directory, so we can remove the prefix in the file names.

davidbrochart: :+1:

trungleduc (approved):

trungleduc:
Thanks @davidbrochart !

PR #1333, opened 2023-06-22
Add nbconvert version check

trungleduc:
## References
Due to the non-backward compatible change in `nbconvert` 7.6.0 on markdown rendering, `VoilaMarkdownRenderer` class needs to be aware of the `nbconvert` version.

### DIFF
--- a/voila/exporter.py
+++ b/voila/exporter.py
@@ -8,6 +8,7 @@
 #############################################################################
 
 import mimetypes
+from typing import Optional
 
 import traitlets
 from traitlets.config import Config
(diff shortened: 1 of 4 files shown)


voila/exporter.py
```diff
@@ -20,7 +21,11 @@
 from nbconvert.exporters.html import HTMLExporter
 from nbconvert.exporters.templateexporter import TemplateExporter
 from nbconvert.filters.highlight import Highlight2HTML
-from nbconvert.filters.markdown_mistune import IPythonRenderer, MarkdownWithMath
+from nbconvert.filters.markdown_mistune import (
+    MISTUNE_V3,
```
martinRenou, line 25: How does it work? Doesn't this fail with an import error with prior versions?

SylvainCorlay:
Should this be backported?

trungleduc:
It looks like GitHub is running out of osx runner

trungleduc:
> Should this be backported?

Yes, only to `0.4.x` since `0.3.x` works with `nbconvert` <7

martinRenou (approved):
Thanks! Let's not wait for the OSX build

trungleduc:
@meeseeksdev please backport to 0.4.x

PR #969, opened 2021-09-17
Cleanup executor resources after execution

martinRenou:
Exploring solutions for https://github.com/voila-dashboards/voila/issues/896

Issue #896, opened 2021-05-26
Voila process is leaking threads


roman-kouzmenko:
Each time a notebook is accessed, voila is adding 3 threads that don't go away even as kernels are culled, eventually depleting all server resources.

Is anyone observing the same / has a fix?

roman-kouzmenko:
Debugging this a bit, it seems these threads are client channels that might not be disposed of properly, below is the output of lsof|grep [VOILA_PID]|grep TCP:

```
ZMQbg/Rea 146700 165411         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
ZMQbg/Rea 146700 165411         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
ZMQbg/IO/ 146700 165412         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
ZMQbg/IO/ 146700 165412         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
ZMQbg/Rea 146700 165413         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
ZMQbg/Rea 146700 165413         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
ZMQbg/IO/ 146700 165414         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
ZMQbg/IO/ 146700 165414         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
voila     146700 165415         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
voila     146700 165415         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
ZMQbg/Rea 146700 165417         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
ZMQbg/Rea 146700 165417         jovyan   50u     IPv4          754614991       0t0        TCP localhost:49283->localhost:49283 (ESTABLISHED)
ZMQbg/IO/ 146700 165418         jovyan    4u     IPv4          753391854       0t0        TCP *:8813 (LISTEN)
[...]

PR #1334, opened 2023-06-22
Do not clean up kernel resources after execution

martinRenou:
Revert https://github.com/voila-dashboards/voila/pull/969

This should never have been merged. The client still needs the kernel and channels running after execution !

When using a kernel provisionner, this would simply **kill the kernel**

### DIFF
--- a/voila/notebook_renderer.py
+++ b/voila/notebook_renderer.py
@@ -244,8 +244,6 @@ class NotebookRenderer(LoggingConfigurable):
         #  (it seems to be local to our block)
         nb.cells = result.cells
 
-        await self._cleanup_resources()
-
     async def _jinja_cell_generator(self, nb, kernel_id):
         """Generator that will execute a single notebook cell at a time"""
         nb, _ = ClearOutputPreprocessor().preprocess(
@@ -294,12 +292,6 @@ class NotebookRenderer(LoggingConfigurable):
                     ]
             finally:
                 yield output_cell
-
-        await self._cleanup_resources()
-
-    async def _cleanup_resources(self):
-        await ensure_async(self.executor.km.cleanup_resources())
-        await ensure_async(self.executor.kc.stop_channels())
 
     async def load_notebook(self, path):
         model = await ensure_async(self.contents_manager.get(path=path))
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):
Thanks for making me a co-author!

martinRenou:
@meeseeksdev please backport to 0.4.x

martinRenou:
@meeseeksdev please backport to 0.3.x

bsyouness:
Given that you're reverting a solution to this issue: https://github.com/voila-dashboards/voila/issues/896, don't we think it's gonna come back?

martinRenou:
It's possible that `self.executor.kc.stop_channels` may still be needed for #896 and not harmful in the case of using a kernel provisionner. Can you confirm on your setup @davidbrochart ?

davidbrochart:
I checked that `self.executor.kc.stop_channels` is not harmful for kernel provisioners. It just disconnects the kernel client, but the kernel is kept alive.

martinRenou:
It makes sense, cleaning up the kernel client in the Voila executor will help reducing the number of threads opened. We don't need the kernel client after execution because we won't execute more code

martinRenou:
I'm working on getting the 0.3.x branch to build, then will release a new patch for 0.3.x and 0.4.x. Will continue this PR after that.

bsyouness:
Also, I'm guessing y'all are planning on releasing a patch for 0.3.x and 0.4.x after the build is complete?

martinRenou:
With Voila there are basically two clients that communicate with the kernel:
- the Voila "executor", responsible for executing the Notebook (here in the Python code `self.executor.kc`). This happens when you see the Voila loading spinner.
- the Voila front-end, that allows rendered widgets to communicate with the kernel

So what we are cleaning here is the first one. After the notebook execution we don't need the executor anymore, so we close its communication channel.
This is why it helped solving https://github.com/voila-dashboards/voila/issues/896 in the first place.

> Also, I'm guessing y'all are planning on releasing a patch for 0.3.x and 0.4.x after the build is complete?

0.3.7 and 0.4.1 are out with this change already :)

PR #1337, opened 2023-06-22
0.3.x: Fix OpenSSL issue

martinRenou:


martinRenou:
ui tests failing just for ipyvolume, the rest passes just fine.

Issue #1304, opened 2023-03-16
Make `voila` compatible with `jupyterlab` 4 


trungleduc:
### Problem
Since `JupyterLab` 4.0  is just around the corner, I think for the next release (0.5.0), `voila` should be compatible with lab 4 and not lab 3. 

### Proposed Solution

- [x] Update to `jupyter-sever` 2 (https://github.com/voila-dashboards/voila/pull/1283)
- [x] Update the jupyterlab frontend components to use the 4.0.0 series, it can fix some remaining issues of https://github.com/voila-dashboards/voila/pull/846
- [ ] Use `type` when adding the preview to the `main` area. See https://github.com/jupyterlab/jupyterlab/pull/11614 for more information

jtpio:
Posting here one of the comments from #846 which could be addressed by updating to lab 4 packages:

> We should now be able to update to the `4.0.0-alpha.x` JupyterLab packages, which include the change from https://github.com/jupyterlab/jupyterlab/pull/11537 to allow for mock services.

jtpio:
FYI JupyterLab 4 final has been released: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.0

martinRenou:
We may be stuck waiting for ipywidgets to support lab 4 before being able to update Voila

trungleduc:
for the ipywidgets PR, a simple version bump is enough?

martinRenou:
Hopefully, some attempts have been made https://github.com/jupyter-widgets/ipywidgets/pull/3752/files

Galata needs to be updated there too

jtpio:
As part of the update of the Voila Preview extension for JupyterLab, it would also be useful to use the `type` property when adding a preview to the `main` area. So end users could decide to open the preview in a different area if they would like to (via the settings).

This functionality was added to JupyterLab in https://github.com/jupyterlab/jupyterlab/pull/11614.

PR #1343, opened 2023-06-28
JupyterLab 4 support

trungleduc:
## References

#1304 

## Code changes
- Update to JupyterLab 4 components
- Mock the service manager to remove unused services.
- Create a theme manager to load JupyterLab themes without depending on the setting plugin.
- Tree page is now a Lab remix app.

## User-facing changes
- New configurations in `VoilaConfiguration`:
  - `VoilaConfiguration.extension_whitelist`: The list of enabled JupyterLab extensions, if `None`, all extensions are loaded.
        This setting has higher priority than the `extension_blacklist`
  - `VoilaConfiguration.extension_blacklist`: The list of disabled JupyterLab extensions, if `None`, all extensions are loaded

- Theme and template config are renamed:
  - `voila-theme` -> `theme`
  - `voila-template` -> `template`

- Theme argument now is the theme name displayed in JupyterLab and not the name of the theme python package.


voila/tornado/treehandler.py
```diff
@@ -40,6 +40,16 @@ def allowed_content(content):
             contents["content"] = sorted(contents["content"], key=lambda i: i["name"])
             contents["content"] = filter(allowed_content, contents["content"])
 
+            theme_arg = (
+                self.get_argument("voila-theme", self.voila_configuration.theme)
```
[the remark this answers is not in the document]
trungleduc: yes, I think it's more nature to use `theme` and `template` instead of `voila-theme` and `voila-template`

packages/voila/src/app.ts
```diff
@@ -13,6 +13,7 @@ import { IRenderMime } from '@jupyterlab/rendermime';
 import { KernelWidgetManager } from '@jupyter-widgets/jupyterlab-manager';
 
 import { IShell, VoilaShell } from './shell';
+// import { ServiceManager } from '@jupyterlab/services';
```
martinRenou, line 16:
```suggestion
```

packages/voila/src/tools.ts
```diff
@@ -0,0 +1,118 @@
...
+//     // No white list, return non black listed only
+//     return pageConfigExtensionData.filter(
+//       (item) => mustHave.includes(item.name) || !blackList.includes(item.name)
+//     );
+//   }
+//   /// Have both black and white list, use only white list
+//   return pageConfigExtensionData.filter(
+//     (item) => mustHave.includes(item.name) || whiteList.includes(item.name)
+//   );
+// }
```
martinRenou, line 118:
```suggestion
```

voila/handler.py
```diff
@@ -111,8 +111,8 @@ async def get_generator(self, path=None):
                     env_name = f'HTTP_{header_name.upper().replace("-", "_")}'
                     request_info[env_name] = self.request.headers.get(header_name)
 
-        template_arg = self.get_argument("voila-template", None)
-        theme_arg = self.get_argument("voila-theme", None)
+        template_arg = self.get_argument("template", None)
+        theme_arg = self.get_argument("theme", None)
```
martinRenou, line 115: Should we still support `voila-theme` and `voila-template` to be backward compatible?

trungleduc: Let's profit from the freedom of major release 😄

trungleduc: and the theme argument is non backward-compatible anyway, users need to use the real theme name, not the package name.

martinRenou: Sounds good!

martinRenou:
Thank you !!

SylvainCorlay:
Yeah!

voila/configuration.py
```diff
@@ -140,3 +140,19 @@ class VoilaConfiguration(traitlets.config.Configurable):
         help="""Size of pre-heated kernel pool for each notebook. Zero or negative number means disabled.
         """,
     )
+
+    extension_whitelist = List(
```
jtpio, line 144: Maybe `0.5.0` could be a good time to switch to `allowlist` and `denylist`?

Since it was also suggested in https://github.com/voila-dashboards/voila/pull/725.

voila/utils.py
```diff
@@ -94,16 +103,86 @@ def get_page_config(base_url, settings, log):
...
         gpc(
             labextensions_path,
             logger=log,
         ),
     )
+    disabled_extensions = [
+        "@voila-dashboards/jupyterlab-preview",
+        "@jupyter/collaboration-extension",
+    ]
+    must_have_extensions = ["@jupyter-widgets/jupyterlab-manager"]
```
jtpio, line 118: Maybe `required_extensions` would also work as an alternative to `must_have_extensions`?

PR #1358, opened 2023-07-26
0.4.x: Rebuild against JupyterLab 4

martinRenou:


martinRenou:
update galata references

PR #1359, opened 2023-07-31
Fix import for jupyter_server 1

martinRenou:


martinRenou:
We are supposed to test jupyter server 1, let's investigate the CI

PR #1362, opened 2023-07-31
Update mathjax setting

trungleduc:
## References

Finally, I still use the CDN for the Mathjax files since the [jupyter_server_mathjax](https://github.com/jupyter-server/jupyter_server_mathjax) does not provide the required files for the current config.

Closes https://github.com/voila-dashboards/voila/issues/516

mathjax

## User-facing changes

New setting for Voila app:
- `mathjax_config`: Configuration for Mathjax , default to `TeX-AMS_CHTML-full,Safe`
- `mathjax_url`: URL to load Mathjax, default to `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js`

### DIFF
--- a/packages/voila/package.json
+++ b/packages/voila/package.json
@@ -22,7 +22,7 @@
     "@jupyterlab/mainmenu": "^4.0.0",
     "@jupyterlab/markdownviewer-extension": "^4.0.0",
     "@jupyterlab/markedparser-extension": "^4.0.0",
-    "@jupyterlab/mathjax-extension": "^4.0.0",
+    "@jupyterlab/mathjax2-extension": "^4.0.0-alpha.21",
     "@jupyterlab/nbformat": "^4.0.0",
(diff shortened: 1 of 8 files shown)


packages/voila/package.json
```diff
@@ -23,6 +23,7 @@
     "@jupyterlab/markdownviewer-extension": "^4.0.0",
     "@jupyterlab/markedparser-extension": "^4.0.0",
     "@jupyterlab/mathjax-extension": "^4.0.0",
+    "@jupyterlab/mathjax2-extension": "^4.0.0-alpha.21",
```
martinRenou, line 26: Do we need mathjax-extension AND mathjax2-extension?

trungleduc: Nope, only the latter is enough.

martinRenou (approved):

Issue #1366, opened 2023-08-01
Update keywords for white/black list


trungleduc:
Maybe `0.5.0` could be a good time to switch to `allowlist` and `denylist`?

Since it was also suggested in https://github.com/voila-dashboards/voila/pull/725.

_Originally posted by @jtpio in https://github.com/voila-dashboards/voila/pull/1343#discussion_r1264890308_

PR #1367, opened 2023-08-01
Update `blacklist` and `whitelist` keywords

trungleduc:
## References
Closes https://github.com/voila-dashboards/voila/issues/1366

## User-facing changes

- Keywords with `blacklist` are replaced by `denylist`
- Keywords with `whitelist` are replaced by `allowlist`
- A `DeprecationWarning` message is raised on using the old keywords.

N/A


voila/__init__.py
```diff
@@ -10,6 +10,9 @@
 from ._version import __version__  # noqa
 from .server_extension import _load_jupyter_server_extension  # noqa
 from .server_extension import load_jupyter_server_extension  # noqa
+import warnings
+
+warnings.filterwarnings("default", category=DeprecationWarning, module="traitlets")
```
martinRenou, line 15: Mmmh, Maybe we should not do that and fix the deprecation warnings instead?

voila/configuration.py
```diff
@@ -52,30 +54,62 @@ class VoilaConfiguration(traitlets.config.Configurable):
...
+    )
+
+    @validate("file_whitelist")
+    def _valid_file_whitelist(self, proposal):
+        warn(
+            "Deprecated, use VoilaConfiguration.file_allowlist instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return proposal["value"]
```
martinRenou, line 82: May I suggest another approach for this?

```python
@property
def file_whitelist(self):
   warn(
        "Deprecated, use VoilaConfiguration.file_allowlist instead.",
        DeprecationWarning,
        stacklevel=2,
    )
   return self.file_allowlist 


@file_whitelist.setter
def file_whitelist(self, value):
   warn(
        "Deprecated, use VoilaConfiguration.file_allowlist instead.",
        DeprecationWarning,
        stacklevel=2,
    )
   self.file_allowlist = value
```

That way we can get rid of the

```python
warnings.filterwarnings("default", category=DeprecationWarning, module="traitlets")
```

?

martinRenou: My suggestion applies if we consider going from a traitlet to a property is not backward incompatible

voila/__init__.py
trungleduc: Nope, the `DeprecationWarning` is hidden by default https://docs.python.org/3/library/exceptions.html#DeprecationWarning

voila/configuration.py
trungleduc: By doing so, users can not set the allow/deny list from the CLI or the `voila.json` file

martinRenou (approved):
Thanks!

PR #1369, opened 2023-08-02
Add extension config

trungleduc:


jtpio:
> but can not use the same layout system as in JupyterLab since the widgets do not show up.

What do you mean by this? Wondering if these areas could be hidden by default, and visible if a widget is added to them?



> Not sure if we should ship this in 0.5.0 or if this feature needs more discussion.

I guess even if a partially working implementation would be available in 0.5.0 it would already be good. And could be polished for 1.0.0. For example maybe the theme toggle top bar extension from https://github.com/jupyterlab-contrib/jupyterlab-topbar could also be reused in Voila?

trungleduc:
Hi all, this PR is ready for review again. Now I have a properly working shell with the top, main and bottom areas.

martinRenou (approved):
LGTM! Thanks! Letting a bit of time for @jtpio to have a look

martinRenou:
Did you have an example of an extension somewhere to was putting something in the top bar? 

If you agree, maybe this repo could go under the voila-dashboards org

trungleduc:
> Did you have an example of an extension somewhere to was putting something in the top bar?
> 
> If you agree, maybe this repo could go under the voila-dashboards org

I'm working on it, will ping you to create the repo later

jtpio:
Would it be possible to add a screenshot to have an idea of what the Voila dashboard looks like when there is a top bar component?

Also being able to reuse of the extensions from https://github.com/jupyterlab-contrib/jupyterlab-topbar would be great.

trungleduc:
> Would it be possible to add a screenshot to have an idea of what the Voila dashboard looks like when there is a top bar component?
> 

The top and bottom are empty divs of 10px min-width and hidden by default. Their positions are fixed and the scrollbar is now shown inside the main content area.
Here is a screenshot where the two are activated, I added the background for visibility.

vl



> Also being able to reuse one of the extensions from https://github.com/jupyterlab-contrib/jupyterlab-topbar would be great.

I doubt it since `Voila` does not provide lots of tokens.

martinRenou:
I think we need to be careful not to break custom templates by adding this kind of things. Custom templates are tested on the CI so for now with just the top and bottom that seems ok to me! Look great! Thanks

martinRenou:
Doing another beta with this now

Issue #1372, opened 2023-08-04
More advanced tree view


martinRenou:
As discussed with @trungleduc 

### Problem

It would be nice to implement a more advanced tree view making use of lab extensions. Even nicer would be to still support the static, template-based tree views custom templates are using nowadays (https://github.com/voila-dashboards/voila-material/blob/master/share/jupyter/voila/templates/material/tree.html for example).

### Proposed Solution

By default, make use of a FileBrowser widget from JupyterLab. With a custom Drive that just looks into the tree view the server provides (very simple content manager that does not leak file contents).

PR #1374, opened 2023-08-08
Create tree plugin

trungleduc:
## References
Closes https://github.com/voila-dashboards/voila/issues/1372

tree

## Code changes

- `pyzmq` is pinned to 25.1.0 on the Mac OS pipeline because `25.1.1` is not available as wheel yet on Mac OS

## User-facing changes
- The default tree page of Voila is now using the file browser widget of JupyterLab
- The jinja-based tree page is still supported, but users need to activate it with the `--classic-tree`  CLI option, the `VoilaConfiguration.classic_tree` config or `?classic-tree=True` in the query string.
- JupyterLab custom theme is supported with the new tree page, for the classic tree page, only the light and dark theme are supported.

- Users need to activate the classic tree to use existing tree templates.


jtpio:
@trungleduc you mentioned offline that interacting with the file browser items in this PR only requires 1 click, instead of the double click like in JupyterLab?

There is an issue to add this behavior in JupyterLab so that it could be configured, which would be especially useful for Notebook 7: https://github.com/jupyterlab/jupyterlab/issues/14071

Do you think this could be implemented upstream? This could go in JupyterLab 4.1 as a new feature.

trungleduc:
> @trungleduc you mentioned offline that interacting with the file browser items in this PR only requires 1 click, instead of the double click like in JupyterLab?
> 
> There is an issue to add this behavior in JupyterLab so that it could be configured, which would be especially useful for Notebook 7: [jupyterlab/jupyterlab#14071](https://github.com/jupyterlab/jupyterlab/issues/14071)
> 
> Do you think this could be implemented upstream? This could go in JupyterLab 4.1 as a new feature.

I was able to implement this feature since in the case of voila, I don't need other interactions like multiple selections or drag/drop. So I just need to overwrite the single click handler and disable all others.

Upstreaming this is not a technical problem but I think we need more UX studies before adding the single click behavior to Lab.

packages/voila/src/plugins/tree/listing.ts
```diff
@@ -0,0 +1,33 @@
...
+      const baseUrl = PageConfig.getBaseUrl();
+      const frontend = PageConfig.getOption('frontend');
+      const query = PageConfig.getOption('query');
+      const url = URLExt.join(baseUrl, frontend, 'render', path) + `?${query}`;
+      window.open(url, '_blank');
+    }
+  }
+
+  handleEvent(event: Event): void {
+    if (event.type === 'click') {
```
jtpio, line 29: Maybe we can make this part configurable in JupyterLab:

https://github.com/jupyterlab/jupyterlab/blob/7678e68391c5f372cb29749389f176c103716cea/packages/filebrowser/src/listing.ts#L690-L692

jtpio:
> Upstreaming this is not a technical problem but I think we need more UX studies before adding the single click behavior to Lab.

Maybe it would be fine if it is behind a user setting (opt-in)?

trungleduc:
> > Upstreaming this is not a technical problem but I think we need more UX studies before adding the single click behavior to Lab.
> 
> Maybe it would be fine if it is behind a user setting (opt-in)?

Indeed, I've just realized that we don't have any advanced interactions (drag/drop, ...) in the Notebook v6 browser. I'll open a PR after finishing this one.

I'll keep the implementation here to release Voila ASAP and will reuse the upstream version once it's released.

trungleduc:

voila/app.py
```diff
@@ -118,6 +120,12 @@ class Voila(Application):
             {"Voila": {"auto_token": True}},
             _(""),
         ),
+        "classic-tree": (
+            {
+                "VoilaConfiguration": {"classic_tree": True},
+            },
+            _("Use the jija2-based tree page instead of the new JupyterLab-based one"),
```
jtpio, line 127:
```suggestion
            _("Use the jinja2-based tree page instead of the new JupyterLab-based one"),
```

voila/treehandler.py
```diff
@@ -19,6 +19,22 @@ def initialize(self, **kwargs):
             ".ipynb",
         ]
 
+    def validate_theme(self, theme: str, classic_tree: bool) -> str:
+        """Check the compability between the requested theme and the tree page"""
```
jtpio, line 23:
```suggestion
        """Check the compatibility between the requested theme and the tree page"""
```

trungleduc:

trungleduc:
@jtpio could you re-take a look at this PR? I will release RC after this PR and update the documentation after that.

.github/workflows/main.yml
```diff
@@ -84,7 +84,8 @@ jobs:
 
       - name: Create the conda environment
         shell: bash -l {0}
-        run: mamba install -q python=${{ matrix.python_version }} pip jupyterlab_pygments==0.1.0 pytest-cov pytest-rerunfailures nodejs=18 yarn ipywidgets matplotlib xeus-cling "traitlets>=5.0.3,<6" ipykernel
+        # TODO unpin pyzmq
```
jtpio, line 87: What's the issue with `pyzmq`?

trungleduc: The `25.1.1` is not available yet on Mac OS

voila/configuration.py
```diff
@@ -35,6 +35,11 @@ class VoilaConfiguration(traitlets.config.Configurable):
     template = Unicode(
         "lab", config=True, allow_none=True, help=("template name to be used by voila.")
     )
+    classic_tree = Bool(
+        False,
+        config=True,
+        help=("Use the jija2-based tree page instead of the new JupyterLab-based one"),
```
jtpio, line 41:
```suggestion
        help=("Use the jinja2-based tree page instead of the new JupyterLab-based one"),
```

.github/workflows/main.yml
trungleduc: Done

jtpio: I meant next to the TODO, or in a new issue.

trungleduc: #1380

voila/app.py
```diff
@@ -671,6 +680,13 @@ def init_handlers(self) -> List:
                         "no_cache_paths": ["/"],
                     },
                 ),
+                (
+                    url_path_join(
+                        self.server_url, r"/voila/api/contents%s" % path_regex
```
jtpio, line 685: Curious: could exposing this endpoint allow someone access the content of other files on disk when they are not supposed to? For example when starting voila with `voila basics.ipynb` and trying to access another notebook via this endpoint?

trungleduc: Indeed this endpoint is leaking the file structure (not the content) of the current working directory in the case of starting Voila with one notebook, I'll fix it.

jtpio: ok thanks for confirming :+1:

trungleduc: Fixed by https://github.com/voila-dashboards/voila/pull/1374/commits/9c48666ca4da98b3a7aef0d50711b64fb0f27c79

jtpio (approved):
Looks good, thanks!

And CI all green.

PR #1385, opened 2023-08-24
Add voila style to tree page

trungleduc:
## References

- Only load voila style in the case of `voila` frontend.
- Add style to the tree page.


martinRenou (approved):
Thanks!

Issue #1392, opened 2023-09-13
Notebook widgets suddenly broke


idomic:
## Description

I had a notebook that was showing the widgets fine:
https://github.com/ploomber/doc/blob/main/examples/quickstart/notebook.ipynb

Deployed it via docker:
http://ploomberapp.io/silent-cake-6887

Now today both locally and on the docker I'm seeing some errors related to jupyter, I assumed it might be due to the release of ipywidgets. I did find out [this issue](https://github.com/voila-dashboards/voila/issues/1271), but locking the dependencies doesn't seem to work. Here's the errors:
```
Unsatisfied version 4.0.3 from _JUPYTERLAB.CORE_OUTPUT of shared singleton module @jupyterlab/translation (required ^4.0.5)
console.warn @ (index):36
p @ remoteEntry.a37e37c87d212fe85e13.js:1
f @ remoteEntry.a37e37c87d212fe85e13.js:1
(anonymous) @ remoteEntry.a37e37c87d212fe85e13.js:1
(anonymous) @ remoteEntry.a37e37c87d212fe85e13.js:1
452 @ remoteEntry.a37e37c87d212fe85e13.js:1
(anonymous) @ remoteEntry.a37e37c87d212fe85e13.js:1
_.f.consumes @ remoteEntry.a37e37c87d212fe85e13.js:1
(anonymous) @ remoteEntry.a37e37c87d212fe85e13.js:1
_.e @ remoteEntry.a37e37c87d212fe85e13.js:1
./extension @ remoteEntry.a37e37c87d212fe85e13.js:1
n @ remoteEntry.a37e37c87d212fe85e13.js:1
a @ 8113.voila.js:1
(anonymous) @ 5544.voila.js:1
(anonymous) @ 5544.voila.js:1
load (async)
75544 @ 5544.voila.js:1
p @ voila.js:1
Promise.then (async)
37559 @ voila.js:1
p @ voila.js:1
(anonymous) @ voila.js:1
(anonymous) @ voila.js:1
(index):36 Unsatisfied version 4.0.3 from _JUPYTERLAB.CORE_OUTPUT of shared singleton module @jupyterlab/settingregistry (required ^4.0.5)
console.warn @ (index):36
p @ remoteEntry.a37e37c87d212fe85e13.js:1
f @ remoteEntry.a37e37c87d212fe85e13.js:1
(anonymous) @ remoteEntry.a37e37c87d212fe85e13.js:1
[...]

PR #1394, opened 2023-09-14
Separate jupyter widgets manager extension

martinRenou:
## References

Attempt to provide a robust fix for  https://github.com/voila-dashboards/voila/issues/1392

## Code changes

Move the Voila jupyter-widgets manager into a separate federated extension. This should have the byproduct to properly share the `@jupyter-widgets/base` package in the federated extension logic, instead of having Voila ship its own copy of this package.


packages/widgets-manager/README.md
```diff
@@ -0,0 +1,7 @@
+# @voila-dashboards/jupyterlab-preview
+
+A JupyterLab preview extension for Voilà.
+
+## Prerequisites
+
+- JupyterLab 1.0+
```
martinRenou, line 7: mmh

martinRenou:
@trungleduc that works well!! Tested with jupyterlab_widgets 3.0.9 and 3.0.7.

packages/widgets-manager/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@voila-dashboards/widgets-manager",
+  "version": "0.1.0",
```
trungleduc, line 3: Should we sync the version with `voila` package?

packages/widgets-manager/schema/plugin.json
```diff
@@ -0,0 +1,16 @@
+{
```
trungleduc, line 1: This file is not needed

trungleduc:
Question about the releaser, we will bump both package versions on each release or they can be released separately?

martinRenou:
By default the releaser will bump both but I guess we can configure that in our bump script if we need.

martinRenou:
The test failure is legit. This new manager breaks jupyterlab because it says it provides the `IJupyterWidgetRegistry` token (so it overwrites the one from ipywidgets), but it fails saying we need a VoilaApp. This is all very weak.

martinRenou:
> This new manager breaks jupyterlab because it says it provides the IJupyterWidgetRegistry token (so it overwrites the one from ipywidgets), but it fails saying we need a VoilaApp

Because our extension **needs** a Voila app, I think we should separate the labextensions specific to Voila from the labextensions for other applications and lab-remixes in the path.

So Voila would look into `/share/jupyter/labextensions` for classic lab-extensions, and `/share/jupyter/voila/labextensions` for labextensions very specific to Voila. Opinions @jtpio @trungleduc ?

That way we're not breaking JupyterLab with our very specific extension.

trungleduc:
how can JupyterLab activate Voila's widget manager? We checked for the app instance before doing anything

martinRenou:
JupyterLab sees the Voila's widget manager provides the token `IJupyterWidgetRegistry` so it tries to activate it and that fails. And it does not fallback to the usual widget manager, the Voila one shadows it.

martinRenou:
It seems depending on `@voila-dashboards/voila` in a labextension for Voila breaks the labextension with `package @jupyterlab/filebrowser is not in the shared scope` so [we need to remove it](https://github.com/voila-dashboards/voila/pull/1394/commits/32a376d8eecfe919a106c236733e2c6bd713084b#diff-aa5a3717ba9e859a83c03b683425ce4063a8475b5501d5331603104bd2343950L47).

I see @trungleduc did the same in https://github.com/voila-dashboards/voila-topbar and need to work with `(app as any)` instead of importing the type of `VoilaApp`. 

Let's open an issue later for this.

martinRenou:
Ready for review

martinRenou:
As discussed with @trungleduc this will be too much of a change for a patch release. @trungleduc found a simpler fix and will come up with another PR.

Though I'd like that we consider these changes in a 0.6.x release, as scoping labextensions specific to Voila may make sense in some cases.

martinRenou:
Closing in favor of #1501

PR #1414, opened 2023-10-31
Add support for es modules, fixes #1413

### LINKED ISSUE #1413: `include_js` no longer compatible with nbconvert / support es modules
https://github.com/voila-dashboards/voila/blob/6418e2521f6901fa896448502f9a36743b8747ce/voila/utils.py#L267

This does not allow for es modules, which is now the second argument:
https://github.com/jupyter/nbconvert/pull/2063

timkpaine:


martinRenou (approved):
Thanks!

PR #1423, opened 2023-11-29
enhancement: Allow users to disable the fix_notebook call

### LINKED ISSUE #1422: Allow users to disable fix_notebook to check/resolve kernel validity
### Problem

[fix_notebook](https://github.com/voila-dashboards/voila/blob/main/voila/notebook_renderer.py#L324) can be very heavy handed for kernel resolution.  If a user wanted to customize the kernel matching logic, they would have to fork voila to achieve any customization in this regard.  While I agree it's good to have some guard rail in, users should be able to disable this logic in favor of handling it within their own kernel manager

### Proposed Solution

Add a toggle that can disable this logic

ClaytonAstrom:
## References
Resolves #1422

## Code changes
Simple boolean toggle that is checked to execute `fix_notebook`, `True` by default

## User-facing changes
N/A

N/A

### DIFF
--- a/docs/customize.md
+++ b/docs/customize.md
@@ -657,3 +657,11 @@ By default, Voilà's grace period for kernel startup time is 60 seconds. It can 
 ```sh
 voila --VoilaExecutor.startup_timeout=60
 ```
+
+## Have voila attempt to solve a best fit kernel spec
+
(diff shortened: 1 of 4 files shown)


martinRenou (approved):
That looks reasonable to me 👍🏽 Leaving some time for others to review

trungleduc (approved):

PR #1434, opened 2024-01-03
Update publish workflow for trusted publishers

jtpio:
## References

Start using trusted publishers for publishing Voila to PyPI.

This will help to not have to deal with 2FA with the bot account: https://blog.pypi.org/posts/2024-01-01-2fa-enforced/

## Code changes

- [x] Update `publish-release.yml` to remove the PyPI token and set up `id-token`
- [x] Setup `voila` on PyPI to use trusted publishers

image

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -15,6 +15,10 @@ on:
 jobs:
   publish_release:
     runs-on: ubuntu-latest
+    permissions:
+      # This is useful if you want to use PyPI trusted publisher
+      # and NPM provenance
+      id-token: write
     steps:
       - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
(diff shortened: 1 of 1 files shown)


jtpio:
We can then remove `voila-bot` from PyPI after this is merged.

jtpio:
> We can then remove `voila-bot` from PyPI after this is merged.

And the `PYPI_TOKEN` in the repo secrets.

Looks like `voila-bot` was only used for 2 projects on PyPI:

image

martinRenou:
Happy to remove that bot, thank you!

jtpio:
Just removed it from PyPI. I think we need to keep as `Admin` on the repos though.

martinRenou:
> I think we need to keep as Admin on the repos though.

Why?

jtpio:
Because it's still the bot creating the GitHub releases:

image

There is also still the need to the admin github token. But this might not be needed in a future version of the releaser: https://github.com/jupyter-server/jupyter_releaser/pull/545

martinRenou:
Ok 👍🏽

PR #1457, opened 2024-04-05
Disambiguate voilà frontend via html data attributes

timkpaine:
## References
https://github.com/finos/perspective/pull/2588
https://github.com/finos/perspective/pull/2352
https://github.com/finos/perspective/issues/1773

## Code changes
This PR fixes Perspective, and maybe some other plugins. But more importantly: "if one is going to distinguish behavior that is specific to Voila, it can't share classnames with Jupyter." One should be able to differentiate voilà-specific behavior, since there are **legimate**, often *subtle* UI/UX differences with Lab/Notebook.

## User-facing changes
This adds some very basic html data attributes to the top level html body, namely `data-in-voila`. I don't think it matters what this string actually is, merely that it is present and unique to voilà. This attribute will allow extensions to differentiate the voilà frontend in what is, imo, a minimally invasive way.

None!

### DIFF
--- a/share/jupyter/voila/templates/base/page.html
+++ b/share/jupyter/voila/templates/base/page.html
@@ -20,7 +20,7 @@
 
   </head>
 
-  <body data-base-url="{{ base_url }}voila/">
+  <body data-base-url="{{ base_url }}voila/" data-voila="voila">
 
     <noscript>
       <div id='noscript'>
--- a/share/jupyter/voila/templates/classic/index.html.j2
+++ b/share/jupyter/voila/templates/classic/index.html.j2
@@ -58,7 +58,7 @@
 {%- endblock html_head_js_mathjax -%}
 
 {% block body_header %}
-<body data-base-url="{{resources.base_url}}voila/">
+<body data-base-url="{{resources.base_url}}voila/" data-voila="voila">
   {{ spinner.html() }}
   {{ voila_setup_helper_functions() }}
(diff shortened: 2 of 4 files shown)


trungleduc:
Hi @timkpaine, IIUC you want to create plugins targeting only `Voila`? In this case, you can check for `app.name ==='Voila'` in the activation function.

timkpaine:
> Hi @timkpaine, IIUC you want to create plugins targeting only `Voila`? In this case, you can check for `app.name ==='Voila'` in the activation function.

Since the behavior is largely visible, it is often via `css` rules. So ideally a pure-css solution would be in place.

share/jupyter/voila/templates/base/page.html
```diff
@@ -20,7 +20,7 @@
 
   </head>
 
-  <body data-base-url="{{ base_url }}voila/">
+  <body data-base-url="{{ base_url }}voila/" data-in-voila>
```
trungleduc, line 23: I propose to use the same naming schema as in Notebook


https://github.com/jupyter/notebook/blob/3b8e6736075279f80cd1e590c71a872dc5ad2c11/notebook/templates/notebooks.html#L15

```suggestion
  <body data-base-url="{{ base_url }}voila/" data-voila="voila">
```

trungleduc (approved):

trungleduc:
Thanks @timkpaine

PR #1470, opened 2024-05-14
fix documentation for voila.json location

### LINKED ISSUE #1468: voila.json file is not taken into account
The doc say:
The easiest way to change these settings is to provide a file named voila.json in the same folder containing the notebooks. 

But I notice that it is especially necessary to launch "voila" from the directory containing the voila.json file.

If you have the directory $HOME/app which contains index.ipynb and voila.json:

```
cd $HOME/app
voila index.ipynb   # works

```
```
cd /tmp
voila $HOME/app/index/ipynb  # doesn't work (the config file is not taken into account)

```

therefore the documentation is false. 
But, it would be better to correct the code so that the documentation becomes valid.
Even better, an option to indicate where the configuration file is located at launch.

Regards

djangoliv:
fix #1468

## References

#1468

### DIFF
--- a/docs/customize.md
+++ b/docs/customize.md
@@ -491,7 +491,7 @@ voila --preheat_kernel=True --VoilaKernelManager.default_env_variables='{"FOO": 
 
 will set the variable "FOO" in all preheated kernels.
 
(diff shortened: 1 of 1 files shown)


trungleduc (approved):

trungleduc:
Thanks!

PR #1471, opened 2024-05-16
Get mybinder.org working again

manics:
## References

N/A

## Code changes

The mybinder configuration is currently broken, this modifies the environment and build script so it's working again. The JupyterLab `Render with Voila` also works:
https://mybinder.org/v2/gh/manics/voila/binder

## User-facing changes
The mybinder.org link previously failed to build, now it works

None


martinRenou (approved):
Thanks!

PR #1479, opened 2024-07-15
ipywidgets 7 support

martinRenou:
This is the approach I'm taking:

If ipywidgets 7 or 8 is spotted as federated extension, we don't enable it but provide our own versions of `@jupyter-widgets/base` and `@jupyter-widgets/controls` in the page using our manager for the given version.

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -8,8 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
-        node-version: [18.x]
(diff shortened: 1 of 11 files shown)


martinRenou:
Testing with ipyleaflet works like a charm, testing bqplot fails with:

```
TypeError: _jupyter_widgets_base__WEBPACK_IMPORTED_MODULE_0__.JupyterPhosphorPanelWidget is not a constructor
```

martinRenou:
I'm closing this, instead of this approach we should do something similar to https://github.com/voila-dashboards/voila/pull/1482 and use the lazy loading approach.

PR #1482, opened 2024-07-16
0.4.x: Lazy loading of widgets packages, for ipywidgets 7 and 8 support

martinRenou:
Lazy load the widgets extensions upon request: wait for the first widget creation (coming from the kernel) before loading the controls/base extension.

Remaining TODO:
- [x] Lazy load CSS so that either the CSS for 7 or 8 is included in the page
- [x] ~~Lazy load output widget for 7~~ There is nothing we can do here, the https://github.com/jupyter-widgets/ipywidgets/blob/main/packages/output/src/output.ts#L3 version was not bumped between 7 and 8 so we need to stick to one available version
- [x] Add UI-tests for both 7 and 8

Screenshot from 2024-07-17 09-14-23
Screenshot from 2024-07-17 09-27-20
Screenshot from 2024-07-17 09-50-24

### DIFF
--- a/packages/voila/src/manager.ts
+++ b/packages/voila/src/manager.ts
@@ -7,7 +7,7 @@
  * The full license is in the file LICENSE, distributed with this software. *
  ****************************************************************************/
 
-import $ from "jquery";
+import $ from 'jquery';
 import 'jquery-ui/ui/widgets/slider';
 
 import {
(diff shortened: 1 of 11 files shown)


packages/voila/src/manager.ts
```diff
@@ -105,34 +101,23 @@ export class WidgetManager extends JupyterLabManager {
...
-        //
-        // This fixes issues with widget models that are explicitly "closed"
-        // but are still referred to in a previous cell output.
-        // Without the try-catch statement, this error interrupts the loop and
-        // prevents the rendering of further cells.
-        //
-        // This workaround may not be necessary anymore with templates that make use
-        // of progressive rendering.
-      }
+
```
martinRenou, line 104: This was just for removing a silent error, I should revert this change

packages/voila/src/manager.ts
```diff
@@ -181,20 +185,54 @@ export class WidgetManager extends JupyterLabManager {
...
     this.register({
       name: '@jupyter-widgets/base',
-      version: base.JUPYTER_WIDGETS_VERSION,
-      exports: base as any
+      version: '2.0.0',
+      exports: async () => {
+        const baseWidgets = require('@jupyter-widgets/base') as any;
+        windowDefine('@jupyter-widgets/base', baseWidgets);
+        return baseWidgets;
+      }
```
trungleduc, line 198: No `await` anywhere, can we drop the `async` keyword ?

trungleduc (approved):

PR #1487, opened 2024-08-19
Backward compat for ipywidgets 7

martinRenou:
Revive https://github.com/voila-dashboards/voila/pull/1479


packages/voila/webpack.config.js
```diff
@@ -95,9 +109,7 @@ module.exports = [
           name: ['_JUPYTERLAB', 'CORE_LIBRARY_FEDERATION']
         },
         name: 'CORE_FEDERATION',
-        shared: {
```
martinRenou, line 98: I'm having issues with the sharing logic which I'm not sure how to tackle.

We would need to have `@jupyter-widgets/base` be a shared package, but the version being loaded in the page would depend on the current ipywidgets version (model version requested by the kernel).
In the current state of the PR, `@jupyter-widgets/base@6.0.6` is loaded.

Would love to chat with you @jtpio on this. 

I will this PR as draft for now

trungleduc: what is the use case we want to support? ipywidgets 7 and voila are installed in the same kernel env or ipywidgets 8 + voila in the base env and kernel env is using ipywidgets 7?

martinRenou: Voila in the base env and either ipywidgets 7 or 8 in the kernel env basically. Resulting in having Voila support both ipywidgets 7 or 8 if the kernel env is the voila base env.

For Voila 0.4.x that was easy using lazy loading of packages, and for some reason the package sharing logic did not come in my way when doing https://github.com/voila-dashboards/voila/pull/1482. Maybe because the build setup did not use webpack federated packages? 

Now in this PR, for some reason the webpack sharing package logic makes it that `@jupyter-widgets/base@6.0.6` is in the page, which incompatible with `@jupyter-widgets/controls` so you can't create a `VBox` widget for example. And we can't decide at build time which version of `@jupyter-widgets/base` should be loaded and shared, I'd like to do that at runtime. Maybe overwriting `__webpack_share_scopes__` at runtime in some way during the lazy load of the widget front-end?

trungleduc: I tested our PR with theses configurations:
- Base = voila + ipywidgets7 + ipyleaflet 0.16.0, Kernel = Base -> work
- Base = voila + ipywidgets 8 + ipyleaflet 0.19.2, Kernel = Base  -> work
- Base  = voila + ipywidgets 8, Kernel = ipywidgets 7 +  ipyleaflet 0.16.0  -> ipywidgets works, ipyleaflet does not, which is understandable. 

vl7

- Base = voila + ipywidgets 7,  Kernel =  ipywidgets 8 -> skipped, who would use is non-sense combination 😄

martinRenou: Thanks for your testing! Yeah hopefully most cases would work, the one that does not work is:

Base = voila + ipywidgets7, Kernel = Base -> VBox does not work.

This is because of the issue I'm trying to explain up there in this thread with mismatching of versions between @jupyter-widgets/base and controls. We have base for 8 and controls for 7 in the page, thankfully most widgets do work in this case, but I'd really prefer to have base for 7 in the page in that case. And that's what I'm failing to figure out.

martinRenou: Indeed, I opened this PR in ipywidgets yesterday for this: https://github.com/jupyter-widgets/ipywidgets/pull/3933

> Bundling base7 and controls7 with Webpack before building voila would make this case work

Bundling base7 and controls7 would break the case for ipywidgets 8, or am I not understand what you're suggesting?

trungleduc: I mean bundling base7 and controls7 into a standalone file then you import base7 and controls7 from this file in your ipywidgets7 plugins to register the classes

martinRenou: Right! Unfortunately I believe that would work for core ipywidgets but custom widget that have not been updated to work with ipywidgets 8 will not work in that case, they would have no way to get access to that bundled base7

trungleduc: > they would have no way to get access to that bundled base7

Why?

martinRenou: Because what will be shared (in the webpack federation sense) would still be `@jupyter-widgets/base` for 8, the bundled one will not be shared.

I guess it would be possible to use the [`requiredVersion` feature of webpack module federation](https://webpack.js.org/plugins/module-federation-plugin/#requiredversion) and share two versions of base. But AFAIK nobody has been using this feature in custom widgets, it would require updating them all. I'm not even sure if `jlpm` support this entry in the `jupyterlab` entry for `package.json`.

trungleduc: > Because what will be shared (in the webpack federation sense) would still be @jupyter-widgets/base for 8, the bundled one will not be shared.

But the custom widgets do not know about the shared package of webpack. They get their classes from the registry of the widget manager, where we register both versions of base and control classes. And when the view or model classes are loaded, depending on how it was registered, it will find the imported classes from the shared packages (ipywidgets 8 cases) or its bundle (ipywidgets 7 cases)

trungleduc: If the base env has the correct version of the lab extension package, `ipyleaflet` 0.16.0 can work with the current approach

leaflet

martinRenou: > But the custom widgets do not know about the shared package of webpack

Yes they do 😕 

You can try to remove `@jupyter-widgets/base` from the shared packages by changing the code to this line 72 of the webpack config in my PR:

```
const shared = {};
for (const dependency of Object.keys(data.dependencies)) {
  // TODO Why can we not share those?
  if (
    ['@jupyter-widgets/base', '@jupyter-widgets/controls', '@jupyter-widgets/base7', '@jupyter-widgets/controls7'].includes(
      dependency
    )
  ) {
    continue;
  }

  shared[dependency] = data.dependencies[dependency];
}
```

The result of this is that `@jupyter-widgets/base` is not shared anymore, and custom widgets like bqplot or ipyleaflet will stop working, complaining at runtime that they can't find `@jupyter-widgets/base`.

In the end, I believe making use of the `requiredVersion` webpack federation feature is the way to go. But I also believe we can't do it in a short timeline, it's more like a long term solution.

Having base8 in the page with the patches from https://github.com/jupyter-widgets/ipywidgets/pull/3933 is probably enough to have widgets working, but it makes me uncomfortable because I'm sure there will be more issues.

martinRenou: > ipyleaflet 0.16.0 can work with the current approach

I think it's only out of luck. As soon as the custom widget makes use of an API that was broken between base7 and base8 (e.g. `JupyterPhosphorPanelWidget` like VBox uses), it will break. We can fix the backward incompatibility for `JupyterPhosphorPanelWidget` by merging my PR in ipywidgets, but I fear there will be more issues that we miss and can't make backward compatible.

trungleduc: ok, I see. I think we should do the minimum for the ipywidgets backward compatible in Voila. The long-term solution is to adopt the voici/jupyterlab execution model. So that we only need to fix the backward issue once.

martinRenou: As mentioned by @maartenbreddels in https://github.com/jupyter-widgets/ipywidgets/pull/3932

> In solara (because we run in the same process), we know beforehand at the server what the version of ipywidgets is, which allows for a different strategy.
We also build the final bundle twice using the webpack alias trick:
https://github.com/widgetti/solara/blob/20b2e8012e7683eb38ff62a74a371eb1a052bf41/packages/solara-vuetify-app/webpack.config.js#L95

I wonder if the two webpack bundles approach could suit us, one bundle with `@jupyter-widgets/base` for 7 being shared, another bundle with 8. And we load the bundle upon widget request depending on the widget version.

So basically very similar to your suggestion @trungleduc

> I mean bundling base7 and controls7 into a standalone file then you import base7 and controls7 from this file in your ipywidgets7 plugins to register the classes

While making sure that the shared package is different for each bundle

trungleduc: yeah I think it could work if we check for the widget version in the backend and write it to the page template

martinRenou: Ah no I don't think checking in the back-end is a viable solution, ipywidgets may not be installed and if it is we don't know if it's the same version as in the kernel.

We'd still need to rely on the lazy loading logic from the widgets register, and lazy load one bundle or the other depending on what the kernel requests.

trungleduc: theoretically, we can peek at the kernel before generating the template, it can be the last resort if the lazy loading logic does not work.

martinRenou:
Closing in favor of #1501

PR #1491, opened 2024-09-23
0.4.x: Fix markdown rendering logic

martinRenou:
Using the following code:

```python
from ipywidgets import Output

from IPython.display import display, Markdown

out = Output()

with out:
    display(Markdown("[google](https://google.com)"))

out
```

Results in not rendering the link properly, instead we see the raw code inlined `"[google](https://google.com)"`

This PR fixes that issue.


martinRenou:
The PR correctly fixes the example above, but it seems to break mathematical expressions:

Screenshot from 2024-10-03 16-45-24


I also see this issue on Voila 0.5.x so this looks like a regression from updating to this new logic of handling mathjax.

martinRenou:
This is working now 🚀

jtpio (approved):
Thanks!

Issue #1494, opened 2024-10-02
custom labextension path?


dfguerrerom:
I was navigating through the documentation and the code but I couldn't find an answer to the following.

My group has a system where we have a `jupyter server` running from where users can launch some voila applications, for each app, we set the kernel that the app has to use and it is then launched.
The problem surges when there's a tool that needs a new labextension which is not installed in the system library, causing the `no version of module xxx is registered`.

By running `jupyter --path` I know that the registered extensions are located at:

```
    /home/username/.local/share/jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
```

I know I can `export JUPYTER_PATH=` var to add the folder of the venv that I'm using, but it won't solve my issue because the server is already running, I cannot change that...

Is there a way to set this path in a different way? or is there a way to use the `cdn` ?

trungleduc:
It's not possible to set the lab extension paths dynamically. 
What we can do is to relax this [`fullLabextensionsUrl`](https://github.com/voila-dashboards/voila/blob/0ffd12ae476d843fa7f4f66c1f19e8cef41f1eeb/voila/utils.py#L100) to make it configurable. Then you can host your extension assets independently of `voila` and update them without touching voila server

dfguerrerom:
I need this feature, and I could start a PR.
I thought that `fullLabextensionsUrl` would be the only entry point used afterwards to handle the extensions but not, to get the metadata from the extensions there's another call to the filesystem https://github.com/voila-dashboards/voila/blob/0ffd12ae476d843fa7f4f66c1f19e8cef41f1eeb/voila/utils.py#L114 ... 
To propose a generic solution I was thinking on adapting or adding  `gcp` method to work directly with the server instead of the filesystem, in which scenario I would have to add some directory listing methods to the server and adapt the logic.
What do you think about it?

PR #1495, opened 2024-10-06
Custom page config hook

dfguerrerom:
## References

https://github.com/voila-dashboards/voila/issues/1494

## Code changes

~~I have added a new query param to set a custom `labextensions_url`, this url has to be used when is provided by the user.~~
~~I have created a draft of [my own server](https://github.com/openforis/sepal/blob/7e06d9dd25c9b05e03bea3b473660ad9a52f3cb6/modules/sandbox/script/labextension-server.py) to serve the custom lab extensions and it works as expected.~~

~~I'm not sure what is the best way to deal with the `labextensions_path` and how to use the ones that are served by the tornado server, so that' why I came doing a request to the server.~~

Any suggestion?

----------------------------------------------------------
- New configs in `VoilaConfiguration`:
```python
    prelaunch_hook = Callable(
        default_value=None,
        allow_none=True,
        config=True,
        help="""A function that is called prior to the launch of a new kernel instance
            when a user visits the voila webpage. Used for custom user authorization
            or any other necessary pre-launch functions.

            Should be of the form:

            def hook(req: tornado.web.RequestHandler,
                    notebook: nbformat.NotebookNode,
                    cwd: str)

            Although most customizations can leverage templates, if you need access
            to the request object (e.g. to inspect cookies for authentication),
            or to modify the notebook itself (e.g. to inject some custom structure,
            although much of this can be done by interacting with the kernel
            in javascript) the prelaunch hook lets you do that.
            """,
    )

    page_config_hook = Callable(
        default_value=None,
        allow_none=True,
        config=True,
        help="""A function that is called to modify the page config for a given notebook.
            Should be of the form:

            def page_config_hook(
                current_page_config: Dict[str, Any],
                base_url: str,
                settings: Dict[str, Any],
                log: Logger,
                voila_configuration: VoilaConfiguration,
                notebook_path: str
            ) -> Dict[str, Any]:

            The hook receives the default page_config dictionary and all its parameters, it should
            return a dictionary that will be passed to the template as the `page_config` variable
            and the NotebookRenderer. This can be used to pass custom configuration.
            """,
    )

```

- Reassign `Voila.prelaunch_hook` to `VoilaConfiguration.prelaunch_hook`, this will make `prelaunch_hook` work with the server extension.
- Add deprecated warning for  `Voila.prelaunch_hook` 

## User-facing changes

Users will have the opportunity to set their custom config_page settings

None


trungleduc:
I think a more generic solution is to allow users to modify the page_config object by a hook function. We have this mechanism for the prelaunch hook (https://voila.readthedocs.io/en/stable/customize.html#accessing-the-tornado-request-prelaunch-hook)

dfguerrerom:
@trungleduc what do you think about this implementation?

dfguerrerom:
while working on this, I have a question, why not use the kernel path to set the labextensions path? 
If we can select which kernel to use within the kernelspec of the notebook, why not use also that path to get the labextensions?
With this hook I can handle the page config, but I don't know what notebook is actually requesting the page config... I would still need the query param or pass the notebook info to the hook?

trungleduc:
I'm thinking about a post-processing function that receives the current page_config dict and then returns a customized one. You can add more arguments to this hook function if needed.

dfguerrerom:
> I'm thinking about a post-processing function that receives the current page_config dict and then returns a customized one. You can add more arguments to this hook function if needed.

I don't get the idea, what would be the difference?
Another question, did you also think about set hooks when the `server_extension` is launched?

What I'm doing to set the hooks is something like:

```python
# voila.py 

from voila.app import Voila

def get_page_config_hook(....)

Voila.get_page_config_hook = get_page_config_hook

```

I'm mutating the Voila class and setting the hook, is that the way?

trungleduc:
> > I'm thinking about a post-processing function that receives the current page_config dict and then returns a customized one. You can add more arguments to this hook function if needed.
> 
> I don't get the idea, what would be the difference? Another question, did you also think about set hooks when the `server_extension` is launched?
> 

It will be easier for someone who just needs a small change in the `page_config`. They don't need to redo all the lab extension gathering logic, etc...

> What I'm doing to set the hooks is something like:
> 
> ```python
> # voila.py 
> 
> from voila.app import Voila
> 
> def get_page_config_hook(....)
> 
> Voila.get_page_config_hook = get_page_config_hook
> ```
> 
> I'm mutating the Voila class and setting the hook, is that the way?

https://voila.readthedocs.io/en/stable/customize.html#adding-the-hook-function-to-voila

dfguerrerom:
I've created this post-processing function, does it look better to you? 
Regarding the documentation for implementing the hook, I see a reference to an `object` "c" but I'm not sure where it comes from, is it the configuration?
In any case, if I go for the first approach which requires to modify the `voila.py`, the `server_extension` doesn't use the `Voila` object which is the one that holds the hook, am I right?

voila/app.py
```diff
@@ -331,24 +331,17 @@ def hook(req: tornado.web.RequestHandler,
...
-                settings: Dict[str, Any],
-                log: Logger,
-                voila_configuration: VoilaConfiguration,
-                **kwargs
-            ) -> Dict
-
-            The hook should return a dictionary that will be passed to the template
-            as the `page_config` variable and the NotebookRenderer. This can be used to pass custom
-            configuration.
+            def hook_fn(page_config, **kwargs) -> Dict
```
trungleduc, line 340: Please precise the arguments of the function

voila/app.py
```diff
@@ -331,24 +331,17 @@ def hook(req: tornado.web.RequestHandler,
     )
 
     get_page_config_hook = Callable(
```
trungleduc, line 333: `page_config_hook` may be a better name

trungleduc:
> Regarding the documentation for implementing the hook, I see a reference to an `object` "c" but I'm not sure where it comes from, is it the configuration?

the `c` object will be injected by `traitlets` https://traitlets.readthedocs.io/en/stable/config.html#python-configuration-files

trungleduc:
> In any case, if I go for the first approach which requires to modify the voila.py, the server_extension doesn't use the Voila object which is the one that holds the hook, am I right?

it's correct

trungleduc:
Please rebase this branch to get the CI fix from https://github.com/voila-dashboards/voila/pull/1497

dfguerrerom (commented):
done!

trungleduc:
> > > In any case, if I go for the first approach which requires to modify the voila.py, the server_extension doesn't use the Voila object which is the one that holds the hook, am I right?
> > 
> > 
> > it's correct
> 
> is there a reason for that? one would expect the hooks to work no matter which server is used, dont you think?

could you elaborate more on your setup?

dfguerrerom:
> could you elaborate more on your setup?

We have a platform that allows users to run custom AWS instances, each of this instances start a jupyter server that we use to launch jupyter and voila apps... 
The apps are located in a shared folder that we mount into users instances alongside with their kernels... if they want to start an app, they use the jupyter server that is already running, and we just pass the path of the notebook... running another server just for voila dashboards would require more memory...

trungleduc:
In this case, you should use the configuration file (`voila.json` or `voila.py`), This file is picked up by both standalone app and the server extension

dfguerrerom:
> In this case, you should use the configuration file (`voila.json` or `voila.py`), This file is picked up by both standalone app and the server extension

but the `Voila` class is never instantiated in `server_extension.py` so it doesn't know about the hooks and they are not passed to the handlers, and c is not automatically imported in that context...

trungleduc:
> , and c is not automatically imported in that context...

yes, It's https://github.com/voila-dashboards/voila/blob/f1722f46811a244b138f08bf6af02647890a6afd/voila/server_extension.py#L49

voila/app.py
```diff
@@ -349,6 +349,29 @@ def hook(req: tornado.web.RequestHandler,
...
+            """A function that is called to modify the page config for a given notebook.
+            Should be of the form:
+
+            def page_config_hook(
+                base_url: str,
+                settings: Dict[str, Any],
+                log: Logger,
+                voila_configuration: VoilaConfiguration,
+                notebook_path: str
+            ) -> Dict[str, Any]:
```
trungleduc, line 366:
```suggestion
            def page_config_hook(
                current_page_config: Dict[str, Any],
                base_url: str,
                settings: Dict[str, Any],
                log: Logger,
                voila_configuration: VoilaConfiguration,
                notebook_path: str
            ) -> Dict[str, Any]:
```

trungleduc:
Thanks! It looks neat now. Could you update the documentation about this feature? In the same section as the pre-launch hook.

dfguerrerom:
done! in 3f2ce382b3b9ea1f164bee953de5a7e3beb275de I also added the page_config and prelaunch hooks to be used by the `server extension` , is that okay?

trungleduc (approved):

trungleduc:
@dfguerrerom I moved the hooks to `VoilaConfiguration` to simplify the setup for the server extension case. 
@martinRenou could you take a look at this PR?

dfguerrerom:
awesome @trungleduc ! Thanks

martinRenou (approved):
Thanks!!

PR #1501, opened 2024-10-16
Add backward compatibility for ipywidgets 7

martinRenou:
Closes https://github.com/voila-dashboards/voila/pull/1487

Closes https://github.com/voila-dashboards/voila/pull/1394 as this PR takes the approach of extracting the widget manager package out of the Voila one too.

--- 
### How?

Voila now ships two new labextensions:
- `@voila-dashboards/widgets-manager7` for dealing with ipywidgets 7
- `@voila-dashboards/widgets-manager8` for dealing with ipywidgets 8+

The server will check which version of `@jupyter-widgets/jupyterlab-manager` is already installed there, and decide to inject either one or the other Voila widget manager in the page.

---

### DIFF
--- a/docs/customize.md
+++ b/docs/customize.md
@@ -759,3 +759,12 @@ or using `voila.json` file
 :::{warning}
 Progressive rendering mode is incompatible with the preheated kernels functionality.
 :::
+
+## Custom labextension for Voila
+
+Since Voila 0.5.0, it can now load JupyterLab extensions. Allowing to load custom mimetype extensions, JupyterLab themes etc.
+
+Voila will load all extensions under `{PREFIX}/share/jupyter/labextensions` just like JupyterLab and Jupyter Notebook v7.
+
+It also has its own path for labextensions that you can use: `{PREFIX}/share/jupyter/voila/labextensions`. Installing extensions there will make those extensions only available to Voila. This is really useful if you
+don't want to pollute JupyterLab or Jupyter Notebook with extensions that only work for Voila.


martinRenou:
Damn, I did not expect the UI-tests to pass 😆

packages/voila/src/app.ts
```diff
@@ -121,23 +119,23 @@ export class VoilaApp extends JupyterFrontEnd<IShell> {
...
-  set widgetManager(manager: VoilaWidgetManager | null) {
+  set widgetManager(manager: any) {
     this._widgetManager = manager;
     if (this._widgetManager) {
       this._widgetManagerPromise.resolve(this._widgetManager);
     }
   }
 
-  get widgetManager(): VoilaWidgetManager | null {
+  get widgetManager(): any {
```
jtpio, line 133: Wondering it this `any` is temporary, or now needed? If so we could likely add a comment to explain the reason?

martinRenou: Yes I was about to add a comment about this 👍🏽 

It is needed and I'm not sure we can workaround this for now. The thing is we should really not import anything that imports `@jupyter-widgets/base` to not pollute the shared packages with a wrong ipywidgets version.

Unless there is a way to just import the type for the build to be happy but not the runtime package?

trungleduc: `import type` can work?

martinRenou: To do this I'd have to make a split of package/extension

martinRenou: Actually no

package.json
```diff
@@ -45,5 +45,133 @@
     "prettier": "^2.8.6",
     "rimraf": "^3.0.2",
     "shell-quote": "^1.7.2"
+  },
+  "resolutions": {
```
trungleduc, line 49: There is no way to avoid pinning all these packages?

martinRenou: Ah yeah we should probably remove some of them. Lazy me just copied the list of packages from jupyterlab+lumino.

I had to pin otherwise the widgets 7 package would pull jupyterlab 3 stuff.

trungleduc: Should we go further and extract the widget 7 extension into standalone packages?

martinRenou: You mean outside of the repo ?

trungleduc: No, just a standalone python package and being built independently to avoid dependencies conflict

martinRenou: 😫 this means more packaging setup, I'm honestly growing tired of our tooling and packaging.

I agree this list of pinning is annoying too we should reduce it as much as possible, let me have a look!

martinRenou: Arf, it's either we pin the world, or we make a separate Python module that makes use of JupyterLab 3 (yarn 1). I hate these tools.

martinRenou: Alternatively, we make a new build of ipywidgets 7 packages using JupyterLab 4. This was planned a long time ago but never done.

trungleduc: Let's address this issue in a follow-up

trungleduc (approved):

package.json
martinRenou: Agreed, let's release Voila with this.

I'm looking into building ipywidgets 7 against lab 4 as we speak so we can get rid of the pinning.

packages/voila/src/plugins/outputs/tools.ts
```diff
@@ -63,7 +62,7 @@ export function handleExecutionResult({
 }: {
   payload: IExecutionResultMessage['payload'];
   rendermime: IRenderMimeRegistry;
-  widgetManager: VoilaWidgetManager;
+  widgetManager: any;
```
jtpio, line 65: Could this have used the `type` approach too?

martinRenou: Ah indeed I missed that one

dfguerrerom:
my `page_config_hook` doesn't work anymore after this PR....

Screenshot from 2024-10-23 10-19-18

my previous logic was only adding the labextensions that were available in the virtual environment that I wanted to use... now I have refactored to also include the voila extensions... however, there's nothing rendered in my voila page.

my federated extensions are:

```python
        "federated_extensions": [
            {
                "name": "jupyterlab-plotly",
                "load": "static/remoteEntry.41e12212782f05d7d17b.js",
                "extension": "./extension",
                "mimeExtension": "./mimeExtension",
                "entrypoints": None,
            },
            {
                "name": "jupyter-vue",
                "load": "static/remoteEntry.f993df579401788f32ef.js",
                "extension": "./extension",
                "entrypoints": None,
            },
            {
                "name": "jupyter-vuetify",
                "load": "static/remoteEntry.cba680c7ed013f0591de.js",
                "extension": "./extension",
                "entrypoints": None,
            },
            {
                "name": "jupyter-leaflet",
                "load": "static/remoteEntry.a83217134ba8dacb85fc.js",
                "extension": "./extension",
                "entrypoints": None,
            },
            {
                "name": "jupyterlab_pygments",
                "load": "static/remoteEntry.5cbb9d2323598fbda535.js",
                "extension": "./extension",
                "style": "./style",
                "entrypoints": None,
            },
            {
                "name": "@jupyter-widgets/jupyterlab-manager",
                "load": "static/remoteEntry.e4ff09401a2f575928c0.js",
                "extension": "./extension",
                "entrypoints": None,
            },
            {
                "name": "@jupyterhub/jupyter-server-proxy",
                "load": "static/remoteEntry.569b1096d01998b85e34.js",
                "extension": "./extension",
                "entrypoints": None,
            },
            {
                "name": "@voila-dashboards/widgets-manager8",
                "load": "static/remoteEntry.958dac8c7410b5fcc9ee.js",
                "extension": "./extension",
                "entrypoints": None,
            },
        ],
```

Is there anything else I've been missing?

PR #1532, opened 2025-06-18
Set JPY_SESSION_NAME to the notebook path (same as jupyer_server does)

nocnokneo:
## References

- https://github.com/jupyterlab/jupyterlab/issues/16282

## Code changes

Set `JPY_SESSION_NAME` to the notebook path, same as `jupyer_server` does. This ensures that notebooks that use this feature in Jupyter Lab continue to work when served by voila.

## User-facing changes

- None

- None

### DIFF
--- a/voila/handler.py
+++ b/voila/handler.py
@@ -27,6 +27,7 @@ from ._version import __version__
 from .notebook_renderer import NotebookRenderer
 from .request_info_handler import RequestInfoSocketHandler
 from .utils import ENV_VARIABLE, create_include_assets_functions, get_page_config
+from jupyter_server.utils import ApiPath, to_os_path
 
 
 class BaseVoilaHandler(JupyterHandler):
@@ -91,8 +92,14 @@ class VoilaHandler(BaseVoilaHandler):
             return
 
         cwd = os.path.dirname(notebook_path)
+        notebook_os_path = to_os_path(
+            ApiPath(notebook_path), os.path.abspath(self.contents_manager.root_dir)
+        )
+
         # Adding request uri to kernel env
         request_info = {}
+        # Expose the absolute OS path to the notebook in the same way as jupyter_server does
+        request_info[ENV_VARIABLE.JPY_SESSION_NAME] = notebook_os_path
         request_info[ENV_VARIABLE.SCRIPT_NAME] = self.request.path
         request_info[ENV_VARIABLE.PATH_INFO] = (
             ""  # would be /foo/bar if voila.ipynb/foo/bar was supported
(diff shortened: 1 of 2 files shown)


trungleduc:
Thanks for working on it. Could you verify if the exposed path is a relative path to the current voila directory or the absolute path to the notebook?

nocnokneo:
> Thanks for working on it. Could you verify if the exposed path is a relative path to the current voila directory or the absolute path to the notebook?

Reworked this to actually work :-)  Now returns the absolute OS path to the notebook. Tested with invocations like `voila` and `voila notebooks/notebook_path.ipynb`

nocnokneo:
Any additional changes you'd like to see?

trungleduc:
Hey, sorry for the late reply. I don't feel comfortable with exposing the full file path to the frontend. In JupyterLab, you have access to the server file system via the terminal or by executing Python code, so there is no point in hiding the file path. But Voila environment is more restrained, and exposing a relative path is enough.

What do you think @jtpio @martinRenou?

nocnokneo:
> Similar feeling at first, without checking if it would be real security issue.

I agree that the security implications are important to consider, but I think there's simple way to look at this. From the security model perspective, keeping the notebook path information secret is hard to do so building a policy that assumes path secrecy creates a brittle contract: one debugging log message or third‑party extension can blow the guarantee. Voila’s existing model of "no code execution, no arbitrary file access" is a cleaner, testable boundary. Not wanting to add the notebook path to an environment variable feels like taking a "security through obscurity" approach.

> Maybe your Pixi provisioner would be able to use the information the pre_launch_hook would make available?

I considered that initially, but it doesn't work with pre-heated kernerls which we plan to use soon. Open to suggestions here, but still feel like having Voila aligned out-of-the-box with the precedent set by Jupyter Server and JupyterLab maximizes portability, simplifies the ecosystem, and keeps security boundaries clear and enforceable.

Thanks for considering — happy to elaborate or help with a documentation note if that would move the PR forward.

trungleduc:
> From the security model perspective, keeping the notebook path information secret is hard to do so building a policy that assumes path secrecy creates a brittle contract: one debugging log message or third‑party extension can blow the guarantee.

I'm curious how can a third party extension can get the notebook path if the server does not expose this info?

ctcjab:
Was about to submit a PR to do the same thing before I found this one! Glad someone beat me to it, but sorry to see it's been stalled for almost a year.

Please consider this a big +1 from someone who supports a lot of Voila users who have some custom integrations, where we have to keep adding special workarounds to ensure they keep working when notebooks are run with Voila, since Voila doesn't currently set `JPY_SESSION_NAME`. As well as an offer to help get this PR landed -- anything I can do?

Issue #1518, opened 2025-03-01
rendering .py notebook with Voila causes 403 error with Jupytext installed


ctcjab:
https://voila.readthedocs.io/en/stable/customize.html#run-scripts says:

> Using Jupytext is another way to support script files. After installing jupytext, Voilà will see script files as if they are notebooks, and requires no extra configuration.

However, with both Voila and Jupytext installed, using Voila to render a .py notebook results in (a redirect to the static handler which in turn throws) a 403 error.

Based on the docs above, I believe the intention is for things to Just Work when following the reproduction steps below. In the meantime, what's the simplest possible way to get this to work? Thanks in advance!

Reproduction steps:
1. `uv venv tmp/voila-env`
2. `source tmp/voila-env/bin/activate`
3. `uv pip install jupyterlab voila jupytext` [1]
4. `jupyter lab`
5. Jupyter opens in browser
6. Click Jupytext's "Python - Percent Format" button in the Launcher
7. Activate "Open with Voila in new browser tab":
<img width="1523" alt="Image" src="https://github.com/user-attachments/assets/c30afcdc-c7ce-44d6-9efb-f0e341fdb159" />
8. http://localhost:8888/voila/files/foo.py opens in a new tab (redirected from /voila/render/foo.py) with the "403 Forbidden" page:
<img width="1523" alt="Image" src="https://github.com/user-attachments/assets/056511f5-1379-4627-8271-ad8ad7e90d6e" />

[1] Here is the output showing the versions of everything installed:
```
Using Python 3.12.9 environment at: tmp/voila-env
Resolved 98 packages in 4.26s
Prepared 21 packages in 5.15s
Installed 98 packages in 915ms
 + anyio==4.8.0
 + appnope==0.1.4
 + argon2-cffi==23.1.0
 + argon2-cffi-bindings==21.2.0
 + arrow==1.3.0
 + asttokens==3.0.0
 + async-lru==2.0.4
 + attrs==25.1.0
 + babel==2.17.0
 + beautifulsoup4==4.13.3
 + bleach==6.2.0
 + certifi==2025.1.31
 + cffi==1.17.1
 + charset-normalizer==3.4.1
 + comm==0.2.2
 + debugpy==1.8.12
 + decorator==5.2.1
 + defusedxml==0.7.1
 + executing==2.2.0
 + fastjsonschema==2.21.1
 + fqdn==1.5.1
 + h11==0.14.0
 + httpcore==1.0.7
[...]

PR #1536, opened 2025-08-01
Don't merge the config if not needed

martinRenou:

### DIFF
--- a/voila/server_extension.py
+++ b/voila/server_extension.py
@@ -9,6 +9,7 @@
 
 import gettext
 import os
+from typing import Optional
 
 from jinja2 import Environment, FileSystemLoader
 from jupyter_server.base.handlers import FileFindHandler, path_regex
--- a/voila/server_extension.py
+++ b/voila/server_extension.py
@@ -53,13 +53,13 @@ def _jupyter_server_extension_points():
     return [{"module": "voila.server_extension"}]
 
 
-def load_config_file() -> Config:
+def load_config_file() -> Optional[Config]:
     """
     Loads voila.json and voila.py config file from current working
     directory and other Jupyter paths
     """
 
-    new_config = Config()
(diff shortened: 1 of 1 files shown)


ctcjab:
@martinRenou, I believe this was not intended to be a breaking change, but this change [broke our use of voila.json](https://github.com/voila-dashboards/voila/issues/1518#issuecomment-2691796775).

You should be able to reproduce this with a Dockerfile like
```
FROM quay.io/jupyter/minimal-notebook@sha256:fa912aec1c8935c2422c9b14d02229ec3824d50cbe333c6faddedb0fb63af506
USER root
RUN conda install voila=5.8.10
COPY voila.json /opt/conda/etc/jupyter/voila.json
USER ${NB_UID}:${NB_GID}
```
where voila.json contains
```json
{
  "VoilaConfiguration": {
    "extension_language_mapping": {
      ".py": "python"
    }
  }
```
Start the container, create a Jupytext .py notebook, choose "Open with Voila in new browser tab", and observe that you get a 403 Forbidden error (i.e. the config above is not respected).

For now we have downgraded back to 0.5.8, but is it possible to change how we're specifying our Voila config such that it still works with 0.5.10?

The https://voila.readthedocs.io/en/stable/customize.html documentation does not currently include a section on where Voila looks for configuration and how it is merged together, which is extra challenging given the recent changes from this PR. Thanks for any help you can provide.

trungleduc:
thanks @ctcjab for reporting it, we will fix it in 0.5.11

ctcjab:
Thanks @trungleduc! Realized I never created a new issue to track this, so just did that now in case it's helpful: #1538

ctcjab:
Along with the fix for this and the docs improvements, I'd also suggest adding a regression test that ensures future changes to configuration handling logic don't break anything.

PR #1541, opened 2025-08-18
fix: use spawn_callback to avoid blocking socket ping/pongs

zealenth:
## References

When using progressive rendering to render reports that take more than 90s, the report hangs indefinitely.

Reproduced with a cell
```
import time
time.sleep(91)
print('hi')
```

## Code changes

I tracked this down to https://github.com/tornadoweb/tornado/issues/2021.  In our case, the execution handler's on_message method blocks any ping/pong requests from being received, resulting in the websocket being closed after 90s

I've taken their recommend approach and offloaded the notebook handling to spawn_callback.

## User-facing changes

Slow reports that previously hung will now render.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
+## 0.5.11
+
+([Full Changelog](https://github.com/voila-dashboards/voila/compare/@voila-dashboards/jupyterlab-preview@2.3.10...86dbefc4436d27f8d90e838d43b8b191eacb98df))
+
+### Enhancements made
+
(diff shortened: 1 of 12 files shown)


voila/tornado/execution_request_handler.py
```diff
@@ -48,91 +49,107 @@ async def on_message(
...
+            self._executor = executor = VoilaExecutor(
+                nb,
+                km=km,
+                config=execution_data["config"],
+                show_tracebacks=execution_data["show_tracebacks"],
+            )
+            executor.kc = await executor.async_start_new_kernel_client()
+
+            # schedule notebook run and return so we don't block other messages
+            IOLoop.current().spawn_callback(self._run_notebook, nb, executor)
```
trungleduc, line 93: Can we call `add_callback` directly?   https://github.com/tornadoweb/tornado/blob/d30ef74020924be5f214091de95314ca3ce919af/tornado/ioloop.py#L673

Issue #1538, opened 2025-08-12
Config merging changes in 0.5.9 broke the use of voila.json


ctcjab:
The changes in 0.5.9 [broke our use of voila.json](https://github.com/voila-dashboards/voila/issues/1518#issuecomment-2691796775).

You should be able to reproduce this with a Dockerfile like
```
FROM quay.io/jupyter/minimal-notebook@sha256:fa912aec1c8935c2422c9b14d02229ec3824d50cbe333c6faddedb0fb63af506
USER root
RUN conda install voila=0.5.9 jupytext=1.17.2
COPY voila.json /opt/conda/etc/jupyter/voila.json
USER ${NB_UID}:${NB_GID}
```
where voila.json contains
```json
{
  "VoilaConfiguration": {
    "extension_language_mapping": {
      ".py": "python"
    }
  }
```
Start the container, create a Jupytext .py notebook, choose "Open with Voila in new browser tab", and observe that you are unable to render the .py notebook (i.e. the config above is not respected).

For now we have downgraded back to 0.5.8.

While we're waiting for the fix, is it possible to change how we're specifying our Voila config such that it will still work with the latest release? The https://voila.readthedocs.io/en/stable/customize.html documentation does not currently include a section on where Voila looks for configuration and how it is merged together, which is extra challenging given the recent changes here. Thanks for any help you can provide.

jtpio:
Closing as fixed by https://github.com/voila-dashboards/voila/pull/1543 and released in https://github.com/voila-dashboards/voila/releases/tag/v0.5.11

@ctcjab please try the new release, and feel free to reopen if needed, thanks!

PR #1543, opened 2025-08-25
Server extension: Fix merging with local config

martinRenou:
Fixing https://github.com/voila-dashboards/voila/issues/1538


trungleduc (approved):

jtpio:
Thanks!

Ideally it would be great to have tests for this. But maybe it could be tracked in an issue if there is none currently and may take a while to add.

Issue #1428, opened 2023-12-10
Rendering stuck on cell that reads data


Dr4cky:
Hi there,

I am having an issue and I have no idea why it is not working. The rendering gets stuck on the following cell:

```
path = Path(r"data.csv")
df = pd.read_csv(path, index_col=0, delimiter=';')
```
It doesn't show an error, it just keeps loading forever "Executing 2 out of 6".
If I just simply run this cell in my notebook, it runs instantly. Any idea what is going wrong?

Thanks!

ggogate:
Anything in the browser Console?

Dr4cky:
> Anything in the browser Console? Also can you try loading a different, smaller sample CSV which is comma separated.

I just found out that it is not actually that cell that causes the problem, but the cell after:

```
returns_plot = widgets.Output()
with returns_plot:
    text = widgets.HTML('Time series data.')
    fig_ts = show_ts(df)
    interact(fig_ts)
    display(text)
```

A portion of the error that I see in the console is as follows:

> id
>     raise web.HTTPError(404, "Kernel does not exist: %s" % kernel_id)
> tornado.web.HTTPError: HTTP 404: Not Found (Kernel does not exist: c5416394-05a1-485e-a9bb-595a9d5e2387)
> 404 GET /api/kernels/c5416394-05a1-485e-a9bb-595a9d5e2387?1702219619084 (::1) 4.17ms
>

ggogate:
Try commenting out the figure part and see if the HTML displays okay. Then try moving the fig_ts part above the 'with'.

Dr4cky:
I did a reinstall and the current error that Voila gives me is that it cannot find the module plotly, even though it is installed in my environment and I can the code in the notebook just fine. Is it calling another kernel?

ggogate:
Try starting Jupyter/voila with --enable_nbextensions=True

Dr4cky:
Gave me the same errors unfortunately, so plotly module not found and kernel does not exist.

trungleduc:
Hi, could post the full reproduction code?

ggogate:
Okay, able to recreate the same on another system/vm? Usually it works very well together, so trying to eliminate any external factors. 
Maybe try installing previous stable versions of libraries. 
If you can share the requirements.txt using 'pip freeze' I can try to recreate it at my end.

Dr4cky:
My code is as follows:

```
import numpy as np
import pandas as pd
from pathlib import Path
[...]

PR #1544, opened 2025-09-15
Send execution requests from frontend in progressive rendering mode

trungleduc:
## References

- This PR refactors progressive rendering in Voilà, enabling the frontend to dispatch cell execution requests incrementally via the Jupyter kernel websocket API instead of a custom execution request endpoint

- Notebook cell execution is now fully compatible with the Jupyter kernel API, reducing reliance on Voilà-specific server endpoints and the `nbconvert` execution mechanism. As a by-product, https://github.com/voila-dashboards/voila/issues/1428 is fixed in the new progressive rendering mode

https://github.com/user-attachments/assets/572524fe-e750-4f1c-b44f-e5ffc12331ec

### Frontend

- The frontend now loops through notebook cells and sends execution requests one at a time using the kernel websocket protocol (instead of a custom execution request handler).
- Output areas for each cell are created and attached dynamically; visual skeletons are removed as soon as execution completes.
- To avoid leaking notebook source code, the execution code is the cell index, which the server translates to the actual cell source before forwarding the message to the kernel.
- The rendering logic is simplified: outputs are updated in-place, and loading indicators are managed per cell.

### Backend

- The custom WebSocket-based `ExecutionRequestHandler` is removed. Instead, progressive rendering is now handled via the standard kernel websocket handler (`VoilaKernelWebsocketHandler`).
- The handler intercepts `execute_request` messages, replaces the code payload (which is the cell index) with the actual notebook cell source, and passes the message to the Jupyter kernel.
- Output and error messages are masked as needed based on configuration.

## User-facing changes
N/A

- The custom execution endpoint `/voila/execution/` is removed
- The kernel handler endpoints `/api/kernels/` and `/api/kernels/%s/channels` are scoped under the `/voila/` prefix and become `/voila/api/kernels` and `/voila/kernels/%s/channels`


trungleduc:
Due to the breaking change, this feature should be released in `0.6`

PR #1559, opened 2026-09-02
Switch to jupyter-builder and fix CI

martinRenou:
- switch from jupyterlab/builder to the new jupyter-builder package
- update lab packages
- update node and yarn
- switch from jlpm to yarn
- patch Backbone to work seamlessly with conflicting rspack/requirejs.


martinRenou:
So here is the state of this PR:
- (1): latest ipywidgets is built with jupyter-builder and rspack instead of webpack
- (2): Point (1) breaks ipywidgets support in Voila -> Our widget manager fails to load the module for ipywidgets.
- (3): If I switch to jupyter-builder, things start to work for all templates **except the Reveal template and ipywidgets 7 support** which gets broken:
  - The Reveal template ships requirejs in the page, which makes BackboneJS [enter this condition](https://github.com/jashkenas/backbone/blob/master/backbone.js#L16), which results in the rspack backbone import `backbone__rspack_import_0` to be `undefined` resulting in widgets to not load in the reveal template **+** It seems there is a race condition now witch the reveal template not rendering at all once out of three with yet another requirejs issue.
  - ipywidgets 7 support gets broken for similar reasons I have not investigated too deep. But we get the same symptom  `*__rspack_import_0` is `undefined` for the jupyter-widgets base and controls packages.

PR #1560, opened 2026-09-03
request confirmation before rendering untrusted notebook

trungleduc:
## User-facing changes

When opening a notebook using the Voila preview, users need to confirm that they want to render untrusted one

https://github.com/user-attachments/assets/79544d5e-477d-47f4-a77e-196ec7caf2f5


martinRenou (approved):
Looks good ! The CI will get green eventually with https://github.com/voila-dashboards/voila/pull/1561

PR #1561, opened 2026-09-04
Patching the reveal template to fix ipywidgets 8 support there + fix CI

martinRenou:
## References

The newly built ipywidgets 8.1.9 was made using the new jupyter-builder package which uses rspack instead of webpack. This does not play well with the combination of:
- requirejs put in the page by the reveal.js template
- backbone detecting require.js and this condition is true https://github.com/jashkenas/backbone/blob/master/backbone.js#L16 then backbone hides itself from rspack (`backbone__rspack_import_0` is undefined) resulting in ipywidgets failing to import backbone.

This PR works around this by removing requirejs from the page, and processing reveal with webpack instead of having it raw in the template.


trungleduc (approved):
🎉

